PR #147, opened 2024-04-05
Added ExternalObjectRequest RPC

johndanenRV:


mejohnnaylor (dismissed):

example_device_models/device.minimal.json
```diff
@@ -79,6 +79,11 @@
     }
   },
   "params": {
+    "externalObject": {
+      "type": "STRING",
+      "oid_aliases": ["0xFF0E", "0xFF0C"],
```
james-peltzer, line 84: FF0C is for the icon
FF0E is for the device UI 
One is an image file and one is an OGLML (XML) file.  They cannot be defined as aliases for the same parameter - they need to be different parameters referencing different external objects.

mejohnnaylor (approved):

PR #159, opened 2024-05-02
Added scope logic

johndanenRV:


sdks/cpp/common/examples/full_service/full_service.cpp
```diff
@@ -148,6 +172,12 @@ std::shared_ptr<grpc::ServerCredentials> getServerCredentials() {
         ssl_opts.pem_key_cert_pairs.push_back(
           grpc::SslServerCredentialsOptions::PemKeyCertPair{server_key, server_cert});
         ans = grpc::SslServerCredentials(ssl_opts);
+
+        if (absl::GetFlag(FLAGS_authz)) {
+            const std::shared_ptr<grpc::AuthMetadataProcessor> authzProcessor(new JWTAuthMetadataProcessor());
+            ans->SetAuthMetadataProcessor(authzProcessor);
```
mejohnnaylor, line 178: authzProcessor object goes out of scope line 179. investigate what SetAuthMetadataProcessor does with it - if it makes a copy, that's ok, if it's storing a reference, or pointer, that's a problem because it's now a dangling reference (i.e. a reference to a deleted object)

sdks/cpp/common/examples/full_service/full_service.cpp
```diff
@@ -284,6 +245,41 @@ class CatenaServiceImpl final : public catena::CatenaService::AsyncService {
...
+        if (absl::GetFlag(FLAGS_authz) == false) {
+            return {AUTHZ_DISABLED};
+        }
+
+        std::vector<grpc::string_ref> claimsStr = context.auth_context()->FindPropertyValues("claims");
+        if (claimsStr.empty()) {
+            throw catena::exception_with_status("No claims found", catena::StatusCode::PERMISSION_DENIED);
+        }
+        // parse string of claims into a picojson object
+        picojson::value claims;
```
mejohnnaylor, line 258: so we're using picojson? How is it included? What are it's license terms?

sdks/cpp/common/examples/full_service/full_service.cpp
```diff
@@ -323,13 +319,14 @@ class CatenaServiceImpl final : public catena::CatenaService::AsyncService {
 
                 case CallStatus::kProcess:
                     new GetValue(service_, dm_, ok);
-                    if (ok) {
+                    if (ok){
                         context_.AsyncNotifyWhenDone(this);
-                        if (validateRequest_() && !context_.IsCancelled()) {
+                        if (validateRequest_()) {
```
mejohnnaylor, line 324: check this doesn't break the fix for failing to terminate server when a client is connected

sdks/cpp/common/examples/full_service/full_service.cpp
```diff
@@ -323,13 +319,14 @@ class CatenaServiceImpl final : public catena::CatenaService::AsyncService {
...
                             try {
+                                std::vector<std::string> clientScopes = getScopes(context_);
```
mejohnnaylor, line 326: it's understandable why you made client scopes a vector of strings. however, it could be better / more efficient to code them as a bit-field. with monitor being bits 0 & 1, operate 2 & 3 and so on.

operate read access is signaled by bit 2 being 1, no read access signaled by it being 0, write access is granted when bit 3 is 1, and prohibited when it is zero.

don't do this now - until it's proven that hauling strings around and parsing them for every parameter access is unduly burdensome - "premature optimization" is a curse!

sdks/cpp/common/include/ParamAccessor.h
```diff
@@ -42,6 +42,8 @@
 #include <typeinfo>
 #include <typeindex>
 
+#define AUTHZ_DISABLED "__AUTHZ_DISABLED__"
```
mejohnnaylor, line 45: hey! 1970's code!
in C++ we prefer to write const std::string kAuthzDisabled("__AUTHZ_DISABLED__"); 
why?
that's a good research topic for you. something to do with being able to pass by reference, take addresses etc.

mejohnnaylor (commented):
please address inline comments.

sdks/cpp/common/examples/full_service/full_service.cpp
johndanenRV: This branch was made before that fix was merged so this would not have affected it. I updated the branch and checked to be safe, and it is fine.

sdks/cpp/common/include/ParamAccessor.h
johndanenRV: Fixed

sdks/cpp/common/examples/full_service/full_service.cpp
johndanenRV: authzProcessor is a std::shared_ptr passed by reference to SetAuthMetaDataProcessor(). It is hard to know exactly what grpc does with it behind the scene but I would assume that it makes some kind of copy of the shared_ptr. Given that the AuthMetaDataProcessor is running as expected I don't think this is an issue.

mejohnnaylor: yep - if it's anything like std::shared_ptr, it'll maintain a reference count and only delete the object once the final owning object is destroyed. This code is now considered safe ;-) Thanks for checking!

sdks/cpp/common/examples/full_service/full_service.cpp
mejohnnaylor: yep, BSD 2-clause is fine. thanks!

mejohnnaylor (approved):

PR #161, opened 2024-05-08
push updates to connected clients

johndanenRV:
Value changes are now broadcast to all connected clients with proper authorization. Also made it so that all active connections are cancelled on server shutdown.


mejohnnaylor (approved):
lots of interesting work here, John. Nice job!

PR #165, opened 2024-05-09
Fixed param accessor scope for nested params

johndanenRV:


sdks/cpp/common/include/ParamAccessor.h
```diff
@@ -228,7 +228,7 @@ class ParamAccessor {
         Value& value = value_.get();
         std::string scope = childParam.access_scope();
         if (scope == "") {
-            scope = parent.access_scope();
```
mejohnnaylor, line 231: this will create a copy of the child scope and, potentially the parent's scope too. In general copies are bad.

better code might be:

`const std::string& scope = childParam.access_scope() == "" ? parent.access_scope() : childParam.access_scope();`

note that the optimizer will cache the 2 calls to `childParam.access_scope()`

it'd also be good for you to comment what's going on here such as:

`// if the child doesn't have a scope defined, we use that of the parent`

mejohnnaylor (approved):

PR #167, opened 2024-05-10
Removed General Response Message

johndanenRV:


interface/service.proto
```diff
@@ -33,16 +33,10 @@ import "externalobject.proto";
 
 /* Catena's API */
 
-message GeneralRequest {
+message EmptyRequest {
```
mejohnnaylor, line 36: is this what JP recommended?

mejohnnaylor: - copilot just told me that this is how to do it, so comment rescinded ;-)

mejohnnaylor (approved):

interface/service.proto
johndanenRV: JP did recommend this.

PR #169, opened 2024-05-14
Added template_oid support to getValue

johndanenRV:
For uninitialized param values, getValue will now return the value from the template param if available. Does not yet support recursive templates.


mejohnnaylor:
better to not merge this now after lengthy discussion on implementation approach

PR #170, opened 2024-05-24
Device request scope

johndanenRV:
Added deviceStream class for sending device components to clients. Also added a way to get params with unauthorized fields filtered out. Probably should have done this as two separate pull requests but it's too late to change that.

### DIFF
--- a/sdks/cpp/common/include/ParamAccessor.h
+++ b/sdks/cpp/common/include/ParamAccessor.h
@@ -655,7 +655,7 @@ class ParamAccessor {
      * 
      * @todo add option to check scope for write access
     */
-    bool checkScope(std::vector<std::string>& clientScopes) const;
+    bool checkScope(const std::vector<std::string>& clientScopes) const;
 
     /**
      * @brief check if the client is authorized to access the parameter
@@ -665,7 +665,7 @@ class ParamAccessor {
      * 
      * @todo add option to check scope for write access
     */
-    bool checkScope(std::vector<std::string>& clientScopes, std::string paramScope) const;
+    bool checkScope(const std::vector<std::string>& clientScopes, const std::string& paramScope) const;
 
     /** 
      * @brief get the parameter's fully qualified object id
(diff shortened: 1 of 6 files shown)


sdks/cpp/common/include/ParamAccessor.h
```diff
@@ -637,6 +640,33 @@ class ParamAccessor {
...
+
+    /**
+     * @brief check if the client is authorized to access the parameter
+     * @param clientScopes the scopes of the client making the request
+     * @param paramScope the scope of the parameter to be accessed
+     * @return true if the client is authorized, false otherwise
+     * 
+     * @todo add option to check scope for write access
+    */
+    bool checkScope(std::vector<std::string>& clientScopes, std::string paramScope) const;
```
mejohnnaylor, line 668: same question about ` clientScopes` should it be const - also why are you passing `paramScope` by value?

johndanenRV:
All of the sources I have looked at say that enum values can only be used as template arguments if they are known at compile time. If we want to use templates for next then it would have to be done like this:
```
const catena::DeviceComponent& DeviceStream::next(){
    if (clientScopes_ == nullptr){
        throw std::runtime_error("Client scopes not attached");
    }
    switch (nextType_){
        case ComponentType::kBasicDeviceInfo:
            return next_<ComponentType::kBasicDeviceInfo>(nextType_);
        case ComponentType::kParam:
            return next_<ComponentType::kParam>(nextType_);
        case ComponentType::kConstraint:
            return next_<ComponentType::kConstraint>(nextType_);
        case ComponentType::kMenu:
            return next_<ComponentType::kMenu>(nextType_);
        case ComponentType::kCommand:
            return next_<ComponentType::kCommand>(nextType_);
        case ComponentType::kLanguagePack:
            return next_<ComponentType::kLanguagePack>(nextType_);
        case ComponentType::kFinished:
            throw std::runtime_error("No more components to send");
    }
}
```

PR #171, opened 2024-05-24
param referencing template_oid fields

Isaac-Robert-Ross:
added getValue that works for basic types and arrays. subParam is temporarily broken and will get fixed with setValue. subParaming into template_oid, recursive templating and Struct support to be added.


Isaac-Robert-Ross:
subParam has been fixed, now with template_oid checking. template_oids can now follow the whole oid path and recursively follow templates with their own template_oids inside.

Isaac-Robert-Ross:
template_oid integration is now re-written to use deep copies into the param, keeping the deviceModel up to date, also supports structs now, being able to iterate through and template nested params

mejohnnaylor (approved):

PR #181, opened 2024-06-10
Created skeleton of lite sdk

johndanenRV:
Added some basic components to the lite sdk as an early proof of concept.

Uses manually created device model files as an example of what the generated files would look like.

### DIFF
--- a/sdks/cpp-lite/common/include/DeviceAccessor.h
+++ b/sdks/cpp-lite/common/include/DeviceAccessor.h
@@ -1,0 +1,17 @@
+
+#pragma once
+
+#include "Param.h"
+#include <example.tiny.json.h>
+
+
+class DeviceAccessor {
+
+  public:
+    DeviceAccessor();
+
+    std::unique_ptr<catena::Value> getSerializedValue(std::string oid) const;
+
+  private:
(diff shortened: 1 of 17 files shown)


sdks/cpp-lite/common/include/DeviceAccessor.h
```diff
@@ -0,0 +1,17 @@
+
+#pragma once
+
+#include "Param.h"
+#include <example.tiny.json.h>
```
johndanenRV, line 5: This include statement would need to change based on the name of the generated file.

sdks/cpp-lite/common/include/DeviceAccessor.h
```diff
@@ -0,0 +1,17 @@
...
+
+class DeviceAccessor {
+
+  public:
+    DeviceAccessor();
+
+    std::unique_ptr<catena::Value> getSerializedValue(std::string oid) const;
+
+  private:
+    tiny::DeviceModel deviceModel_;
```
johndanenRV, line 16: Same issue for generated namespace.

Isaac-Robert-Ross (approved):

PR #185, opened 2024-06-19
Reorganized C++ File Structure

johndanenRV:
* Reorganized the file structure so that both full and lite sdk are stored under the same cpp directory. 
* Fixed CMake build to reflect these changes.
* Changed the VSCode workspace to reflect changes

cpp-lite sdk is now built using:
```
cmake -G Ninja -DBUILD_LITE=ON ..
```


mejohnnaylor:
duplicates work in refactor-for-lite

PR #203, opened 2024-07-22
Add UI

mejohnnaylor:
adds example DashBoard custom panel to the status_update gRPC example.
adds support for PolyglotText type and gives Param objects names of this type when constructed in the lite SDK
PolyglotText objects are included in Param's `toProto(Param&)` method so the UI should be able to display parameter names as well as their values.

### DIFF
--- a/sdks/cpp/lite/include/PolyglotText.h
+++ b/sdks/cpp/lite/include/PolyglotText.h
@@ -1,0 +1,37 @@
+#pragma once
+
+#include <IPolyglotText.h>
+
+#include <lite/language.pb.h>
+
+#include <string>
+#include <unordered_map>
+#include <initializer_list>
+
+namespace catena {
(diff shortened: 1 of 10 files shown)


sdks/cpp/lite/include/PolyglotText.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <string>
+#include <unordered_map>
+#include <initializer_list>
+
+namespace catena {
+namespace lite {
+class PolyglotText : public catena::common::IPolyglotText {
+  public:
+    using DisplayStrings = std::unordered_map<std::string, std::string>;
```
mejohnnaylor, line 15: looks redundant as this is also defined in IPolyglotText - should be removed.l

johndanenRV (dismissed):

PR #205, opened 2024-07-22
Changed oid to non const

johndanenRV:
Changed oid to not be const so that move constructor still works when compiling with clang. Also added default constructor back to IParam.

### DIFF
--- a/sdks/cpp/lite/include/IParam.h
+++ b/sdks/cpp/lite/include/IParam.h
@@ -16,6 +16,7 @@ class IParam {
     using ParamType = catena::patterns::EnumDecorator<catena::ParamType>;
   public:
     IParam() = default;
+    IParam(std::string oid) : oid_(oid) {};
     IParam(IParam&&) = default;
     IParam& operator=(IParam&&) = default;
     virtual ~IParam() = default;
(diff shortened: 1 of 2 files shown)


mejohnnaylor (dismissed):

sdks/cpp/lite/include/IParam.h
```diff
@@ -43,10 +43,7 @@ class IParam {
     /**
      * @brief return the oid of the param
      */
-    virtual const std::string& getOid() const { return oid_; }
-    
-  protected:
-    const std::string oid_;
+    virtual const std::string& getOid() const = 0;
```
mejohnnaylor, line 46: I think this should be a public method. what's the rationale for making it protected?

mejohnnaylor (changes requested):
getOid method should be public.

sdks/cpp/lite/include/IParam.h
johndanenRV: I think you misread the change. Protected was removed so it is public.

mejohnnaylor (dismissed):

mejohnnaylor (approved):

PR #210, opened 2024-07-24
Added valueSetByService signal

johndanenRV:


sdks/cpp/connections/gRPC/examples/status_update/status_update.cpp
```diff
@@ -117,13 +117,16 @@ void statusUpdateExample(){
...
         Param<int32_t>& aNumber = *dynamic_cast<Param<int32_t>*>(dm.getItem("/counter", Device::ParamTag{}));
         while (globalLoop) {
             std::this_thread::sleep_for(std::chrono::seconds(1));
-            Device::LockGuard lg(dm); 
-            aNumber.get()++;
+            {
+                Device::LockGuard lg(dm); 
+                aNumber.get()++;
+            }
+            dm.valueSetByServer.emit("/counter", &aNumber, 0);
```
mejohnnaylor, line 129: this is a problem - aNumber could be updated by another thread so it should be called within the scope of the lock guard. Of course this means that any recipients of the signal must not try to apply the lock, and they should be minimize the time spent in the function call.

sdks/cpp/connections/gRPC/src/ServiceImpl.cpp
```diff
@@ -217,9 +217,14 @@ void CatenaServiceImpl::GetValue::proceed(CatenaServiceImpl *service, bool ok) {
...
                 // std::vector<std::string> clientScopes = getScopes(context_);
                 catena::Value ans;
+                catena::lite::IParam* param = dm_.getItem(req_.oid(), Device::ParamTag{});
+                    if (param == nullptr) {
+                    std::stringstream why;
+                    why << __PRETTY_FUNCTION__ << "\nparam '" << req_.oid() << "' not found";
+                    throw catena::exception_with_status(why.str(), catena::StatusCode::NOT_FOUND);
+                }
                 {
                     Device::LockGuard lg(dm_);
```
mejohnnaylor, line 227: this lock guard should not be asserted. see prior comment

sdks/cpp/connections/gRPC/src/ServiceImpl.cpp
```diff
@@ -351,22 +361,22 @@ void CatenaServiceImpl::Connect::proceed(CatenaServiceImpl *service, bool ok) {
...
-            //         // Don't need to send any updates to unauthorized clients
-            //     } 
-            // });
+            valueSetByServerId_ = dm_.valueSetByServer.connect([this](const std::string& oid, const IParam* p, const int32_t idx){
+                try{
+                    if (!this->context_.IsCancelled()){
+                        //std::vector<std::string> scopes = getScopes(this->context_);
+                        this->res_.mutable_value()->set_oid(oid);
+                        this->res_.mutable_value()->set_element_index(idx);
+                        Device::LockGuard lg(dm_);
```
mejohnnaylor, line 370: and here

mejohnnaylor (changes requested):

mejohnnaylor (approved):

mejohnnaylor (approved):

PR #226, opened 2024-08-01
Added read_only flag to params

johndanenRV:


sdks/cpp/lite/src/Param.cpp
```diff
@@ -22,6 +22,7 @@ void Param<int32_t>::toProto(Value& value) const {
 
 template <>
 void Param<int32_t>::fromProto(const Value& value) {
+    if (read_only_) throw std::runtime_error("Cannot set read-only parameter");
```
mejohnnaylor, line 25: it's better to not throw an exception here, just don't call the following `fromProto()` and return from the function

mejohnnaylor: same comment for the other `throw` statements in this file

tools/codegen/cppgen.js
```diff
@@ -258,6 +265,7 @@ class CppGen {
                 // instantiate the deserialize specialization
                 bloc(`template<>`, indent);
                 bloc(`void catena::lite::Param<${fqname}>::fromProto(const catena::Value& value) {`, indent);
+                bloc(`if (read_only_) throw std::runtime_error("Cannot set read-only parameter");`, indent+1);
```
mejohnnaylor, line 268: don't throw, just ignore the setValue request

sdks/cpp/lite/include/Param.h
```diff
@@ -133,6 +138,7 @@ template <typename T> class Param : public IParam {
     std::reference_wrapper<T> value_;
     std::reference_wrapper<Device> dm_;
     std::string widget_;
+    const bool read_only_;
```
mejohnnaylor, line 141: this shouldn't be a const. it's valid for business logic to change this flag. It's not valid for connection logic to do so though, so IParam should only provide a read accessor for this flag, the write accessor would be provided by either lite::Param or full::Param.

mejohnnaylor (changes requested):
one minor change requested

mejohnnaylor (approved):

Issue #219, opened 2024-07-26
codegen refactor STRUCT


mejohnnaylor:
do this before adding support for STRUCT_ARRAY. This work is already started but not complete.

mejohnnaylor:
feat_#219_jrn

PR #242, opened 2024-08-20
Feat #219 jrn

mejohnnaylor:
a good staging post on the refactoring journey in that the 3 simple examples of the lite device model all work. This will break the update_service because toProto/fromProto are not yet supported, but I think it's worth breaking and fixing everything than letting this branch diverge more than it already has.


tools/codegen/device.minimal.json.cpp
```diff
@@ -0,0 +1,68 @@
+// This file was auto-generated. Do not modify by hand.
```
johndanenRV, line 1: file should be deleted

tools/codegen/device.minimal.json.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
johndanenRV, line 1: file should be deleted

sdks/cpp/lite/include/wibble.cpp
```diff
@@ -0,0 +1,108 @@
+#include <functional>
```
johndanenRV, line 1: file should be deleted

johndanenRV (changes requested):

johndanenRV (approved):

Issue #235, opened 2024-08-12
shared constraints shouldn't be references


mejohnnaylor:
The schema currently allows members of the shared constraints object (part of the device object) to be references.

Overly complex and confusing. Let's change the schema to make this invalid.

mejohnnaylor:
the same restriction applied to `ParamDescriptor`s that are inherited via the `template_oid` mechanism. What I did there was this:

```js
    let fqoid = `${parentOid}/${oid}`;
    if (~p.isTemplated()) {
      this.templateParams[fqoid] = new TemplateParam(objectType, init);
    }
```

The param object, p, knows whether it's templated or not on construction.
The only params that are available to inherit from are in the `templateParams` object, owned by `cppgen`.

We can do the same with constraints. 

i.e. create a `constraints.js` that's along the same lines as `params.js` inasmuch as it defines a `cppCtor` sub-class that can then be used by cppgen.js and params.js to process their constraints

PR #245, opened 2024-08-21
Bug #235 ipr

Isaac-Robert-Ross:
follows issue #235

### DIFF
--- a/schema/catena.schema.json
+++ b/schema/catena.schema.json
@@ -208,7 +208,14 @@
           "$ref": "#/$defs/value"
         },
         "constraint": {
-          "$ref": "#/$defs/constraint"
+          "oneOf": [
+            {
+              "$ref": "#/$defs/constraint"
+            },
+            {
+              "$ref": "#/$defs/constraint_ref"
+            }
+          ]
         },
         "max_length": {
           "$ref": "#/$defs/max_length"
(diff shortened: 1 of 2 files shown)


sdks/cpp/connections/gRPC/examples/status_update/device.status_update.json
```diff
@@ -32,13 +32,26 @@
...
+          "es": "Contador",
+          "fr": "Compteur"
+        }
+      },
+      "value": { "int32_value": 0 },
+      "constraint": { "ref_oid": "/constraints/counter_cap" }
     },
     "hello": {
       "type": "STRING",
-      "name": {"display_strings": {"en": "Text Box", "es": "Caja de texto", "fr": "Boîte de texte"}},
```
mejohnnaylor, line 40: please leave this as a single line - we've had "sensitivity" to too much whitespace.

schema/catena.schema.json
```diff
@@ -1162,66 +1204,47 @@
...
+        },
+        {
+          "$ref": "#/$defs/string_constraint"
         }
       ]
     },
+    "constraint_ref": {
+      "properties": {
+        "ref_oid": {
+          "type": "string",
```
mejohnnaylor, line 1241: can we add a pattern specifier so that only valid `oids` can be part of the json pointer? https://json-schema.org/understanding-json-schema/reference/string

sdks/cpp/connections/gRPC/examples/status_update/device.status_update.json
Isaac-Robert-Ross: I see, I might have to tweak my prettier config from the default

mejohnnaylor (approved):

PR #250, opened 2024-08-23
Moved cpp codegen files into their own folder

gRid08:
Moved the codegen cpp files into cpp folder and made changes to require statements.


mejohnnaylor (commented):
Hi Ashutosh,

some comments

- validator doesn't need to move
- it doesn't look like you changed the `require` statements that include these modules in the various source files, so the code shouldn't run successfully

Please test that the program works after these changes.

Note - I'm on vacation next week, moving house, so may not be able to respond quickly

John

PR #251, opened 2024-08-28
Struct_Array Codegen

johndanenRV:


tools/codegen/param.js
```diff
@@ -229,6 +224,24 @@ class Param extends CppCtor {
...
+   * 
+   * @returns true if the param has a value, false otherwise
+   */
+  hasValue() {
+    return "value" in this.desc;
+  }
+
+  isArrayType() {
+    return (
+      this.desc.type == "STRING_ARRAY" ||
```
mejohnnaylor, line 237: maybe `this.desc.type.includes("ARRAY")` would be better here?

sdks/cpp/lite/examples/use_templates/device.AudioDeck.json
mejohnnaylor, line 1: I see this is in use_templates, but the example there doesn't use it and I don't see any .h or .cpp files generated. Is this PR a WIP?

mejohnnaylor (changes requested):

sdks/cpp/lite/examples/use_templates/device.AudioDeck.json
johndanenRV: I have moved it to the example_device_models folder

mejohnnaylor (approved):

PR #255, opened 2024-09-16
Ibc 2024

mejohnnaylor:
fixed constraints codegen, fixed setValue, added setValue / getValue methods to device model for connection managers to use to support the RPCs / API calls of the same name.

### DIFF
--- a/sdks/cpp/lite/include/ParamWithValue.h
+++ b/sdks/cpp/lite/include/ParamWithValue.h
@@ -74,19 +74,35 @@ class ParamWithValue : public catena::common::IParam {
         catena::lite::toProto<T>(value, &value_.get());
     }
 
-    void fromProto(catena::Value& value) override {
-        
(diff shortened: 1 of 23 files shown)


sdks/cpp/lite/include/ParamWithValue.h
```diff
@@ -70,17 +72,40 @@ class ParamWithValue : public catena::common::IParam {
...
     }
 
+    /**
+     * @brief deserialize the parameter value from protobuf
+     * @param dst the destination value
+     * @param value the protobuf value to deserialize from
+     * @note should only be called for non structured types
+     */
+    void fromProto(void* dst, catena::Value& value) override {
+        assert(dst == &value_.get());
```
johndanenRV, line 94: Why is this here? If dst and value.get() are the same then why not remove the dst parameter?

mejohnnaylor:
overtaken by events

PR #257, opened 2024-09-21
Moved CPP codegen files to CPP folder

BeanJ4m:
Moved CPP codegen files to CPP folder and updated require statements.

The program builds lite and full versions, but fails to run ./connections command in a similar manner to /Develop branch.


BeanJ4m:
Not sure if constraints should be in CPP or codegen as it currently only holds CPP constraints.

johndanenRV (approved):

johndanenRV (approved):

Issue #258, opened 2024-09-24
Error handling for Set Value


johndanenRV:
We need to decide how the fromProto method should handle request from the client that are partially invalid. i.e. if the client sends a set value request for a struct and one of the struct_value fields given is a field that the client is not authorized to write to, or it does not exist within the param, then how should the error be handled.

**Option 1: throw an error**
If the request includes an invalid field, then throw an error and cancel the request with bad status. In order to avoid misleading error messages, the bad status response from the service would include json pointer to whatever field caused the error.

**Option 2: Ignore invalid fields**
If the request includes an invalid field, then fromProto will simply ignore that field and continue to write any valid fields from the incoming value. The client would then see in the update pushed by the connect rpc that the invalid field did not change.

johndanenRV:
Notes from today's meeting:
- In the case of a partially valid request, the valid parts of the request will still be processed. The invalid parts of the request will be skipped, and error info should be logged and passed to the client. The call should end with status::ok.
- For functions that need to be able to return both a value and an error status, we will define a result info struct so that both can be returned together. 
  - The implementation of this should include documentation to clarify that this makes use of [C++ return value optimization](https://stackoverflow.com/questions/12953127/what-are-copy-elision-and-return-value-optimization/12953129#12953129) to avoid making unnecessary copies of the return value and the error status
  - for cases where the error is fatal and there is no param to return, we can use [std::optional](https://en.cppreference.com/w/cpp/utility/optional) to return std::nullopt instead of returning null pointers
- Important implementation detail to avoid information leaking to attackers: 
  - trying to write to a param that a client DOES have read access for should return an unauthorized error status. 
  - trying to read/write to a param that a client DOES NOT have read access for should return a does not exist error status.

johndanenRV:
I've been reading the [gRPC Error Handling](https://grpc.io/docs/guides/error/) documentation. Currently we are using the standard error model provided gRPC but the documentation also talks about a richer error model provided by Protobuf. I am wondering if this model would be better suited for us because, as the documentation states, "This model enables servers to return and clients to consume additional error details expressed as one or more protobuf messages." This would be useful because set value request can include an arbitrary number of invalid sub-params so it would be better to send the error details as multiple protobuf messages.

There are some drawbacks also explained in the documentation, but this would be worth considering. With the generic gRPC error handling it will be difficult to provide detailed error info to clients.

mejohnnaylor:
I think we need to keep it simple and move on to bigger issues.

[...]

PR #259, opened 2024-09-25
Navigation to params using reflection

### LINKED ISSUE #239: Design for navigating to parameter values - requires implementation
We'll use a simple 4-channel audio deck as a worked example.

Its value will be defined something like this:
image

Visualized thus:
image

where the Q-factor control we want to change is shaded yellow.

This would require a `setValue` or `getValue` RPC with this json-pointer: `/audioDeck/1/eq/1/QFactor`.

let's create a `catena::Path` object from the json-pointer, which is simply an array of the elements, in order.

We can then ask the device model for the top-level parameter:

```cpp
using Audiodeck = std::vector<AudioChannel>;
Path path(jptr); // jptr is part of the RPC payload
auto deck_param = dynamic_cast<ParamWithValue<Audiodeck>*> dm.getParam(path, 0); // dm is the device model
auto& deck = reinterpret_cast<Audiodeck>deck_param.get();

// 0 is the index to use into the path object

// deck is now our std::vector<AudioChannel> object.
```

then we can get a pointer to the object we want like this:

```cpp
void * q_ptr = dm.pointAt(deck, path, 1); // base address, path to follow, part of path to use
```

sketch implementation of point at...
```
void* pointAt(void* base, Path path, idx) {
  void* next;
  IParam* currentParam = this;
  if (type of path[idx] is a string) {
    // it's a child parameter oid, so recurse
    std::string oid = path[idx];
    currentParam = getParam(oid);
    const StructInfo& si = T::getStructInfo();
    const FieldInfo& fi = si.findField(oid);
    next = reinterpret_cast<char*>(base) + fi.offset;
  } else {
    // its a number, so index into the object
    T* arr = reinterpret_cast<T*>(base);
    uint32_t = index = path[idx];
    next = arr[index];
  }
  if (path.length - 1 == idx) {
     // i.e. we're done
    return next;
  } else {
    currentParam->pointAt(next, path, idx+1);
  }
}
```

`pointAt` should be a protected member of `IParam`, we don't want client code calling it, but methods like `toProto` and `fromProto` will need to use it.

johndanenRV:
closes #239 

Reworked structInfo and param classes to allow for navigating to params via template specialization. This pull request also includes changes from ibc_2024 branch (mainly changes to Path).

There is still a bit more work to be done but this pull request is already large enough. Specifically, I need to add error checking for getParam and fromProto (issue #258) and add documentation for the new functions in this pull request.


mejohnnaylor (approved):
reviewed in live meeting.

PR #270, opened 2024-10-07
Added basic command support

### LINKED ISSUE #267: Add support for Commands
The connection manager gets a new RPC

The device model gets a new `Item` (command)

And we need a service to demo it. Suggest a new example under `connections/gRPC`

a good example would be a media transport with commands for: `stop`, `play`, `fast forward`, `rewind`, `pause`, `eject`, `load`, these would be buttons in a `.grid` file.

The `play` command would have a backing parameter (play speed), float, constrained to +/- 2.0
The `load` command would have a clip name parameter

The device should also have a param to indicate it's operating state `stopped`, `playing`, ... etc

and, of course  the `commands` object would need to be piped though `Device::toProto`...

johndanenRV:
closes #267 

Added basic command support to allow devices to define commands without streaming.


interface/param.proto
```diff
@@ -292,14 +292,17 @@ message UpdateSubscriptionsPayload {
  *   Param definition = 1; // provides command's input values
  *   bool respond = 2;     // true when the device will respond to the command, false otherwise.
  * }
- *
- * When streamed, only the value field will be used after the first message. */
+ * 
+ * The proceed flag allows the client to indicate to the server that is has already recieved a payload
+ * Ex. If a command steams a large file, the clent can set proceed to false to indicate that the server
```
mejohnnaylor, line 297: typo: "client"

sdks/cpp/common/include/IParam.h
```diff
@@ -34,11 +34,14 @@
...
-class Value; // forward reference
-class Param; // forward reference
-
 namespace common { 
 
+/**
+ * @brief EmptyValue is used for commands that do not have a value
+ */
+struct EmptyValue {};
+static EmptyValue emptyValue;
```
mejohnnaylor, line 43: this will cause emptyValue to be defined in every .cpp file that includes this header. better to do this:

IParam.h
``` .cpp
struct EmptyValue {};
extern EmptyValue emptyValue;
```

IParam.h:
``` .cpp
#include <IParam.h>
EmptyValue emptyValue;
```

```
```

sdks/cpp/connections/gRPC/examples/use_commands/use_commands.cpp
```diff
@@ -0,0 +1,258 @@
...
+int main(int argc, char* argv[])
+{
+    std::string addr;
+    absl::SetProgramUsageMessage("Runs the Catena Service");
+    absl::ParseCommandLine(argc, argv);
+  
+    addr = absl::StrFormat("0.0.0.0:%d", absl::GetFlag(FLAGS_port));
+
+    // commands should be defined before starting the RPC server 
+    defineCommands();
```
mejohnnaylor, line 253: try ... catch here, please.

sdks/cpp/lite/src/Device.cpp
```diff
@@ -88,6 +90,25 @@ std::unique_ptr<IParam> Device::getParam(const std::string& fqoid) const {
...
+        return nullptr;
+    }
+    if (path.front_is_string()) {
+        IParam* topCommand = getItem<common::CommandTag>(path.front_as_string());
+        path.pop();
+        if (!topCommand) {return nullptr;}
+        if (path.empty()) {
+            return topCommand->copy();
+        } else {
+            throw catena::exception_with_status("Subcommands not supported", catena::StatusCode::UNIMPLEMENTED);
```
mejohnnaylor, line 105: return nullptr instead

mejohnnaylor (changes requested):
see inline comments

mejohnnaylor (approved):

Issue #268, opened 2024-09-27
add support for menus


mejohnnaylor:
I thought we already had an issue for this, but here goes anyway...

- reuse the design pattern that we have for constraints - i.e. interface definition `IMenuGroup`, `IMenu` with concrete "lite" versions
- add support for passing this info to the client in `Device::toProto`
- create a device model with a few parameters in it
- add the parameter oids to a menu in each menu group

Note that DB will only render menu groups 0 (status) and 1 (config) see the documentation in opener software development guide v9.5 or later.

<img width="697" alt="image" src="https://github.com/user-attachments/assets/9d4836a0-a3c5-423d-8414-2d9e9096d61f">

PR #279, opened 2024-10-16
Feat-#268-bm

BeanJ4m:
Issue#268

Logic Check for menu support.

Codegen does not implement menus yet.


sdks/cpp/lite/src/Menu.cpp
```diff
@@ -0,0 +1,129 @@
...
+ * @author Ben Mostafa Ben.Mostafa@rossvideo.com
+ * @date 2024-10-10
+ * @copyright Copyright (c) 2024 Ross Video
+ */
+
+// lite
+#include <Menu.h>
+
+using namespace catena::lite;
+using catena::common::MenuGroupTag; // There is no MenuTag in the code
```
johndanenRV, line 30: The tags are defined in common/include/Tags.h

Take a look at the addItem and getItem functions in Device.h for an example of how they are used.

BeanJ4m: I added MenuTag and IMenu to Tags.h

PR #280, opened 2024-10-16
Finished implementing constraints

### LINKED ISSUE #211: codegen and constraints
see UML diagram of class hierarchy  here. https://lucid.app/lucidchart/f8e5c336-3c28-4f45-9844-6f8f8cb4d1bc/edit?invitationId=inv_4a6be56a-bf42-4a9f-ab54-262174f9b14c&page=RtlNJyFC5H4j#

`IConstraint` and `RangeConstraint` have been implemented, but not codegen'd on the branch called constraints.

let's take an example of a constraint

```json
"params": {
  "offset": {
    "type": "INT32",
    "name": {"display_strings": {"en": "offset"}},
    "value": {"int32_value": 0},
    "constraint": {
      "type": "INT_RANGE",
      "int32_range": {
        "min_value": -10,
        "max_value": 10,
        "step": 1
      }
   }
}
```

the generated code should resemble:

```cpp
using namespace catena;
std::unordered_map<std::string,common::IConstraint*> constraints;
NullConstraint nullConstraint; // implements Constraint but does nothing

// instantiated the parameter's value object
int32_t offset{0};

// constructor adds Range constraint to global constraints map
// name needs to be unique, it belongs to a param, so put the word "Param" in its name
RangeConstraint<int32_t> offsetParamConstraint {-10,10,1,-10,10,"/params/offset",constraints}; 

// Param constructor needs a new argument - pointer to IConstraint
lite::Param<int32_t> offsetParam{/* other arguments */, &offsetParamConstraint, /* other arguments */};

// N.B. for params without constraints, a pointer to nullConstraint should be passed


```

## example client code

```cpp
// probably in the SetValue gRPC
Param<int32_t>& aNumber = *dynamic_cast<Param<int32_t>*>(dm.getItem("/counter", Device::ParamTag{}));
if (aNumber.fromProto(msg)) {
  // happy path, the value was set within the constraint, return the ok code.
} else {
  // unhappy path, value not set, return an error code
}
```

## Constraints can be shared too

```json
{
  "constraints": {
    "sexagesimal": {
      "type": "INT_RANGE",
      "min_value": 0,
      "max_value": 59,
      "step": 1,
      "display_max": 60
    }
  }
}
```

would generate this cpp:

```cpp
// repeated from above
using namespace catena;
std::unordered_map<std::string,common::IConstraint*> constraints;
NullConstraint nullConstraint; // implements Constraint but does nothing

RangeConstraint<int32_t> sexagesimalConstraint {0,59,1,0,60, "/constraints/sexagesimal",constraints};

```

## possible code in Param's fromProto method, which now returns a success flag.
```cpp
bool fromProto(catena::Value& src) {
  bool happy = true;
  if (constraint_ != nullptr) {
    if (constraint_.isValid(src.get_value())) { // this line changes with the param's value type
      value_ = src.get_value();
    } else {
      happy = false;
    }
  }
  return happy;
}
```

### LINKED ISSUE #228: shared constraints
Add shared constraints to codegen and ref_oid support in json schema

### LINKED ISSUE #236: Constraints need to work with Arrays
a constraint's apply method should work for the array versions of int, float and string. 

Solution is likely adding constexpr conditions for each type

### LINKED ISSUE #253: DeviceRequest needs updating in the gRPC connection manager
Now that parameters support constraints, and we have shared constraints and language packs, the DeviceRequest RPC needs to pass these object thru as part of the device model to the client.

### LINKED ISSUE #263: use_constraints example is incomplete and out of date
intended purpose for this example is to demonstrate to an on-boarder how to use constraints.

It should, therefore:

- define a dm that includes one of each type of constraint (except ALARM_TABLE)
- deliver a GUI for DB to render using a `.grid` file
- demonstrate constraining behavior - i.e. setting an out of bounds value using Postman should preserve the value set within the constraints.

### LINKED ISSUE #272: Add commands to device toProto()
Commands need to be included in the DeviceRequest() response.

Also, some more examples should be added to use_commands.cpp showing how commands can require certain value types and respond with values of various types.

### LINKED ISSUE #278: Range constraints should enforce step size
This will be pretty simple to add for INT_RANGE. 

For FLOAT_RANGE we will need to decide how much precision is needed when constraining step size.

johndanenRV:
closes #211
closes #228
closes #236 
closes #263 
closes #278 
closes #253 
closes #272 

- Updated constraints to work with new from proto functions.
- Added constraint support for arrays
- Finished implementing shared constraints
- Added template support for constraints
- Finished codegen for constraints
- Added examples for all constraint types
- Added constraints and commands to DeviceRequest

### DIFF
--- a/sdks/cpp/common/include/IConstraint.h
+++ b/sdks/cpp/common/include/IConstraint.h
@@ -43,8 +43,7 @@ public:
 
     /**
      * @brief serialize the constraint to a protobuf message
-     * @param constraint the protobuf message to populate, NB, implementations should 
-     * dynamically cast this to catena::Constraint
+     * @param constraint the protobuf message to serialize to
      */
     virtual void toProto(catena::Constraint& constraint) const = 0;
 
--- a/sdks/cpp/common/include/IConstraint.h
+++ b/sdks/cpp/common/include/IConstraint.h
@@ -26,7 +26,9 @@
  * @copyright Copyright (c) 2024 Ross Video
  */
 
-#include "google/protobuf/message_lite.h" 
+// protobuf interface
+#include "interface/param.pb.h"
+#include "interface/constraint.pb.h" 
 
 namespace catena {
 namespace common {
@@ -36,8 +38,7 @@ namespace common {
  */
 class IConstraint {
 public:
-    IConstraint() : oid_{}, isShared_{} {}
-    IConstraint(std::string oid, bool shared) : oid_{oid}, isShared_{shared} {}
+    IConstraint() = default;
     virtual ~IConstraint() = default;
 
     /**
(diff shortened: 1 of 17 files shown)


mejohnnaylor (changes requested):

sdks/cpp/common/include/IConstraint.h
```diff
@@ -36,49 +38,46 @@ namespace common {
...
+     * @param src a catena::Value to check the constraint against
+     * @return true if the constraint is satisfied, false otherwise
      */
-    virtual void apply(void* src) const = 0;
+    virtual bool satisfied(const catena::Value& src) const = 0;
 
     /**
-     * @brief check if the constraint is shared
-     * @return true if the constraint is shared, false otherwise
+     * @brief applies constraint to src and writes constrained value to dst
```
mejohnnaylor, line 58: should be returns constrained value object

mejohnnaylor (dismissed):

PR #282, opened 2024-10-21
Added callback registry to business logic examples

### LINKED ISSUE #281: callback registry for business logic support
our example services don't really do much with the `valueSetByClient` event:

```cpp
        dm.valueSetByClient.connect([](const std::string& oid, const IParam* p, const int32_t idx) {
            // all we do here is print out the oid of the parameter that was changed
            // your biz logic would do something _even_more_ interesting!
            std::cout << "*** signal received: " << oid << " has been changed by client" << '\n';
        });
```

it'd be nice if the biz logic could use the `oid` to look up an event handler.

There's some subtlety about where the event handler lives in the `oid`'s hierarchy.

consider these params:

```json
"params": {
  "gain": {"type": "FLOAT", "value": {"float32_value": 10.0}},
  "eq": {
      "type":  "STRUCT",
     "params": {
        "response": {"type": "INT32"},
        "freq": {"type": "FLOAT"},
        "q_factor": {"type": "FLOAT"},
        "gain": {"type": "FLOAT"}
      },
      "value": {
          "struct_value": {
             "fields": {
                "response": { "value": {"int32_value": 1}},
                 "frequency": { "value": {"float32_value": 1000.0}},
                  "gain": { "value": {"float32_value": 1.0}},
                  "q_factor": { "value": {"float32_value": 1.0}}
            }     
         }
      }
  },
  "two_band_eq": {
      "type": "STRUCT_ARRAY",
      "template_oid": "/eq",
      "value": {
        "struct_values": [
           {
              "struct_value": {
                "fields": {
                   "response": { "value": {"int32_value": 1}},
                    "frequency": { "value": {"float32_value": 1000.0}},
                     "gain": { "value": {"float32_value": 1.0}},
                     "q_factor": { "value": {"float32_value": 1.0}}
                  }
               }
           },
           {
              "struct_value": {
                "fields": {
                   "response": { "value": {"int32_value": 2}},
                    "frequency": { "value": {"float32_value": 2000.0}},
                     "gain": { "value": {"float32_value": 2.0}},
                     "q_factor": { "value": {"float32_value": 2.0}}
                  }
               }
           },
        ]
     }
  }
}
```

if the `oid` for the event is `/gain` the search for the callback terminates after one iteration because the `path` object will be empty.

it gets tricker with `/eq/q_factor`. The filter coefficients to implement the eq depend on all the members of the structure - if any one of them changes, all of the coefficients need to be recalculated.

Therefore the search for the callback should terminate at `/eq` and the handler should be passed the whole eq value object, not just the q_factor part of it.

When it comes to array accesses, e.g. `/2_band_eq/2`, the callback handler should be located at `/2_band_eq`, and the array index should be passed into the handler.

There's also the case where the whole array has been updated, this probably requires a different handler, so we need 2 handlers at address `/2_band_eq` - one for when a single element has changed, another for when every element has changed.

johndanenRV:
closes #281 

Added updated status_update and structs_with_authz examples to show how a callback registry could be used to handle updates from clients. Also updated Path so that path objects are still valid after copy/move operations.

We should be good to merge develop into main after this pull request.

### DIFF
--- a/sdks/cpp/common/include/Path.h
+++ b/sdks/cpp/common/include/Path.h
@@ -141,9 +141,18 @@ class Path {
     Index front_as_index() const;
 
     /**
+     * @brief return a string representation of the Path
+     * @return std::string the path as a string
+     * 
+     * any popped segments are not included in the  returned string.
+     */
+    std::string toString() const;
+
+    /**
      * @brief return a fully qualified, albeit escaped oid
-     *
      * @return std::string
+     * 
+     * The fully qualified oid is rooted at the device and includes all popped segments.
      */
     std::string fqoid() const;
 
--- a/sdks/cpp/common/include/Path.h
+++ b/sdks/cpp/common/include/Path.h
@@ -100,9 +100,14 @@ class Path {
      *
      * @return number of segments
      */
-    inline Index size() const { return cend() - front_; }
+    inline Index size() const { return segments_.size() - frontIdx_; }
 
-    inline bool empty() const { return front_ == cend(); }
+    /**
+     * @brief return true if the Path is empty
+     *
+     * @return true if the Path is empty
+     */
+    inline bool empty() const { return frontIdx_ >= segments_.size(); }
 
     /**
      * @return true if the front of the path is a string, false if it's an Index or empty.
@@ -140,7 +145,7 @@ class Path {
      *
      * @return std::string
      */
-    std::string fqoid();
+    std::string fqoid() const;
 
     /**
      * @brief pop the front of the path.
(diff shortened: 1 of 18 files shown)


sdks/cpp/common/include/Path.h
```diff
@@ -100,9 +100,14 @@ class Path {
      *
      * @return number of segments
      */
-    inline Index size() const { return cend() - front_; }
+    inline Index size() const { return segments_.size() - frontIdx_; }
```
mejohnnaylor, line 103: I'm curious to know why you switched from using iterators to an index.

johndanenRV: There was a bug that whenever a Path was copied/moved the front_ iterator still pointed to the old location. Using an index instead of an iterator seemed like the easiest way to solve this problem.

mejohnnaylor: I saw you'd made a copy of a Path object somewhere and wanted to talk about that with you today. Looking forward to it!

mejohnnaylor (approved):

PR #289, opened 2024-10-28
Implemented shallow device request

### LINKED ISSUE #265: shallow version of DeviceRequest RPC
Our current implementation only provides the deep response.

This'll be a problem for larger models.

The shallow version should:

- deliver the top level of the `Device` i.e. `params`, `commands`, and all the other maps are there but empty.
- stream the contents of each map one at a time
- no need to get more fine-grained than this - i.e. stream top level params separately from their children - until someone proves it's needed

johndanenRV:
closes #265 

Added a coroutine that streams the device as DeviceComponents if shallow is set to true. I wasn't sure where the shallow flag is supposed to come from so, for now, I have hard coded DeviceRequest to always be shallow.

I made sure to add very thorough comments so that it hopefully makes sense to anyone who hasn't worked with coroutines before.

It is interesting to compare this to pull request #170 where I had previously implemented this same feature without coroutines. The logic is clearly much simpler now. 😃

### DIFF
--- a/sdks/cpp/connections/gRPC/src/ServiceImpl.cpp
+++ b/sdks/cpp/connections/gRPC/src/ServiceImpl.cpp
@@ -491,15 +491,24 @@ void CatenaServiceImpl::DeviceRequest::proceed(CatenaServiceImpl *service, bool 
                     grpc::Status errorStatus(grpc::StatusCode::INTERNAL, "illegal state");
                     writer_.Finish(errorStatus, this);
                     break;
-                }      
-                catena::DeviceComponent component{};
-                std::vector<std::string> clientScopes = service_->getScopes(context_);
-                {
-                    Device::LockGuard lg(dm_);
-                    component = serializer_->operator()();
+                } 
+                
+                try {     
+                    catena::DeviceComponent component{};
+                    std::vector<std::string> clientScopes = service_->getScopes(context_);
+                    {
+                        Device::LockGuard lg(dm_);
+                        component = serializer_->getNext();
+                    }
+                    status_ = serializer_->hasMore() ? CallStatus::kWrite : CallStatus::kPostWrite;
+                    writer_.Write(component, this);
+                } catch (catena::exception_with_status &e) {
+                    status_ = CallStatus::kFinish;
+                    writer_.Finish(Status(static_cast<grpc::StatusCode>(e.status), e.what()), this);
+                } catch (...) {
+                    status_ = CallStatus::kFinish;
+                    writer_.Finish(Status::CANCELLED, this);
                 }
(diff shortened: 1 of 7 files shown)


sdks/cpp/connections/gRPC/src/ServiceImpl.cpp
```diff
@@ -478,20 +478,37 @@ void CatenaServiceImpl::DeviceRequest::proceed(CatenaServiceImpl *service, bool
...
+                } 
+                
+                try {     
+                    catena::DeviceComponent component{};
+                    std::vector<std::string> clientScopes = service_->getScopes(context_);
+                    {
+                        Device::LockGuard lg(dm_);
+                        component = serializer_->getNext();
+                    }
+                    status_ = serializer_->hasMore() ? CallStatus::kWrite : CallStatus::kPostWrite;
```
mejohnnaylor, line 503: shouldn't this line be within the lock guard?

sdks/cpp/lite/include/Device.h
```diff
@@ -146,6 +147,140 @@ class Device {
      */
     void toProto(::catena::LanguageList& list) const;
 
+    /**
+     * @brief DeviceSerializer is a coroutine that serializes the device into a stream of DeviceComponents
+     * This struct manages the state and lifetime of the coroutine. It also provides the interface for resuming the coroutine.
+     */
+    struct DeviceSerializer {
```
mejohnnaylor, line 154: this is  a struct with private attributes - why isn't it a class?

sdks/cpp/lite/include/Device.h
```diff
@@ -146,6 +147,140 @@ class Device {
...
+         * member functions that the compiler will call when the coroutine is created, destroyed,
+         * or resumed.
+         */
+        struct promise_type {
+
+            /**
+             * get_return_object() is called when the coroutine is created (i.e. when getComponentSerializer() 
+             * is called). It creates the handle for the coroutine and returns the DeviceSerializer object.
+             */
+            DeviceSerializer get_return_object() {
```
mejohnnaylor, line 183: should be declared inline or defined in a .cpp file, same comment applied to other methods defined here. recommend inlining them as they're all v short.

mejohnnaylor (approved):

sdks/cpp/connections/gRPC/src/ServiceImpl.cpp
johndanenRV: No, even if something else changes the device it shouldn't affect how many more components need to be sent.

mejohnnaylor (approved):

Issue #291, opened 2024-11-04
simplify the c++ SDK's build


mejohnnaylor:
today we have 2 lists to control the build:

- catena model - can be either full or lite
- connections - can be gRPC, REST or both, and may extend to include WSS

I think we can simplify this to:

- always make the lite version of the device model - this is the thing that the business logic and connection manager interact with
- keep the lite examples their intent of teaching on boarders to integrate their biz logic with the device model is still valid
- when making the REST connection manager, create the `full` version of the `catena_interface` but do not include the gRPC service definition.
- the definition of EmptyRequest should move into `language.proto` although I struggle with why we even need it

| item | gRPC | REST |
------|-------|------|
device model | lite | lite
connection manager | catena_interface_lite | catena_interface_full |
biz logic | lite | lite |
catena_interface lib | lite runtime, plus service definition | full runtime, minus service definition |

should be what we need

We can address Issue #290 at the same time.

mejohnnaylor:
done :-)

Issue #264, opened 2024-09-27
we should define a REST API


mejohnnaylor:
some ambitions around this...

- use JSON as the serialization method
- implies we need to build off of the full version of the proto files
- use OpenAPI to define / document the interface
- use SSE (server sent events) for RPCs that expect a stream of replies - i.e. `Connect`, `DeviceRequest`

Chat GPT say:

Server-Sent Events (SSE) over HTTP/1.1 provides a way for servers to push updates to clients over a single, long-lived HTTP connection. It's commonly used for real-time updates like notifications, live feeds, or any other kind of streaming data.

### Key Components of SSE over HTTP/1.1:

1. **HTTP Headers**: 
   The server must respond with the following HTTP headers to initiate the SSE connection:
   ```http
   Content-Type: text/event-stream
   Cache-Control: no-cache
   Connection: keep-alive
   ```

   - `Content-Type: text/event-stream`: This tells the client to expect an SSE stream.
   - `Cache-Control: no-cache`: Ensures that the connection is not cached, allowing it to remain open.
   - `Connection: keep-alive`: Keeps the connection alive for as long as possible.

2. **Event Stream Format**: 
   The data sent over SSE is plain text, but it follows a specific format where events are represented by lines of text. Each event is composed of multiple fields:
   - `event`: Specifies the type of event (optional).
   - `data`: The actual data being sent (required).
   - `id`: A unique identifier for the event (optional, useful for reconnections).
   - `retry`: A reconnection time in milliseconds (optional).

   Fields are separated by newline characters, and each event is terminated by two consecutive newline characters. For example:

   ```
   id: 1
   event: message
   data: This is the first message.

   id: 2
   event: message
   data: This is another message.

   ```

3. **Client-Side Handling**: 
   On the client side, SSE is supported natively by JavaScript through the `EventSource` API. To establish an SSE connection, you simply create an instance of `EventSource`, pointing it to the URL where the server is serving the event stream.

   Example JavaScript client:
   ```javascript
   const eventSource = new EventSource('/sse-endpoint');

   eventSource.onmessage = function(event) {
       console.log('Received data:', event.data);
   };

   eventSource.onerror = function(error) {
       console.error('Error:', error);
   };
   ```

### Example SSE HTTP Response

Here is a full example of an SSE response from the server over HTTP/1.1:

```http
HTTP/1.1 200 OK
Content-Type: text/event-stream
Cache-Control: no-cache
Connection: keep-alive

id: 1
event: message
[...]

PR #295, opened 2024-11-06
Feat #264 2 jrn

mejohnnaylor:
in the spirit of keeping things incremental, I'd like to get this into the repo before tackling #291.
This just adds crow as a submodule and builds a super simple REST server


mejohnnaylor:
I added some documentation - and removed the old page about using native types because that's now implemented.

johndanenRV (approved):

Issue #307, opened 2024-11-13
complete migration of old lite examples to common directory


mejohnnaylor:

mejohnnaylor:
- `import_params`
- `use_struct_arrays`
- `use_structs`
- `use_templates`

the job is mainly to replace all references to the `catena::lite` namespace with `catena::common` and to screen `lite` out of any comments.

N.B. `start_here` is not in the list because it was renamed to `hello_world` upon migration.

then delete the `lite` and `full` branches with `git rm -rf`

PR #313, opened 2024-11-19
Feat-#307-bm

BeanJ4m:
Moved lite/examples to common/examples and removed Lite and Full folders


johndanenRV (changes requested):
You probably shouldn't delete full/tests yet so that you can use it as an example for implementing gtest.

johndanenRV (approved):

johndanenRV (approved):

PR #334, opened 2024-12-10
Added scope checking when authorization is enabled

### LINKED ISSUE #175: Check access scopes in toProto and fromProto
Access scopes should be checked before accessing params and sub-params with to/fromProto.

johndanenRV:
closes #175 

Added scope checking for all types. Request without the required access scopes will be blocked. If writing to a struct, if the struct value has an unauthorized sub-param then only that sub-param is ignored and the rest of the write operation succeeds.

### DIFF
--- a/sdks/cpp/common/include/Authorization.h
+++ b/sdks/cpp/common/include/Authorization.h
@@ -66,9 +66,24 @@ class Authorizer {
     Authorizer(const std::vector<std::string>& clientScopes)
         : clientScopes_{clientScopes} {}
 
+    /**
+     * @brief Authorizer does not have copy semantics
+     */
     Authorizer(const Authorizer&) = delete;
+
+    /**
+     * @brief Authorizer does not have copy semantics
+     */
     Authorizer& operator=(const Authorizer&) = delete;
+
+    /**
+     * @brief Authorizer has move semantics
+     */
     Authorizer(Authorizer&&) = default;
+
+    /**
+     * @brief Authorizer has move semantics
+     */
     Authorizer& operator=(Authorizer&&) = default;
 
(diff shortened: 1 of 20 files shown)


sdks/cpp/common/include/Authorization.h
```diff
@@ -50,64 +51,57 @@ namespace common {
...
+     * @brief Construct a new Authorizer object
      * @param pd the ParamDescriptor of the object
      * @param scope the scope of the object
      */
-    AuthzInfo(const ParamDescriptor& pd, const std::string& scope)
-      : pd_{pd}, clientScope_{scope}{}
+    Authorizer(const std::vector<std::string>& clientScopes)
+        : clientScopes_{clientScopes} {}
+
+    Authorizer(const Authorizer&) = delete;
```
mejohnnaylor, line 69: please add doxygen comments to say that Authorizer doesn't have copy semantics, and that Authorizer does have move semantics.

sdks/cpp/common/include/Device.h
```diff
@@ -138,7 +134,7 @@ class Device {
      */
     inline DetailLevel_e detail_level() const { return detail_level_; }
 
-    inline std::string getDefaultScope() const { return default_scope_.toString(); }
+    inline const std::string& getDefaultScope() const { return default_scope_.toString(); }
```
mejohnnaylor, line 137: does this return a reference to a deleted object? IDK but it depends on the implementation of default_scope_. please check

mejohnnaylor (changes requested):
just a few comments needed. thx!

mejohnnaylor (approved):

Issue #368, opened 2025-01-06
complete doxygen for all classes and namespaces


mejohnnaylor:
The list of classes (and namespaces etc) [here](https://rossvideo.github.io/Catena/doxygen/annotated.html) should have a descriptive entry in the 2nd column for every entry in the 1st.

The namespaces (catena, catena::common, etc) only need to be doxygenated once, not every time they appear.

PR #371, opened 2025-01-13
Doc #368 bw

BenWhitten-RV:
Added half the missing doxygen briefs, other half to be done by Zuhayr.

### DIFF
--- a/sdks/cpp/common/include/Authorization.h
+++ b/sdks/cpp/common/include/Authorization.h
@@ -47,11 +47,13 @@
 #include <functional>
 
 /**
- * @brief A collection of classes used in Catena
+ * @brief Functionality at this scope includes the protoc generated classes.
+ * Most everything else if in child namespaces such as common, meta, etc.
  */
 namespace catena {
 /**
- * @brief Functionality that can be shared between the different connection management types
+ * @brief Functionality that can be shared between the different connection
+ * management types
  */
 namespace common {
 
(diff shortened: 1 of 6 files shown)


sdks/cpp/common/include/Authorization.h
```diff
@@ -46,8 +46,13 @@
 
 #include <functional>
 
-
+/**
+ * @brief A collection of classes used in Catena
```
mejohnnaylor, line 50: better would be: top level namespace for catena. Functionality at this scope includes the protoc generated classes. Most everything else is in child namespaces of common, gRPC, REST, etc. 

needs fact checking!

sdks/cpp/common/include/Device.h
```diff
@@ -38,6 +38,10 @@
  * @copyright Copyright (c) 2024 Ross Video
  */
 
+/** @example hello_world.cpp Demonstrates creation of a simple device model,
```
mejohnnaylor, line 41: /** on its own line, please

sdks/cpp/common/include/Fake.h
```diff
@@ -29,7 +29,7 @@
 namespace catena {
 namespace common {
 /**
- * a fake lock for use in recursive function calls
+ * @brief A fake lock for use in recursive function calls
```
mejohnnaylor, line 32: this isn't used anymore, is there a way to mark something as deprecated in doxygen?

mejohnnaylor (changes requested):
good first PR!

mejohnnaylor (approved):

mejohnnaylor (approved):

Issue #327, opened 2024-12-06
Implement MultiSetValue RPC


mejohnnaylor:
this should be very similar to SetValue

it may even behoove us to make SetValue a special case of MultiSetValue (or the other way round)

PR #382, opened 2025-01-20
Enh #327 bwzs

BenWhitten-RV:
MultiSetValue gRPC command enhancement.
- Took code from SetValue and created GenericSetValue.cpp/.h to handle all setValue requests by looking through a MultiSetValuePayload..
- Created MultiSetValue.cpp/.h to work with GenericSetValue.
- Overhauled original SetValue.cpp/.h to work with GenericSetValue (using a MultiSetValyePayload containing a single request).


BenWhitten-RV:
Making changes before making another request

PR #383, opened 2025-01-21
Enh #327 bwzs

BenWhitten-RV:
Made changes discussed during meeting
- Rearranged for loop so we aren't checking the authorization and user scope every time.
- Removed GenericSetValue to instead have MultiSetValue and child class SetValue.


sdks/cpp/connections/gRPC/src/MultiSetValue.cpp
```diff
@@ -0,0 +1,162 @@
...
+         * kFinish and notifying the responder once finished.
+         */
+        case CallStatus::kProcess:
+            // Used to serve other clients while processing.
+            create(service_, dm_, ok);
+            context_.AsyncNotifyWhenDone(this);
+            try {
+                catena::exception_with_status rc{"", catena::StatusCode::OK};
+                // If authorization is enabled, check the client's scopes.
+                catena::common::Authorizer* authz;
```
mejohnnaylor, line 110: you probably want to use a std::shared_ptr for authz and initialize it with make_shared().
read up on RAII (resource acquisition is initialization) as the guiding principle behind this.

sdks/cpp/connections/gRPC/src/MultiSetValue.cpp
```diff
@@ -0,0 +1,162 @@
...
+                if (service->authorizationEnabled()) {
+                    std::vector<std::string> clientScopes = service->getScopes(context_);
+                    catena::common::Authorizer authorizer{clientScopes};
+                    authz = &authorizer;
+                } else {
+                    authz = &catena::common::Authorizer::kAuthzDisabled;
```
mejohnnaylor, line 116: you don't want whatever you use to remember kAuthzdisabled to try to destroy it. Have a think about how to square this requirement with the one above.

sdks/cpp/connections/gRPC/src/MultiSetValue.cpp
```diff
@@ -0,0 +1,162 @@
...
+                 * Looping through each request. In case of SetValue, reqs_ is
+                 * a MultiSetValuePayload containing a single request.
+                 */
+                for (auto &setValuePayload : reqs_.values()) {
+                    std::string oid = setValuePayload.oid();
+                    catena::Value value = setValuePayload.value();
+                    Device::LockGuard lg(dm_);
+                    // Taking rc if it is OK, otherwise saving error rc.
+                    if (rc.status == catena::StatusCode::OK) {
+                        rc = dm_.setValue(oid, value, *authz);
```
mejohnnaylor, line 128: thinking on, the entire multisite should succeed or fail, not something between the two. this provides the benefit of clarity and idempotent interactions to clients. One way to achieve this would be to add a method:

catena::exception_with_status setValueTry(// args);  


which does a dry run to set the value and returns the appropriate error code.

If there's a single failure, terminate the RPC.

If there a no failures, proceed to the value-setting stage of the RPC like you've implemented here - and no need to check the return code of every call to setValue.

mejohnnaylor (changes requested):

sdks/cpp/connections/gRPC/src/MultiSetValue.cpp
BenWhitten-RV: Used shared ptr declared outside of if-else to maintain ownership while using .get() to assign it to the raw ptr authz. By doing this, we never set kAuthzDisabled to a shared ptr and as such it doesn't attempt to delete the object once out of scope.

sdks/cpp/connections/gRPC/src/MultiSetValue.cpp
BenWhitten-RV: Created setValueTry in device.cpp/h and an additional for loop before we set the values to make sure we have proper authority to make all changes.

BenWhitten-RV:
Latest push implements the changes flagged above.

sdks/cpp/connections/gRPC/src/MultiSetValue.cpp
```diff
@@ -0,0 +1,158 @@
...
+                        break;
+                    }
+                }
+                // If above is valid, then proceed with setting the values.
+                if (rc.status == catena::StatusCode::OK) {
+                    for (auto &setValuePayload : reqs_.values()) {
+                        std::string oid = setValuePayload.oid();
+                        catena::Value value = setValuePayload.value();
+                        Device::LockGuard lg(dm_);
+                        dm_.setValue(oid, value, *authz);
```
mejohnnaylor, line 127: because we've done a pass already with authz, we can do the actual value-setting with authz disabled. be sure to comment that clearly.

mejohnnaylor (changes requested):

sdks/cpp/connections/gRPC/src/MultiSetValue.cpp
BenWhitten-RV: Disabled authz when setting values and reasoning in comment before for loop.

mejohnnaylor (approved):

PR #397, opened 2025-02-03
Implemented AddLanguage gRPC

BenWhitten-RV:
Title. Additionally overhauled authorizer to use helper function hasAuthz which checks if a client has the specified authorization.


sdks/cpp/common/include/Device.h
```diff
@@ -406,6 +418,8 @@ class Device {
     std::unordered_map<std::string, common::IMenuGroup*> menu_groups_;
     std::unordered_map<std::string, IParam*> commands_;
     std::unordered_map<std::string, common::ILanguagePack*> language_packs_;
+    // Required to maintain ownership of added language packs.
+    std::unordered_map<std::string, std::shared_ptr<common::ILanguagePack>> added_packs_;
```
mejohnnaylor, line 422: I'd expected you to add the packs to the device model's language_packs object. Why did you decide to use a parallel map for added packs?

mejohnnaylor (changes requested):

sdks/cpp/common/include/Device.h
BenWhitten-RV: The language packs are added to the device model's lanugage_packs_ object, however simply adding them to the unordered_map does not retain ownership causing the added language_pack to go out of scope. The easiest fix was to just have a map of shared pointers to keep ownership for as long as the device is active.

Another solution I looked at was changing language_packs_ to instead store unique_ptrs to ILanguagePacks, but as far as I could tell that would require changes to the json.cpp generation and the functions it calls when initializing LanguagePacks.

If you would rather the latter I can do that, I just figured it would be best to try the easier option to start.

sdks/cpp/common/src/Device.cpp
```diff
@@ -80,6 +80,24 @@ catena::exception_with_status Device::getValue (const std::string& jptr, catena:
...
+    } else {
+        auto& name = language.language_pack().name();
+        auto& id = language.id();
+        // Making sure LanguagePack is properly formatted.
+        if (name.empty() || id.empty()) {
+            return catena::exception_with_status("Invalid language pack", catena::StatusCode::INVALID_ARGUMENT);
+        }
+        // added_packs_ here to maintain ownership in device scope.
+        added_packs_[id] = std::make_shared<LanguagePack>(id, name, LanguagePack::ListInitializer{}, *this);
+        language_packs_[id]->fromProto(language.language_pack());
```
mejohnnaylor, line 96: There should also be a ComponentLanguagePack emitted as a PushUpdate to all connected clients. Look at how SetValue generates Value messages to connected clients for the pattern on how to do this.

BenWhitten-RV: Implemented a signal LanguageAddedPushUpdate which emits a ComponentLanguagePack at end of AddLanguage. **Changes have not been pushed yet** as I am waiting until RequestLanguagePack is merged so the code to populate a ComponentLanguagePack can be reused.

BenWhitten-RV:
Do not pull latest push, explanation in comment above.

BenWhitten-RV:
All changes implemented and tested, should be good to merge with dev once reviewed.

mejohnnaylor (approved):

PR #409, opened 2025-02-05
Extended functionality of path 

ZuhayrSarkerRV:
- Added member initializers to correctly create default constructor for new implementation
- Added leading_slash parameter to toString that is false by default
- Added push_back method for building paths incrementally 
- Added ValidType concept for type-safe construction and push_back
- Added variadic constructor for path creation

### DIFF
--- a/sdks/cpp/common/include/Path.h
+++ b/sdks/cpp/common/include/Path.h
@@ -14,7 +14,7 @@
 
  3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
 
- THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS “AS IS” AND ANY EXPRESS OR IMPLIED WARRANTIES, 
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, 
  INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, 
  INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER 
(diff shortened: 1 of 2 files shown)


sdks/cpp/common/include/Path.h
```diff
@@ -37,6 +37,10 @@
 
 namespace catena {
 namespace common {
+
+template<typename T>
+concept ValidType = std::same_as<T, size_t> || std::same_as<T, int> || std::convertible_to<T, std::string>;
```
mejohnnaylor, line 42: nice building on the pseudo code in the issue - I didn't think I could use std::string& but wrote my intention rather than bothering to get the code perfect - that's what you're for ;-)

sdks/cpp/common/include/Path.h
```diff
@@ -143,11 +156,12 @@ class Path {
 
     /**
      * @brief return a string representation of the Path
+     * @param leading_slash if true, include leading '/' in output (defaults to true)
```
mejohnnaylor, line 159: defaults to false, actually

sdks/cpp/common/include/Path.h
```diff
@@ -96,6 +100,15 @@ class Path {
      */
     explicit Path(const char* literal);
 
+    /**
+     * @brief Construct a Path from a sequence of segments
+     * @param args Variable number of arguments that are either Index or string types
+     */
+    template<ValidType... Args>
+    explicit Path(Args... args) : segments_{}, frontIdx_{0} {
+        (push_back(std::forward<Args>(args)), ...);
```
mejohnnaylor, line 109: the string arguments need to be escaped using the `escape()` method. It's possible to have a slash inside of a string: `like/this` which gets escaped to `like~0this` (or it may be `like~1this`, I forget). Twiddles get escaped to the complement of whatever slashes are - i.e. `like~1this` if slashes are escaped to ~0. 

so, you can't just push them onto the back of segments

actually - this comment makes more sense for the `push_back` method - it should be doing the escaping, so apply it there.

sdks/cpp/common/include/Path.h
```diff
@@ -179,10 +193,30 @@ class Path {
...
+    template<ValidType T>
+    void push_back(T segment) {
+        if constexpr (std::same_as<T, Index> || std::same_as<T, int>) {
+            segments_.emplace_back(std::in_place_type<Index>, static_cast<Index>(segment));
+        } else {
+            std::string str{segment};
+            if (str == "-") {
+                segments_.emplace_back(std::in_place_type<Index>, kEnd);
+            } else {
+                escape(str);
```
mejohnnaylor, line 209: I only just noticed this - disregard the comment about having to escape things because you already did.

sdks/cpp/common/include/Path.h
```diff
@@ -179,10 +193,30 @@ class Path {
...
+            }
+        }
+    }
+
+
   private:
     using Segments = std::vector<Segment>;
-    Segments segments_; /**< the path split into its components */
-    std::size_t frontIdx_; /**< index of current front of path */
+    Segments segments_{}; /**< the path split into its components */
```
mejohnnaylor, line 218: this is better than the code it replaces. thank you !

mejohnnaylor (approved):

sdks/cpp/common/include/Path.h
ZuhayrSarkerRV: Sorry, I had it default to true before, just noticed this was a mistake. Forgot to update the comment. I will add it in the next push to develop.

PR #412, opened 2025-02-06
Apply limits to array sizes.

BenWhitten-RV:
Title. Below is a changelog:

- SetValue now ensures string and array length falls within param's max_length value.
- SetValue .../-, which appends to the end of an array, now returns an error if the add would cause the array to exceed the param's max_length value.
- Added param_default_length to device. This governs the max_length for the device's params without the field defined. Default is 1024.
- If max_length is an invalid value (<= 0), then it is set to 1024
- Added set_param_default_max() function to device. Intended used in business logic to set param_default_length.
- Added shared flag for gRPC device's FLAGS_param_default_max. Can be implemented in business logic to allow the cmd line arg param_default_max=# to set the device's param_default_length.

Logic behind keeping the default param length in device is to give the client the choice on whether they allow it to be set through cmd or not. This also allows them to set a constant value of their own if they choose.

GetParam already does not allow array accesses outside of its size, so nothing changed there.

This pull only implements max_length. String arrays, which have an additional field total_length, are still subject to DoS attacks. This is to be done at a later date.


sdks/cpp/common/include/Device.h
```diff
@@ -66,6 +66,8 @@
 #include <type_traits>
 #include <coroutine>
 
+#define DEFAULT_PARAM_MAX 1024
```
mejohnnaylor, line 69: using `#define` is not modern practice which is to do something like `constexpr uint32 kDefaultMaxArraySize{1024};` please switch to this. It should be within `catena::common` scope too.

sdks/cpp/common/include/Device.h
```diff
@@ -104,7 +106,8 @@ class Device {
     Device(uint32_t slot, Device_DetailLevel detail_level, std::vector<std::string> access_scopes,
       std::string default_scope, bool multi_set_enabled, bool subscriptions)
       : slot_{slot}, detail_level_{detail_level}, access_scopes_{access_scopes},
-      default_scope_{default_scope}, multi_set_enabled_{multi_set_enabled}, subscriptions_{subscriptions} {}
+      default_scope_{default_scope}, multi_set_enabled_{multi_set_enabled},
+	  subscriptions_{subscriptions}, param_default_max_{DEFAULT_PARAM_MAX} {}
```
mejohnnaylor, line 110: I don't understand why Device needs a `param_default_max_` attribute unless you were going to make a command-line attribute `default-max-array-size` with a default of the constexpr, then it'd make sense.  I also would prefer then name to be `default_max_length` to match the name of the field in the Param object.

sdks/cpp/common/include/Device.h
```diff
@@ -137,6 +140,20 @@ class Device {
...
+	 * @param default_max The default max length for params.
+	 */
+    void set_param_default_max(const int default_max) {
+		if (default_max > 0) {param_default_max_ = default_max;}
+    }
+
+	/**
+	 * @return The default max length for this device's params.
+	 */
+    inline const int param_default_max() const { return param_default_max_;}
```
mejohnnaylor, line 155: the function type doesn't need to be `const int` you should leave it to the caller to choose whether they want their `int` to be const. It's returned by making a copy, so the internal attribute value won't change. Also array indices should be `uint32_t`, not signed ints.

sdks/cpp/common/include/ParamWithValue.h
```diff
@@ -126,6 +126,43 @@ class ParamWithValue : public catena::common::IParam {
...
+     * someone could still mount an attack on the system by sending a
+     * string_array with a large value.
+     * @param value The value to validate the size of.
+     */
+    const bool validateSize(const catena::Value& value) {
+        /**
+         * Not really a better way since all value types have different
+         * function names for retrieving the value.
+        */ 
+		switch(value.kind_case()) {
```
mejohnnaylor, line 142: whitespace issue here

mejohnnaylor: I'd prefer the implementation to be a map of whatever type kind_case is to functions like this:

``` 
const std::unordered_map<kind, std::function<bool(const Catena::Value &value, uint32_t max_len)>> funcMap {
  {kStringValue, []() {return value.string_value().length() < max_len;}},
  {kInt32ArrayValue, []() {return value.int32_array_values().ints_size() < max_len;}}
  /// etc
};

const auto kc = value.kind_case();
return funcMap.contains(kc) ? funcMap[kc](value, max_len) : true;
```

rationale - switch statements are hard to maintain and can be error prone with unintended fall thrus etc

sdks/cpp/common/include/ParamWithValue.h
```diff
@@ -126,6 +126,43 @@ class ParamWithValue : public catena::common::IParam {
...
     virtual ~ParamWithValue() = default;
+    
+    /**
+     * @brief Validates the size of a string or array value.
+     * This does not check size of all elements in a string array. Therefore
+     * someone could still mount an attack on the system by sending a
+     * string_array with a large value.
+     * @param value The value to validate the size of.
+     */
+    const bool validateSize(const catena::Value& value) {
```
mejohnnaylor, line 137: doesn't need to be const unless you're returning a reference

mejohnnaylor (changes requested):
I didn't get all the way thru but there's plenty to chew on here ;-)

sdks/cpp/common/include/Device.h
BenWhitten-RV: Removed #define in place of just setting the default_max_length at the bottom of device.h as it's the only place the number really shows up aside from SharedFlags.cpp.

sdks/cpp/common/include/Device.h
BenWhitten-RV: See comment above.

sdks/cpp/common/include/Device.h
BenWhitten-RV: Removed unnecessary const and changed to uint32_t.

sdks/cpp/common/include/ParamWithValue.h
BenWhitten-RV: Added unordered map seen below which gets the size a value.
#`using Kind = catena::Value::KindCase;
    /**
     * @brief A map which returns a value's length using a function determined
	 * by the indexed Value.kind_case().
	 * Private for now, might make public later if useful in other scopes.
     */
    const std::unordered_map<Kind, std::function<uint32_t(const catena::Value& value)>> getSize {
        {Kind::kStringValue,              [](const catena::Value& value) {return value.string_value().length();}},
        {Kind::kInt32ArrayValues,         [](const catena::Value& value) {return value.int32_array_values().ints_size();}},
        {Kind::kFloat32ArrayValues,       [](const catena::Value& value) {return value.float32_array_values().floats_size();}},
        {Kind::kStringArrayValues,        [](const catena::Value& value) {return value.string_array_values().strings_size();}},
        {Kind::kStructArrayValues,        [](const catena::Value& value) {return value.struct_array_values().struct_values_size();}},
        {Kind::kStructVariantArrayValues, [](const catena::Value& value) {return value.struct_variant_array_values().struct_variants_size();}}
    };`

Opted for this instead of the suggested above because this will also prove useful when checking a string array's total size in the future.
Validate function now looks like this:
`bool validateSize(const catena::Value& value) const override {
        return getSize.contains(value.kind_case()) ? getSize.at(value.kind_case())(value) <= descriptor_.max_length() : true;
    }`

sdks/cpp/common/include/ParamWithValue.h
BenWhitten-RV: Removed unnecessary const

sdks/cpp/common/include/Device.h
```diff
@@ -410,7 +425,8 @@ class Device {
     std::string default_scope_;
     bool multi_set_enabled_;
     bool subscriptions_;
-    
+    int default_max_length_ = 1024;
```
mejohnnaylor: we agreed that this should come from the flag.

sdks/cpp/common/src/Device.cpp
```diff
@@ -54,6 +54,9 @@ catena::exception_with_status Device::setValueTry (const std::string& jptr, Auth
             if (!authz.writeAuthz(*param)) {
                 return catena::exception_with_status("Not authorized to write to param", catena::StatusCode::PERMISSION_DENIED);
```
mejohnnaylor, line 70: I'd prefer you to set `ans = ...` to avoid having multiple returns from the method. it's easier to understand code which only has one return.  Same comment for the other if test here.

mejohnnaylor (changes requested):

sdks/cpp/common/src/Device.cpp
```diff
@@ -44,15 +44,33 @@
 
 using namespace catena::common;
 
-catena::exception_with_status Device::setValueTry (const std::string& jptr, Authorizer& authz) {
+/**
+ * Flag used to set default_max_length_. Moved from SharedFlags as you run into
+ * compilation issues otherwise.
+ */
+#include "absl/flags/flag.h"
+ABSL_FLAG(uint32_t, default_max_array_size, 1024, "use this to define the default max length for array and string params.");
```
mejohnnaylor, line 52: what compile issues?

mejohnnaylor (changes requested):

sdks/cpp/common/include/Device.h
```diff
@@ -66,6 +66,9 @@
 #include <type_traits>
 #include <coroutine>
 
+#include "absl/flags/flag.h"
+#include <../../connections/gRPC/include/SharedFlags.h>
```
mejohnnaylor, line 70: this tells me why we/you had the link problem. unfortunately, this makes `Device.h` which is part of the `common` namespace dependent on an object in the`connection` namespace which is bad because it creates a circular dependency. `connection` depends on `common` which depends on `connection` which ....

I'm afraid the correct way to fix this is to define the `constexpr uint32_t kDefaultMaxArrayLength{1024}` in the `common` namespace, and then use it in the flag declaration in the `connection` namespace. You'll have to back out all the changes that made `common/examples` dependent on the flags too.

I'm sorry I didn't catch this.

mejohnnaylor (changes requested):

mejohnnaylor (approved):

Issue #403, opened 2025-02-05
Extend functionality of `common::Path` to build from segments


mejohnnaylor:
The `Path` class currently has a one-way design - give it a json pointer, and it'll split it into segments that can then be popped & un-popped.

To make it 2-way, we need to add functionality that will allow the segments to be added as separate strings in the constructor, or built incrementally using a `push_back` method.

```cpp
// Define a concept to constrain argument types
template<typename T>
concept ValidType = std::same_as<T, Path::Index> || std::same_as<T, std::string&>;

// variadic constructor
// escapes each argument and converts '-' to kEnd then adds them to the segments storage in order.
template<ValidType... Args)>>
Path(Args... args);

// push a segment
// escapes the segment, substitutes kEnd for '-' and adds to the back of the segment storage
template<ValidType T>
push_back(T segment);
```

example use:

```cpp
Path path_1{"root", 1, "child", "-"};
std::cout <<path_1.toString() << std::ends; /// "root/1/child/-"

Path path_2();
path_2.push_back("root");
path_2.push_back(1);
path_2.push_back("child");
path_2.push_back(kEnd);
std::cout << path_2.toString() << std::endl; // "root/1/child/-"
```

The documentation for the toString should be improved to say that it returns a headless json-pointer built from the segments. by headless we mean no leading /

mejohnnaylor:
maybe give `toString` a default boolean argument that determines whether a leading `/` is added to the return value.

PR #413, opened 2025-02-06
Added BasicParamInfoRequest with top-level param functionality only

ZuhayrSarkerRV:
- Added functionality to fetch top-level parameters, the syntax in postman is as follows: 
``` javascript
{
    "slot": 0,

    // Identifies starting parameter relative to device.params
    "oid_prefix": "", 
    
    //Selects whether child params should also be returned
    "recursive": false 
}
```
- Program flow mostly follows typical state machine system observed in other RPCs, note that it is currently setup to send responses in a stream.
- Fixed the comment in Path.h where it listed the wrong default value
- Bug(?): It doesn't go to kFinish for some reason I have yet to figure out. It still completes the request, just doesn't exit it as cleanly as I'd hope. I will return to this since it isn't breaking anything that I know of (yet).
- There are also still some print statements in places, I may remove them later on when I have finished the RPC

### DIFF
--- a/sdks/cpp/common/include/ParamWithValue.h
+++ b/sdks/cpp/common/include/ParamWithValue.h
@@ -17,7 +17,7 @@
  * contributors may be used to endorse or promote products derived from this
  * software without specific prior written permission.
  *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS “AS IS”
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
  * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
  * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
  * RE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
@@ -167,6 +167,21 @@ class ParamWithValue : public catena::common::IParam {
     }
 
     /**
+     * @brief serialize the parameter descriptor to protobuf
+     * include both the descriptor and the value
+     * @param paramInfo the protobuf value to serialize to
+     * @param authz the authorization information
+     */
+    catena::exception_with_status toProto(catena::BasicParamInfoResponse& paramInfo, Authorizer& authz) const override {
+        if (!authz.readAuthz(*this)) {
+            return catena::exception_with_status("Param does not exist", catena::StatusCode::INVALID_ARGUMENT);
+        }
+        descriptor_.toProto(*paramInfo.mutable_info(), authz);
+        return catena::exception_with_status("", catena::StatusCode::OK);
+    }
+
+
+    /**
      * @brief deserialize the parameter value from protobuf if authorized
      * @param value the protobuf value to deserialize from
      * @param clientScope the client scope
@@ -253,6 +268,14 @@ class ParamWithValue : public catena::common::IParam {
      */
     catena::CommandResponse executeCommand(const catena::Value& value) const override {
         return descriptor_.executeCommand(value);
+    }
+
+    /**
+     * @brief get the descriptor of the parameter
+     * @return the descriptor of the parameter
+     */
+    const ParamDescriptor& getDescriptor() const override {
+        return descriptor_;
     }
 
   private:
--- a/sdks/cpp/common/include/ParamWithValue.h
+++ b/sdks/cpp/common/include/ParamWithValue.h
@@ -174,7 +174,7 @@ class ParamWithValue : public catena::common::IParam {
      */
     catena::exception_with_status toProto(catena::BasicParamInfoResponse& paramInfo, Authorizer& authz) const override {
         if (!authz.readAuthz(*this)) {
-            return catena::exception_with_status("Param does not exist", catena::StatusCode::INVALID_ARGUMENT);
+            return catena::exception_with_status("Authorization failed", catena::StatusCode::PERMISSION_DENIED);
         }
         descriptor_.toProto(*paramInfo.mutable_info(), authz);
         return catena::exception_with_status("", catena::StatusCode::OK);
--- a/sdks/cpp/common/src/Device.cpp
+++ b/sdks/cpp/common/src/Device.cpp
@@ -123,6 +123,36 @@ std::unique_ptr<IParam> Device::getParam(const std::string& fqoid, catena::excep
     }
 }
 
+
+std::vector<std::unique_ptr<IParam>> Device::getTopLevelParams(catena::exception_with_status& status, Authorizer& authz) const {
+    std::vector<std::unique_ptr<IParam>> result;
+    std::cout << "Checking params_ map contents:" << std::endl;
+    try {
+        for (const auto& [name, param] : params_) {
+            std::cout << "Found param in map: '" << name << "'" << std::endl;
+            if (authz.readAuthz(*param)) { 
+                Path path{name, "-"};
+                std::cout << "Checking path: '" << path.toString(true) << "'" << std::endl;
+                auto param_ptr = getParam(path.toString(true), status, authz);  
+                if (param_ptr) {
+                    std::cout << "Successfully got param: '" << path.toString(true) << "'" << std::endl;
+                    result.push_back(std::move(param_ptr));
+                }
+            }
+        }
+    } catch (const catena::exception_with_status& why) {
+        status = catena::exception_with_status(why.what(), why.status);
+    }
+    return result;
+}
+
+
+
+
+
+
+
+
 std::unique_ptr<IParam> Device::getCommand(const std::string& fqoid, catena::exception_with_status& status, Authorizer& authz) const {
    // The Path constructor will throw an exception if the json pointer is invalid, so we use a try catch block to catch it.
     try {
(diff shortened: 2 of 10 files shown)


sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
```diff
@@ -0,0 +1,156 @@
...
+                            responses_.emplace_back();
+                            if (service->authorizationEnabled()) {
+                                top_level_param->toProto(responses_.back(), authz);
+                            } else {
+                                top_level_param->toProto(responses_.back(), catena::common::Authorizer::kAuthzDisabled);
+                            }
+                        }
+                        
+                        status_ = CallStatus::kWrite;
+                        [[fallthrough]];
```
ZuhayrSarkerRV, line 117: I did not add this, this was added by Cursor. I was trying to figure out how to force it to go to kFinish.

ZuhayrSarkerRV:
There might be some funny code here and there while I was trying to debug why it doesn't go to kFinish. For example, I'm pretty sure the Asnynotifywhendone in kWrite doesn't actually do anything useful.

sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
```diff
@@ -0,0 +1,156 @@
...
+                catena::exception_with_status rc{"", catena::StatusCode::OK};
+                catena::common::Authorizer authz = std::vector<std::string>{};
+
+                if (req_.oid_prefix().empty() && !req_.recursive()) {
+                    std::vector<std::unique_ptr<IParam>> top_level_params;
+
+                    Device::LockGuard lg(dm_); 
+                    if (service->authorizationEnabled()) {
+                        std::vector<std::string> scopes = service->getScopes(context_);
+                        authz = catena::common::Authorizer{scopes};
```
mejohnnaylor, line 98: this copy should be avoidable as well by adding `Authorizer::setScopes(const std::vector<std::string> & new_scopes) {scopes_.get() = new_scopes}` or something like that.

sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
```diff
@@ -0,0 +1,156 @@
...
+                        top_level_params = dm_.getTopLevelParams(rc, authz);
+                    } else {
+                        top_level_params = dm_.getTopLevelParams(rc);
+                    }
+
+                    lg.~LockGuard();
```
mejohnnaylor, line 104: its better to do this to release the lock:

```
  // existing code
  {
    Device::LockGuard lg(dm_);
    // critical section here
  } // lock is implicitly freed here

File: sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp [hunk 4]
```diff
@@ -0,0 +1,156 @@
...
+                            }
+                        }
+                        
+                        status_ = CallStatus::kWrite;
+                        [[fallthrough]];
+                    }
+                }
+            } catch (...){
+                status_ = CallStatus::kFinish;
+                writer_.Finish(Status::CANCELLED, this);
```
mejohnnaylor, line 122: status code should be internal error (or equivalent) not CANCELLED.

sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
mejohnnaylor, line 117: do we really fall thru to the next switch? seems odd. maybe I just hadn't noticed this before?

mejohnnaylor (changes requested):

sdks/cpp/common/include/ParamWithValue.h
```diff
@@ -166,6 +166,21 @@ class ParamWithValue : public catena::common::IParam {
...
 
+    /**
+     * @brief serialize the parameter descriptor to protobuf
+     * include both the descriptor and the value
+     * @param paramInfo the protobuf value to serialize to
+     * @param authz the authorization information
+     */
+    catena::exception_with_status toProto(catena::BasicParamInfoResponse& paramInfo, Authorizer& authz) const override {
+        if (!authz.readAuthz(*this)) {
+            return catena::exception_with_status("Param does not exist", catena::StatusCode::INVALID_ARGUMENT);
```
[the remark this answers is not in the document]
ZuhayrSarkerRV: Fixed this for next push.

sdks/cpp/common/src/Device.cpp
```diff
@@ -123,6 +123,36 @@ std::unique_ptr<IParam> Device::getParam(const std::string& fqoid, catena::excep
...
 
+
+std::vector<std::unique_ptr<IParam>> Device::getTopLevelParams(catena::exception_with_status& status, Authorizer& authz) const {
+    std::vector<std::unique_ptr<IParam>> result;
+    std::cout << "Checking params_ map contents:" << std::endl;
+    try {
+        for (const auto& [name, param] : params_) {
+            std::cout << "Found param in map: '" << name << "'" << std::endl;
+            if (authz.readAuthz(*param)) { 
+                Path path{name, "-"};
```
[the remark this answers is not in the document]
ZuhayrSarkerRV: It cannot build the path without this annotation. It fails the regex check - I'm trying to understand why this is at the moment.

sdks/cpp/common/src/Device.cpp
```diff
@@ -123,6 +123,36 @@ std::unique_ptr<IParam> Device::getParam(const std::string& fqoid, catena::excep
...
+                std::cout << "Checking path: '" << path.toString(true) << "'" << std::endl;
+                auto param_ptr = getParam(path.toString(true), status, authz);  
+                if (param_ptr) {
+                    std::cout << "Successfully got param: '" << path.toString(true) << "'" << std::endl;
+                    result.push_back(std::move(param_ptr));
+                }
```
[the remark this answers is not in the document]
ZuhayrSarkerRV: Yes, getParam handles null values and throws an exception. I have added an else statement, but it simply prints:

```c++
                if (param_ptr) {
                    std::cout << "Successfully got param: '" << path.toString(true) << "'" << std::endl;
                    result.push_back(std::move(param_ptr));
                } else {
                    std::cout << "Failed to get param: " << status.what() << std::endl;
                }
                ```

File: sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp [hunk 8]
```diff
@@ -0,0 +1,156 @@
...
+                std::unique_ptr<IParam> param;
+                catena::exception_with_status rc{"", catena::StatusCode::OK};
+                catena::common::Authorizer authz = std::vector<std::string>{};
+
+                if (req_.oid_prefix().empty() && !req_.recursive()) {
+                    std::vector<std::unique_ptr<IParam>> top_level_params;
+
+                    Device::LockGuard lg(dm_); 
+                    if (service->authorizationEnabled()) {
+                        std::vector<std::string> scopes = service->getScopes(context_);
```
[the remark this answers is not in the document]
ZuhayrSarkerRV: Done.

sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
ZuhayrSarkerRV: setScopes is not a function yet, but we can revisit this late when we implement authorization. This if statement is not possible to currently test.

sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
ZuhayrSarkerRV: Understood, I called the destructor because I was trying to debug the code hanging at some point. Did not know it freed implicitly like so. I have modified to be like so:

```c++
                    if (service->authorizationEnabled()) {
                        Device::LockGuard lg(dm_);
                        std::vector<std::string> scopes = std::move(service->getScopes(context_));
                        authz = catena::common::Authorizer{scopes};
                        top_level_params = dm_.getTopLevelParams(rc, authz);
                    } else {
                        Device::LockGuard lg(dm_);
                        top_level_params = dm_.getTopLevelParams(rc);
                    }
                   
```

sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
ZuhayrSarkerRV: Understood and fixed, although this CANCELLED statement appears in other RPCs. Another case of cargo cult coding?

sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
ZuhayrSarkerRV: Yes, the code will fail because there isn't something forcing the call to proceed such as a write. You could forcefully call proceed(), but I'm not sure what design is better/worse. And the falling through is, again, a case of just doing what was done in other RPCs.

sdks/cpp/common/src/Device.cpp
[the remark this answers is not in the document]
ZuhayrSarkerRV: Yeah, no matter what I try, it has to have that hyphen at the end. It would require some changes in Path.h. I was simply following the way it was demonstrated by you in #403 . You can see that when creating the path you still add the end bit.

sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
ZuhayrSarkerRV: Seperate issue

sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
ZuhayrSarkerRV: Will be fixed in issue #415

ZuhayrSarkerRV:
Note that the Path I included here is the exact same as the other branch's Path, so there should be no merge conflicts. This is ready to merge into develop, however I stated in the commit message, I have not been able to figure out how to avoiding falling through to write. 

Feel free to do your review as usual and point out anything that sticks out.

mejohnnaylor (approved):

PR #418, opened 2025-02-07
Corrected Path class to properly overload constructor with ({})

ZuhayrSarkerRV:
Also fixed push_back to work with new system.


mejohnnaylor (changes requested):

sdks/cpp/common/include/Path.h
```diff
@@ -194,28 +188,26 @@ class Path {
...
+            escape(str);
+            segments_.emplace_back(std::in_place_type<std::string>, str);
+        }
     }
 
+    /**
+     * @brief Add a new index segment to the path
+     * @param segment Index to add as a new segment
+     */
+    void push_back(size_t segment) {
```
mejohnnaylor, line 208: sorry to be a pain, but can this be changed to 
```
void push_back(Index idx) {
  segments_.emplace_back(std::in_place_type<Index>, idx);
}
```

I'd forgotten we'd made a typedef for Index, but we should use it.

mejohnnaylor (changes requested):

mejohnnaylor (approved):

PR #426, opened 2025-02-13
Simplified Authorizer to own its scopes vector

ZuhayrSarkerRV:
- Changed constructor to take scopes by value through std::move
- Removed reference wrapper in favor of direct scope member
- Slight changes to hasAuthz to accomodate for these changes


ZuhayrSarkerRV:
Failed check, looking into this.

PR #427, opened 2025-02-13
Simplified Authorizer to own its scopes vector

ZuhayrSarkerRV:
- Changed constructor to take scopes by value through std::move
- Removed reference wrapper in favor of direct scope member
- Slight changes to hasAuthz to accomodate for these changes


mejohnnaylor (changes requested):
wasn't the simplification requested - get the scopes directly from the service and make an object for the scopes that's part of the authorizer's state and is owned by the authorizer.

ZuhayrSarkerRV:
This has been transferred over to Ben

PR #428, opened 2025-02-13
Fixed issue when appending to the back of array params

BenWhitten-RV:
Title.
Below is a list of changes:
- Overhauled SetValue and TrySetValue. Now they are a single function MultiSetValue.
- SetValue still exists, only its functionality has been changed to create a MultiSetValue payload from the information and push it into MultiSetValue.
- MultiSetValue will now return error if the sum of appends exceeds the Max_array_size. 
- GetValue now throws an error when trying to get value at index "/-"
- readAuthz now returns true if the client has read or write access, i.e. if the client has the scope with or without ":w". This is as opposed to previously where it would only return true if they have the scope excluding ":w".
- GetParam now throws an error if path ends in "/-".

Below is a list of additions.
- Added three new functions to ParamWithValue.
    - addBack() adds and returns an empty entry to the back of an array param.
    - popBack() removes the last entry in an array param. Required to revert addBack() in multiSetValue.
    - size() returns the size of the array param.
- Added five new functions to Path related to accessing the back of the path.
    - back_as/is_string() and back_as/is_index() accomplish the same thing as their front counterparts.
    - popBack() removes the last index of the path.

*Additions were required for MultiSetValue to work. Hopefully the overhauled MultiSetValue will also make implementing total_array_size easier.


sdks/cpp/common/include/Device.h
```diff
@@ -377,14 +377,25 @@ class Device {
...
      * @return a unique pointer to the parameter, or nullptr if it does not exist
      * 
      * gets a parameter if it exists and the client is authorized to read it.
      */
     std::unique_ptr<IParam> getParam(const std::string& fqoid, catena::exception_with_status& status, Authorizer& authz = Authorizer::kAuthzDisabled) const;
 
+    /**
+     * @brief get a parameter by oid with authorization
+     * @param path the full path to the parameter
+     ** @param authz The Authorizer to test read permission with.
```
mejohnnaylor, line 391: ** should be *

sdks/cpp/common/include/IParam.h
```diff
@@ -143,6 +141,21 @@ class IParam {
      */
     virtual std::unique_ptr<IParam> getParam(Path& oid, Authorizer& authz, catena::exception_with_status& status) = 0;
 
+    /**
+     * @brief Return the size of an array parameter.
+     */
+    virtual uint32_t size() const = 0;
+
+    /**
+     * @brief Add an empty value to and return the back of an array parameter.
```
mejohnnaylor, line 150: add `@param documentation`

sdks/cpp/common/include/ParamWithValue.h
```diff
@@ -228,6 +228,7 @@ class ParamWithValue : public catena::common::IParam {
             ans = catena::exception_with_status("Value exceeds maximum length", catena::StatusCode::OUT_OF_RANGE);
         } else {
             catena::common::fromProto<T>(value, &value_.get(), descriptor_, authz);
+            return ans;
```
mejohnnaylor, line 231: delete this line

sdks/cpp/common/src/Device.cpp
```diff
@@ -44,42 +44,119 @@
...
-            }
-            else if (!authz.writeAuthz(*param)) {
-                ans = catena::exception_with_status("Not authorized to write to param", catena::StatusCode::PERMISSION_DENIED);
+    // Vector of verified set value requests.
+    std::vector<SetValueRequest> requests;
+    // Vector of paths to verified appends.
+    std::vector<Path*> appends;
+    // Failing multiSetValue if multi set is disabled.
+    if (multi_set_enabled_ == false && src.values().size() > 1) {
+        ans = catena::exception_with_status("Multi-set is disabled", catena::StatusCode::PERMISSION_DENIED);
```
mejohnnaylor, line 56: `invalid argument`

mejohnnaylor (dismissed):

mejohnnaylor (approved):

PR #430, opened 2025-02-19
Added support for structs and struct variants to BasicParamInfo

ZuhayrSarkerRV:
- Supports fetching struct variants both directly and recursively
- Fixed missing logic where it only supported struct arrays, not it supports various types of arrays
- Added recursive function for getting children of parent parameters 
- Added locks for writing, which was a long-standing race condition issue that exists in previous versions of BasicParamInfoRequest 
- Fixed not being able to go kFinish (rpc now exits cleanly)
- Clears responses to avoid duplicates

### DIFF
--- a/sdks/cpp/connections/gRPC/include/BasicParamInfoRequest.h
+++ b/sdks/cpp/connections/gRPC/include/BasicParamInfoRequest.h
@@ -167,6 +167,13 @@ class CatenaServiceImpl::BasicParamInfoRequest : public CallData {
          */
         uint32_t max_index_{0};
 
+        /**
+         * @brief The mutex for the writer lock.
+         */
         std::mutex mtx_;
+
+        /**
+         * @brief The writer lock.
+         */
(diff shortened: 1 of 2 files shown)


sdks/cpp/connections/gRPC/include/BasicParamInfoRequest.h
```diff
@@ -73,7 +73,19 @@ class CatenaServiceImpl::BasicParamInfoRequest : public CallData {
...
 
     private:
-
+        /**
+         * @brief Checks if the parameter is an array type.
+         * 
+         * @param type - The type of the parameter.
+         * @return True if the parameter is an array type, false otherwise.
+         */
+        static bool isArrayType(catena::ParamType type) {
```
mejohnnaylor, line 82: maybe make this a free function that can be used elsewhere?

mejohnnaylor (approved):

PR #433, opened 2025-02-20
Fixed DRY violations when updating responses in Connect

ZuhayrSarkerRV:
- Created a function to process response updates
- Only sends updates to client if authorization passes

NOTE: I'm not sure if we want to use "const_cast". It also looks like Cursor just won't let the quotation mark "fix" go.


mejohnnaylor (dismissed):
fix `const` usage and refactor as discussed.

ZuhayrSarkerRV:
Sorry, ignore the force push, I accidentally merged two branches, it should be fixed now.

johnnaylor (approved):

johnnaylor (approved):

johnnaylor (approved):

mejohnnaylor (approved):

PR #438, opened 2025-02-21
Fixed a few array length calculation bugs + more comments

ZuhayrSarkerRV:
- Top-level parameter array length calculation was missing and not correctly resetting max_index_


sdks/cpp/connections/gRPC/src/BasicParamInfoRequest.cpp
```diff
@@ -106,29 +106,46 @@ void CatenaServiceImpl::BasicParamInfoRequest::proceed(CatenaServiceImpl *servic
...
                         responses_.clear();  
+                        // Process each top-level parameter
                         for (auto& top_level_param : top_level_params) {
+                            max_index_ = 0;
+                            
+                            // Create a path for the parameter and check if it's an array type
+                            Path top_level_path{top_level_param->getOid()};
+                            
+                            // For array types, calculate how many elements exist
+                            if (isArrayType(top_level_param->type().value())) {
```
mejohnnaylor, line 121: this call requires too much knowledge of the param object. please add `bool isArrayType()` as a member of IParam (pure virtual) . The client code then becomes clearer: `if (top_level_param->isArrayType()) {...}`

mejohnnaylor (changes requested):

PR #450, opened 2025-02-26
Couple of logic issues resolved in BasicParamInfo

ZuhayrSarkerRV:
- max_index_ was incredibly problematic, so I removed it and gave each array it's own local length variable
- it now supports taking a prefix like "/audio_deck/0", before it only worked to take "/audio_deck" because it was calculating array length solely based on the provided path and not if it found a parameter or not
- when running the audio_deck example, there were both duplicate and missing parent array parameters - this has been addressed


mejohnnaylor (dismissed):
work on using the IParam support for size() rather than reimplementing calculateArraySize.

mejohnnaylor (approved):

PR #453, opened 2025-02-27
Added subscriptions support for DeviceRequest RPC

ZuhayrSarkerRV:
- Added an overloaded serializer to support subscriptions in Device.h
- Added level of detail support for commands, full, minimum, none and subscriptions
- Modified each parameter for loop in Device.cpp to work with associated detail level requirements 
- A lot of this code is "repeated" logic, but I don't believe it violates DRY
- NOTE: I modified the video player json to allow for some testing with DeviceRequest
- **IMPORTANT**: Codegen has new additions to accommodate for the minimal set


ZuhayrSarkerRV:
I make an expensive copy with the subscriptions vector, and this copy is needed because the vector can be destroy in the lifetime of the coroutine. However, in the plan I have outlined for subscriptions, this may be fixed by having the subscriptions stored elsewhere. So I will leave this code in the background for now.

Issue #452, opened 2025-02-27
Update DeviceRequest to support level of detail and subscriptions


ZuhayrSarkerRV:
This is already done, just opening it as an issue to organize/consolidate things.

ZuhayrSarkerRV:
Creating a new branch because I am having difficulty merging the new changes that were implemented.

PR #467, opened 2025-03-14
Additional support for subscriptions + UpdateSubscriptions RPC

ZuhayrSarkerRV:
This is a fairly large pull request which adds several key functionalities for subscriptions. 

## Compatibility Changes
- Changed naming scheme (from camel case to underscores) in param descriptor to be more consistent 

**Device**:
- Checks for constraints, language packs and menu groups to NOT send if currently in minimal detail level
  - These are only sent in FULL or SUBSCRIPTIONS mode
- Support added for **NONE** (send nothing)
- Support added for **SUBSCRIPTIONS** with a helper function that checks if an OID is subscribed 
  - This function checks for exact oid matches or wildcard matches (ending in /*)
- Parameters are sent if authorized and either in the minimal set or subscribed to
- **COMMANDS** are sent if authorized 

**Examples**:
- Made dashboard UI a minimal set for the use_commands example
  - You can add state to be a subscription to play around with as an example

**Connect**:
- Changed the subscription manager to be a reference instead of an instance
- Initialize this reference with the manager from the service

**ServiceImpl**:
- Added subscription manager to be initialized HERE instead 
- UpdateSubscriptions is created as a CallData object here

**SubscriptionManager**:
- Modified add function to take device parameter (function is used in UpdateSubscriptions)
- Added a function to process children, similar to how it is done recursively in BasicParamInfoRequest
- BIG: Expanded the add subscription function to work with wildcards
  - This essentially just checks that if the provided OID is given with a /*, it adds all the children
  - If it can't find the parameter, it tries to find parameters under this path

## UpdateSubscriptions RPC
- Follows the same flow of most other RPCs, although is more similar to BasicParamInfoRequest as it sends responses in a stream
- Has a helper function to send a stream of responses

## **Alternate "fixes"**
**ServiceCredentials**:
- Modified some of the functions to use the global namespace because for some reason my code just refused to compile without these changes (and VS code also complained)

**Codegen**:
- Added minimal set, although I'm not even sure why this had to be added -I thought it was already pulled into the develop branch

## TODO

- Add the remaining functionality in Device Request (refer to #452 )
- Add the ability to test all 5 levels of detail (commands is "missing" a test)


sdks/cpp/common/src/Device.cpp
```diff
@@ -313,7 +313,11 @@ void Device::toProto(::catena::Device& dst, Authorizer& authz, bool shallow) con
     // if we're not doing a shallow copy, we need to copy all the Items
     google::protobuf::Map<std::string, ::catena::Param> dstParams{};
     for (const auto& [name, param] : params_) {
-        if (authz.readAuthz(*param)) {
```
mejohnnaylor, line 316: opportunity for DRY improvement here. look at using a lambda or free function as discussed in review

sdks/cpp/connections/gRPC/include/Connect.h
```diff
@@ -146,9 +146,9 @@ class CatenaServiceImpl::Connect : public CallData {
         unsigned int shutdownSignalId_;
 
         /**
-         * @brief Subscription manager instance for this connection
+         * @brief Reference to the subscription manager
          */
-        catena::grpc::SubscriptionManager subscriptionManager_;
+        catena::grpc::SubscriptionManager& subscriptionManager_;
```
mejohnnaylor, line 151: violates encapsulation principle of OOP - only the owner class should be able to manipulate its state, unless there are really good reasons to expose its state to other client code. ideally provide public methods to access this object.

mejohnnaylor: https://chatgpt.com/share/e/67d8777a-2260-8000-98fa-d263938d953b

mejohnnaylor (changes requested):

sdks/cpp/connections/gRPC/src/SubscriptionManager.cpp
```diff
@@ -47,21 +48,66 @@ bool SubscriptionManager::isWildcard(const std::string& oid) {
...
 
 // Add a subscription (either unique or wildcard). Returns true if added, false if already exists
-bool SubscriptionManager::addSubscription(const std::string& oid) {
+bool SubscriptionManager::addSubscription(const std::string& oid, catena::common::Device& dm) {
+    catena::exception_with_status rc{"", catena::StatusCode::OK};
+
+    // Check if this is a wildcard subscription
     if (isWildcard(oid)) {
+        // Extract base OID (everything before the wildcard)
+        std::string baseOid = oid.substr(0, oid.length() - 2); // Remove "/*"
```
mejohnnaylor, line 57: need to trap invalid oids and bail out - e.g. zero length after removing `/*`

mejohnnaylor: discussion - isWildcard does some length validation - but this needs to be documented in the doxygen / header.

mejohnnaylor (dismissed):

sdks/cpp/connections/gRPC/src/SubscriptionManager.cpp
```diff
@@ -73,22 +119,86 @@ void SubscriptionManager::updateAllSubscribedOids_(catena::common::Device& dm) {
...
+                        allSubscribedOids_.push_back(paramOid);
+                        processChildren_(paramOid, param.get(), dm);
+                    }
+                }
+            }
         }
+    }
+}
+
+void SubscriptionManager::processChildren_(const std::string& parent_path, IParam* current_param, catena::common::Device& dm) {
```
ZuhayrSarkerRV, line 176: Suggested to look into visitor pattern to avoid DRY with this class and BPIR. Also, add additional comments if possible.

sdks/cpp/common/src/Device.cpp
ZuhayrSarkerRV: I tried removing as much DRY as possible in the file, but I can't put c++ coroutines in lambda functions and its redundant in regular functions as each object being serialized is of a different type.

sdks/cpp/connections/gRPC/include/Connect.h
ZuhayrSarkerRV: Cursor suggested dependency injection through the constructor, this is probably better.

sdks/cpp/connections/gRPC/src/SubscriptionManager.cpp
ZuhayrSarkerRV: Referenced in issue #473 , will do this afterwards.

PR #469, opened 2025-03-17
Overhaul of MultiSetValue to support totalLength and fix issue with setting values

BenWhitten-RV:
This is a large enhancement containing an overhaul of MultiSetValue which now supports totalLength for string arrays as well as fixing issues present in the last iteration. Each param now contains an mSizeTracker (size_t) and tSizeTracker (vector<size_t>) used to tracker maxSize and totalSize respectively (latter is only applicable to String array params, nullptr otherwise). MultiSetValue is now the combination of two function try and commit multiSetValue similar to the original iteration. The difference here is that the try uses the trackers described above to keep track of updates and does not change anything about the actual value of the param.  The biggest benefit of this change will come once slots are implemented, as it will allow the command to be completely validated across all devices before committing, as opposed to some changes getting commited as would be with the previous implementation.

Changelog:
- Overhauled MultiSetValue to support totalLength for string arrays
- Added command line option to set defaultTotalLength. Similar to defaultMaxLength this must be implemented in the business implementation.
- Fixed issue with setting values in MultiSetValue
- Fixed issue with int arrays not clearing on set


sdks/cpp/common/include/IParam.h
```diff
@@ -196,7 +191,18 @@ class IParam {
...
     virtual bool isArrayType() const = 0;
-
+    /**
+     * @brief Validates a setValue operation without changing the param's value.
+     * @param value The value we want to set the param to.
+     * @param index The index of the subparam to set (or nullptr if none).
+     * @param ans Catena::exception_with_status output.
+     * @returns true if valid.
+     */
+    virtual bool validateSetValue(const catena::Value& value, uint32_t* index, Authorizer& authz, catena::exception_with_status& ans) = 0;
```
mejohnnaylor, line 201: use a signal from Path (e.gl `kEnd`, or whatever's appropriate) feel free to extend the number of signals as needed.

mejohnnaylor (changes requested):

mejohnnaylor (approved):

NR-RV (approved):

PR #471, opened 2025-03-17
Added ability to specify authorization server and realm via command line args

BenWhitten-RV:
Title. You can now use AUTHZ_SERVER=your.authorization.server and REALM=realm before grpc-compose to specify the authorization server and realm. Updating documentation as well to reflect this addition.

### DIFF
--- a/sdks/cpp/connections/docker/catena_envoy_dockerfile
+++ b/sdks/cpp/connections/docker/catena_envoy_dockerfile
@@ -1,4 +1,4 @@
-#
+# 
 # Copyright 2024 Ross Video Ltd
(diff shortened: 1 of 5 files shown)


sdks/cpp/connections/docker/catena_envoy_dockerfile
```diff
@@ -0,0 +1,48 @@
+# 
+# Copyright 2024 Ross Video Ltd
```
mejohnnaylor, line 2: 2025

mejohnnaylor (changes requested):

PR #474, opened 2025-03-18
Made the container hosted example more scalable

BenWhitten-RV:
Title. Envoy now only listen's to one port and routes based on the url (status_update.$HOST, use_menus.$HOST, etc. where $HOST is the host name). Additionally, $HOST can be specified by the command line argument $HOST=... similar to authz server and realm.

_**Note:** If running on a VM you will most likely need to go into your host machine's hosts file and add mappings for your addresses (status_update.localhost, etc...)_

_**Note:** this pull request is not actually as big as it seems. This is a sub branch of the sub branch enh-#470-b-bw as it depends on changes and would conflict otherwise. enh-#460-b-bw just needs to sign off to get merged into DEV, and as the last review was just pointing out that the copyright was out of date that will hopefully be soon._

### DIFF
--- a/sdks/cpp/connections/docker/envoy/envoy.yaml
+++ b/sdks/cpp/connections/docker/envoy/envoy.yaml
@@ -47,19 +47,6 @@ static_resources:
       socket_address:
         address: 0.0.0.0
         port_value: 6254
-    additional_addresses:
-    - address:
-        socket_address:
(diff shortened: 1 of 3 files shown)


NR-RV (commented):
closing tell Kubernetes setup is done

PR #476, opened 2025-03-19
DeviceRequest subscriptions support

ZuhayrSarkerRV:
This was already demo'd earlier today, but I have fixed it to work with the UpdateSubscriptions RPC correctly now. In addition, it correctly returns the minimal set in subscriptions mode.


sdks/cpp/common/src/Device.cpp
```diff
@@ -524,11 +527,28 @@ Device::DeviceSerializer Device::getComponentSerializer(Authorizer& authz, const
 }
 
 bool Device::shouldSendParam(const IParam& param, bool is_subscribed, Authorizer& authz) const {
-    bool auth_check = authz.readAuthz(param);
-    bool minimal_check = param.getDescriptor().minimalSet();
-    bool full_check = detail_level_ == DetailLevel(Device_DetailLevel_FULL);
-    bool subscription_check = detail_level_ == DetailLevel(Device_DetailLevel_SUBSCRIPTIONS) && is_subscribed;
-    
-    return auth_check && (minimal_check || full_check || subscription_check);
+    // First check authorization
```
NR-RV, line 530: try to avoid switch staments (they are a pain to test for full code coverage) :D

also try to avoid >1 return in a function (ill be going back though the rest of the code and noting other places this happens but for new code try to refraim from that)

```suggestion
    // treat this as sudo code
    
    bool results = false;
    // Use positive case ifs
    if (authz.readAuthz(param)) {
        if (detail_level_ == Device_DetailLevel_MINIMAL) {
            results = param.getDescriptor().minimalSet();
        } else if (detail_level_ == Device_DetailLevel_FULL) {
            results = true;
        } else if (detail_level_ == Device_DetailLevel_SUBSCRIPTIONS) {
            results = param.getDescriptor().minimalSet() ? true : is_subscribed;
        } else if (detail_level_ == Device_DetailLevel_COMMANDS) {
            results = param.getDescriptor().isCommand();
        }
    }
    return results;
```

NR-RV (approved):
updateding the switch to if and the negitive if logic to positive 
looks

NR-RV (approved):

NR-RV (approved):
logic looks good and now casting the enums to int

PR #479, opened 2025-03-21
Fixed getValue not working for indices. Fixed numerous error messages. Updated toParam flow.

BenWhitten-RV:
Changelog:
- Fixed getValue not working for indices. Previously, an if statement oversight prevented and params with an oid ending in an index being retrieves. This has been fixed.
- Removed trailing whitespace from error messages throughout Device and ParamWithValue. Postman does not display an error message if this is present.
- Updated toParam's if else flow to resemble something like the following:
```
catena::error ans = OK
if ...
    ans = ERROR
else
    do something
return ans
```
  This is the format present in other places throughout Catena and has been previously mentioned to be the ideal formatting.


NR-RV (approved):
I like the more info in the error msgs

Issue #480, opened 2025-03-21
Refactor subscription management to exist within catena::common


ZuhayrSarkerRV:
This should make development easier with REST

PR #481, opened 2025-03-21
Visitor pattern implementation for subscriptions

ZuhayrSarkerRV:
This is also within catena:common, so it will be beneficial for when I refactor subscriptions in issue #480 

Essentially I took the parameter traversal logic from BasicParamInfoRequest and SubscriptionManager and moved it to a visitor class, which not only removes duplicate code but centralizes the logic for use in other parts of the codebase.


ZuhayrSarkerRV:
Please wait for me to clean up the code and finish the build issues before accepting this PR

sdks/cpp/common/include/ParamVisitor.h
NR-RV, line 1: make the interface file for this and inherit it

NR-RV (approved):

PR #485, opened 2025-03-26
[BREAKING] Fixed SetValue messages in param.proto and updated gRPCs to reflect this.

BenWhitten-RV:
Title.

This involved creating a new function for (Multi)SetValue toMulti() which converts a SingleSetValuePayload to a multi one.

One thing to note is that I reused field numbers. However, as far as I know we are still in dev and everything will be rebuilt anyways so it should be fine.

### DIFF
--- a/sdks/cpp/connections/gRPC/include/MultiSetValue.h
+++ b/sdks/cpp/connections/gRPC/include/MultiSetValue.h
@@ -68,32 +68,39 @@ class CatenaServiceImpl::MultiSetValue : public CallData {
          */
         void proceed(CatenaServiceImpl *service, bool ok) override;
     protected:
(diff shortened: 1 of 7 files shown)


sdks/cpp/connections/gRPC/include/MultiSetValue.h
```diff
@@ -68,6 +68,7 @@ class CatenaServiceImpl::MultiSetValue : public CallData {
          */
         void proceed(CatenaServiceImpl *service, bool ok) override;
     protected:
+        // Helper functions to allow child setValue to use proceed()
```
NR-RV, line 71: this is in the wrong place

NR-RV (approved):

NR-RV (approved):

PR #490, opened 2025-04-02
Start of REST API using Boost::asio

BenWhitten-RV:
**Changelog:**
- Implemented working REST API using Boost::asio
- Implemented Connect, DeviceRequest, GetValue, SetValue, MultiSetValue, and GetPopulatedSlots RPCs
    - By extension created outline for all RPCs implemented in the future.
- Created helper functions for reading from/writing to sockets to streamline RPC translation.
- Implemented StatusUpdate example from connections/gRPC into REST
- Moved Connect update() behavior to shared common/rpc file/folder.

**Todo:**
- Update field parsing to match decided upon format.
- Document using openAPI
- Organize file structure
- Remove crow from project? Currently unused.
- Update Connect gRPC to use shared update() behavior.

I am probably missing a few things on either list above. My suggestion however is to push this in an incomplete but working state for now so we can split everything else up into smaller, more manageable issues. This is all separate from our implemented gRPC connection so none of this should interfere with what we already have.


BenWhitten-RV:
```
-- jwt-cpp: using FetchContent for nlohmann-json required for tests
CMake Error at build/_deps/nlohmann_json-src/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


-- Configuring incomplete, errors occurred!
```

Seems to be an issue with CMAKE and jwt-cpp version... will look into this.

NR-RV (changes requested):
dont forget to also add the comments before functions that give say what it does
(unless the name is super obvious like: get* set*)

NR-RV (approved):

PR #493, opened 2025-04-03
Dev containers and tool chain dockized

NR-RV:
this has the dev containers working with cursor
also has the dockerfile for the tool chain and the .env file for managing tool versions 
the compose file is set up for pushing to the rossvideo container registry
(you need to setup your api key if you want to push)
image


.devcontainer/README.md
```diff
@@ -0,0 +1,59 @@
+# Dev Containers
+
+## file brake down
```
mejohnnaylor, line 3: sp: break down

.devcontainer/devcontainer.dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM ghcr.io/rossvideo/catena-toolchain:latest
+LABEL org.opencontainers.image.title="Catena SDK"
+LABEL org.opencontainers.image.description="Rossvideo Catena"
+LABEL org.opencontainers.image.version="v0.0.1-dev"
```
mejohnnaylor, line 4: is there a way to get the version from VERSION.TXT?

.devcontainer/devcontainer.dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM ghcr.io/rossvideo/catena-toolchain:latest
+LABEL org.opencontainers.image.title="Catena SDK"
```
mejohnnaylor, line 2: the language should be included, we'll have different images for the C++ and Java SDKs

.devcontainer/devcontainer.dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    && git pull origin develop \
+    && git submodule update --init --recursive
+
+# Build Catena
+RUN mkdir -p ~/Catena/sdks/cpp/build \
+    && cd ~/Catena/sdks/cpp/build \
+    && cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCONNECTIONS='gRPC;REST' -DCMAKE_INSTALL_PREFIX=~/.local ~/Catena/sdks/cpp \
+    && ninja
+
+# Set the default user
```
mejohnnaylor, line 22: `# Set the working directory` is more accurate. We shouldn't be operating as root, either. The user's `gid` and `uid` can be passed-in to the container at start-up

mejohnnaylor (dismissed):

PR #518, opened 2025-04-10
ListLanguages REST

BenWhitten-RV:
Changelog:
- Implemented ListLanguages in REST
- Updates old comments in other RPCs


NR-RV (approved):
the rpc will be changed latter

PR #523, opened 2025-04-11
Fixed REST example device models

BenWhitten-RV:
Fixed error in the status_update_REST and use_menus_REST device models preventing compilation.


NR-RV (approved):
good catch

PR #531, opened 2025-04-14
Fully implement DeviceRequest within REST

ZuhayrSarkerRV:
DeviceRequest should now support level of detail correctly and subscriptions (using the central manager that was added to catena::common). 

I imagine I might revisit this class when I add additional support to connect, because I need to test that it works between the two classes.


ZuhayrSarkerRV:
I am currently dealing with issues syncing the manager with the REST service. I am going to close this and either edit or open a new PR detailing the changes.

PR #533, opened 2025-04-14
Replaced ChunkedWriter with SSEWriter

BenWhitten-RV:
Changelog:
- ChunkedWriter has been replaced with SSEWriter, which writes Server Sent Events to the client rather than Chunked encoded responses.
- SSEWriter is no longer a subclass of SocketWriter and is now its own class inheriting ISocketWriter
- WriteOptions moved from SocketWriter to ServiceImpl.
- userAgent removed from SocketReader as the switch to Server Sent Events rendered it useless (Postman supports SSEs)
- Updated openAPI docs for Connect and DeviceRequest. OpenAPI buffers SSEs, but it also bufefd chunked encoding so there is nothing new there.
- moved codeMap_ from SocketWriter to catena::REST. 
- Made many of SocketWriter's variables private since it is no longer inherited.

**Note:** John wants a demo of this working, so maybe hold off on merging until then. This works SO much better than Chunked Encoding ever did. It is also like half the size of ChunkedWriter as the response formatting is a lot simpler.

### DIFF
--- a/sdks/cpp/connections/REST/include/ServiceImpl.h
+++ b/sdks/cpp/connections/REST/include/ServiceImpl.h
@@ -64,7 +64,7 @@ using boost::asio::ip::tcp;
 
 using catena::REST::SocketReader;
 using catena::REST::SocketWriter;
-using catena::REST::ChunkedWriter;
+using catena::REST::SSEWriter;
 
 namespace catena {
 /**
@@ -111,6 +111,10 @@ class CatenaServiceImpl : public catena::REST::IServiceImpl {
     bool authorizationEnabled() override { return authorizationEnabled_; };
     
   private:
+    /**
+     * @brief Writes a client's options to the socket.
+     */
+    void writeOptions(tcp::socket& socket, const std::string& origin);
     /**
      * @brief Returns true if port_ is already in use.
      * 
--- a/sdks/cpp/connections/REST/include/SocketReader.h
+++ b/sdks/cpp/connections/REST/include/SocketReader.h
@@ -99,10 +99,6 @@ class SocketReader : public ISocketReader {
      * @brief Returns the origin of the request.
      */
     const std::string& origin() const override { return origin_; }
-    /**
-     * @brief Returns the agent used to send the request.
-     */
-    const std::string& userAgent() const override { return userAgent_; }
 
     /**
      * @brief Returns true if authorization is enabled.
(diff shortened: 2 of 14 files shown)


sdks/cpp/connections/REST/include/ServiceImpl.h
```diff
@@ -111,6 +111,10 @@ class CatenaServiceImpl : public catena::REST::IServiceImpl {
     bool authorizationEnabled() override { return authorizationEnabled_; };
     
   private:
+    /**
+     * @brief Writes a client's options to the socket.
+     */
+    void writeOptions(tcp::socket& socket, const std::string& origin);
```
BenWhitten-RV, line 117: Moved from SocketWriter

sdks/cpp/connections/REST/include/SocketReader.h
```diff
@@ -99,10 +99,6 @@ class SocketReader : public ISocketReader {
      * @brief Returns the origin of the request.
      */
     const std::string& origin() const override { return origin_; }
-    /**
-     * @brief Returns the agent used to send the request.
-     */
-    const std::string& userAgent() const override { return userAgent_; }
```
BenWhitten-RV, line 105: No longer necessary as PostMan supports SSEs

sdks/cpp/connections/REST/include/SocketWriter.h
```diff
@@ -97,21 +97,12 @@ class SocketWriter : public ISocketWriter {
...
      */
     void finish(google::protobuf::Message& msg);
-    /**
-     * @brief Writes a response to the client detaining their options.
-     * Used when method = OPTIONS.
-     */
-    void writeOptions() override;
 
-  protected:
+  private:
```
BenWhitten-RV, line 101: SocketReader is no longer inherited so everything has been made private

sdks/cpp/connections/REST/include/SocketWriter.h
```diff
@@ -120,77 +111,75 @@ class SocketWriter : public ISocketWriter {
...
+     * @brief The socket to write to.
      */
-    std::string userAgent_ = "";
+    tcp::socket& socket_;
+};
+
+/**
+ * @brief Maps catena::StatusCode to HTTP status codes.
+ */
+const std::map<catena::StatusCode, int> codeMap_ {
```
BenWhitten-RV, line 164: Moved outside of SocketWriter as SSEWriter no longer inherits it

NR-RV (approved):

PR #535, opened 2025-04-15
DeviceRequest subscription support functioning in REST, pending some possible changes.

ZuhayrSarkerRV:
This pull request needs to be pulled after the switch to SSE over chunked encoding is approved, and after we have decided on how to format the requests (csv, json, etc.). Here are the list of changes/fixes/enhancements:

### ServiceImpl

- SubscriptionManager was switched to be initialized within the service, simlar to the gRPC implementation
- The ServiceImplementation now initializes a SocketReader variable with itself (the service)

### SocketReader
- SocketReader has a constructor that has the service passed into it (similarly, a service member variable is added)
  - This needed to be done because the manager needs to persist for the lifespan of the service, not for the lifespan of a SocketReader
  - This also means SocketReader has a reference to the manager  
 - Modified the "fields" fieldMap to split the request into parts without throwing - when it throws, the client side hangs up on an unknown error
   - We can manage these issues in their respective request handlers (ex. if a parameter is not found, DeviceRequest will catch it and thrown an error) 

### DeviceRequest
- The service is referenced in DeviceRequest, mainly to access the subscription manager
- Fix detail level to be catena::Device_DetailLevel instead of int
- Needed an optional serializer for DeviceRequest (again, similar to gRPC)
- The constructor now looks for empty fields and skips them, instead of pushing them onto other fields
  - Had a funny bug where if I set the language to // (empty) instead of en (or any language) or {}, it would set the detail level to "slot"


ZuhayrSarkerRV:
We need to look into making DeviceRequest a POST, not a GET.

ZuhayrSarkerRV:
Closing this because there are some issues that I would like to fix in this and other handlers.

PR #543, opened 2025-04-21
Fix some linking issues on messages.yaml

ZuhayrSarkerRV:
Waiting for Ben to review this before it gets pushed in.

### DIFF
--- a/interface/openapi/paths/Messages.yaml
+++ b/interface/openapi/paths/Messages.yaml
@@ -15,7 +15,7 @@
 # may be used to endorse or promote products derived from this software without
 # specific prior written permission.
 #
-# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS “AS IS”
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 # AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 # IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
 # ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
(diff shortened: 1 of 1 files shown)


interface/openapi/paths/Messages.yaml
```diff
@@ -67,12 +67,33 @@ $defs:
...
+            type: boolean
+          access_scopes:
+            type: array
+            items:
+              type: string
+          default_scope:
+            type: string
 
-      - $ref: "#/$defs/component_param"
+      - title: Component Param
```
BenWhitten-RV, line 89: Repeat of component param which is defined above. This is one of a few device components used elsewhere other than deviceReq

interface/openapi/paths/Messages.yaml
```diff
@@ -98,13 +134,19 @@ $defs:
...
             $ref: "../../../schema/catena.device_schema.json#/$defs/slot"
-          # Problems linking param
           param:
-            title: Parameter
-            description: Defines the type, value, constraints, UI, access modes and configuration options of a parameter.
-            type: object
+            $ref: "../../../schema/catena.param_schema.json#"
 
-      - $ref: "#/$defs/component_language_pack"
+      - title: Component Language Pack
```
BenWhitten-RV, line 140: Repeat of component language pack which is defined above. This is one of a few messages defined outside of device component as it's used in multiple places other than deviceReq

interface/openapi/paths/Messages.yaml
```diff
@@ -114,11 +156,8 @@ $defs:
       type:
         title: String
         type: string
-      # Problems linking PolyGlotText
       error_message:
-        title: PolyGlot Text
-        description: Text that a client can display in one of multiple languages
-        type: object
+        $ref: "../../../schema/catena.param_schema.json#/$defs/polyglot_text"
```
BenWhitten-RV, line 160: PolyGlot text is a circular dependency (I believe), which openAPI does not like, hence why I redefined it previously.

interface/openapi/paths/Messages.yaml
```diff
@@ -81,10 +102,25 @@ $defs:
           slot:
             $ref: "../../../schema/catena.device_schema.json#/$defs/slot"
           menu:
-            $ref: "../../../schema/catena.device_schema.json#/$defs/menu"
+            type: object
+            properties:
+              name:
+                $ref: "../../../schema/catena.param_schema.json#/$defs/polyglot_text"
```
BenWhitten-RV, line 108: PolyGlotText here is a circular dependency, which is also an issue with the original version. This causes openAPI to break when attempting to expand menu.

BenWhitten-RV (changes requested):
This new messages.yaml does not work on my end, possibly due to differences between vsCode and Cursor. Furthermore, it nets me with more errors than before, and across close to all API calls.

Also this new iteration has a lot of repeated code.

I had previously chopped up many of these to be issues on openAPI's side, since they only showed up when expanding DeviceRequest before Connect. I stand by this claim. If we really wanted to fix linking issues, it would probably be best to add swagger to the dev container and test using that rather than the extension that allows you to preview the docs.

This is a much deeper issue. The APIs should still work regardless since none of the inputs have issues and the API does not check the schema when displaying response. I suggest leaving this as an open issue for now and we can tackle this after finishing some other higher priority tasks.

ZuhayrSarkerRV:
As Ben mentioned, there are larger issues, so I'm going to leave this PR closed, perhaps more as a reference point.

PR #549, opened 2025-04-23
Implemented GetParam

BenWhitten-RV:
Title, will need to be updated once IDevice is merged or vice versa

### DIFF
--- a/sdks/cpp/connections/REST/include/controllers/GetParam.h
+++ b/sdks/cpp/connections/REST/include/controllers/GetParam.h
@@ -45,7 +45,7 @@
 #include <rpc/TimeNow.h>
 #include <Status.h>
 #include <IParam.h>
-#include <Device.h>
+#include <IDevice.h>
 #include <utils.h>
 #include <Authorization.h>
 #include <Enums.h>
@@ -64,7 +64,7 @@ namespace REST {
 class GetParam : public ICallData {
   public:
     // Specifying which Device and IParam to use (defaults to catena::...)
-    using Device = catena::common::Device;
+    using IDevice = catena::common::IDevice;
     using IParam = catena::common::IParam;
 
     /**
@@ -74,7 +74,7 @@ class GetParam : public ICallData {
      * @param context The SocketReader object.
      * @param dm The device to get the value from.
      */ 
-    GetParam(tcp::socket& socket, SocketReader& context, Device& dm);
+    GetParam(tcp::socket& socket, SocketReader& context, IDevice& dm);
     /**
      * @brief GetValue's main process.
      */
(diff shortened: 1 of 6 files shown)


BenWhitten-RV:
Updated to work with IDevice. Should be good to merge

sdks/cpp/connections/REST/include/controllers/GetParam.h
```diff
@@ -0,0 +1,138 @@
...
+      return new GetParam(socket, context, dm);
+    }
+  private:
+    /**
+     * @brief Helper function to write status messages to the API console.
+     * 
+     * @param status The current state of the RPC (kCreate, kFinish, etc.)
+     * @param ok The status of the RPC (open or closed).
+     */
+    inline void writeConsole(CallStatus status, bool ok) const override {
```
mejohnnaylor, line 103: as a private method, its name should have a trailing underscore

sdks/cpp/connections/REST/src/controllers/GetParam.cpp
```diff
@@ -0,0 +1,54 @@
...
+        }
+        // Locking device and getting the param.
+        std::lock_guard lg(dm_.mutex());
+        std::unique_ptr<IParam> param = dm_.getParam("/" + context_.fields("oid"), rc, *authz);
+        if (rc.status == catena::StatusCode::OK && param) {
+            ans.set_oid(param->getOid());
+            param->toProto(*ans.mutable_param(), *authz);
+        }
+    // ERROR
+    } catch (...) {
```
mejohnnaylor, line 39: does this need another `catch` block for `catena::exception_with_status&` so that known errors can be reported accurately? IDK if anything in the `try` block can throw, so I'm just asking the question.

sdks/cpp/connections/REST/include/controllers/GetParam.h
```diff
@@ -0,0 +1,138 @@
...
+    using IParam = catena::common::IParam;
+
+    /**
+     * @brief Constructor for the GetValue RPC.
+     *
+     * @param socket The socket to write the response to.
+     * @param context The SocketReader object.
+     * @param dm The device to get the value from.
+     */ 
+    GetParam(tcp::socket& socket, SocketReader& context, IDevice& dm);
```
mejohnnaylor, line 77: if `makeOne` is the only way these objects are created, maybe the ctor should be private?

mejohnnaylor (changes requested):

sdks/cpp/connections/REST/src/controllers/GetParam.cpp
BenWhitten-RV: Added extra catch to getParam and getValue

Issue #532, opened 2025-04-14
Add UpdateSubscriptions to REST


ZuhayrSarkerRV:
As the title states - this should be a lot easier now that everything has been moved to common.

Aha! Link: https://rossvideo.aha.io/features/CTN-18

PR #550, opened 2025-04-23
Preliminary REST LOD support, primarily on DeviceRequest

ZuhayrSarkerRV:
This is the first part of adding LOD (Level of Detail) support to REST, the second part is going to be with Connect. 

Again, it is mostly similar to the gRPC implementation aside from one MAJOR change - DeviceRequest NO LONGER supports adding/removing subscriptions in REST. This overcomplicates things as it requires an additional POST/PUT handler. This one is simply a GET method that does exactly what it says - request information on device.

It supports all the levels of detail, including subscriptions, but we can fully test subscriptions on it when I finally implement the update subscriptions handler in #532 .

Here are a list of changes:

### Enums.h / device.proto
- Added new `UNSET` detail level in order to allow openAPI to recognize when the client has set a detail level (used to be `NONE`, but the client can set that too)

### ServiceImpl
- SubscriptionManager was switched to be initialized within the service, similar to the gRPC implementation
- The ServiceImplementation now initializes a SocketReader variable with itself (the service)

### SocketReader
- SocketReader has a constructor that has the service passed into it (similarly, a service member variable is added)
  - This needed to be done because the manager needs to persist for the lifespan of the service, not for the lifespan of a SocketReader
  - This also means SocketReader has a reference to the manager  
- Default device LOD is now `UNSET` - this is the uninitialized state 

### Device
- Minor changes to add `UNSET` to the list of detail levels - again, its just a flag to check if the client explicitly set a detail level

### DeviceRequest
- The service is referenced in DeviceRequest, mainly to access the subscription manager
- Fix detail level to be `catena::Device_DetailLevel` instead of `int`
- Needed an optional serializer for DeviceRequest (again, similar to gRPC)
- Utilizes updated system to grab fields from the context
- General functionality for setting the serializer based on detail level
- No longer has update/removal functionality for subscriptions 

### General
- Renamed `service_` variable (the url) to be `endpoint_`, this is more fitting for REST -then the actual `service_` variable can be of type `CatenaServiceImpl`

### DIFF
--- a/interface/device.proto
+++ b/interface/device.proto
@@ -42,6 +42,7 @@ message Device {
     MINIMAL = 2;
     COMMANDS = 3;
     NONE = 4;
+    UNSET = 5;
   }
   DetailLevel detail_level = 2;
 
--- a/sdks/cpp/common/include/Enums.h
+++ b/sdks/cpp/common/include/Enums.h
@@ -65,7 +65,8 @@ inline const common::DetailLevel::FwdMap common::DetailLevel::fwdMap_ = {
   {Device_DetailLevel_SUBSCRIPTIONS, "SUBSCRIPTIONS"},
   {Device_DetailLevel_MINIMAL,       "MINIMAL"},
   {Device_DetailLevel_COMMANDS,      "COMMANDS"},
-  {Device_DetailLevel_NONE,          "NONE"}
+  {Device_DetailLevel_NONE,          "NONE"},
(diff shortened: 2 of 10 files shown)


ZuhayrSarkerRV:
Hold on, don't accept this, there's something that's missing with UNSET.

mejohnnaylor (approved):

PR #551, opened 2025-04-23
Fix the failed automatic merge for BPIR

ZuhayrSarkerRV:
The automatic merge didn't work (that's a first). So here's a hotfix.


NR-RV (approved):
good catch

PR #569, opened 2025-04-28
Fixed gRPC Connect command update bug + refactoring .find to be .contain

ZuhayrSarkerRV:
The bug was essentially that we had duplicate update variables so they have since been removed from gRPC (and the common variables are used instead)

Also changed the map in the common connect to have a return variable instead of multiple returns

NEW: Updated a bunch of files that normally used .find to instead use .contains

### DIFF
--- a/sdks/cpp/common/include/rpc/Connect.h
+++ b/sdks/cpp/common/include/rpc/Connect.h
@@ -76,7 +76,7 @@ class Connect : public IConnect {
      * @param jwsToken The client's JWS token.
      * @param subscriptionManager The subscription manager.
      */
-    Connect(IDevice& dm, bool authz, const sFixed gRPC connect command update bug d::string& jwsToken, ISubscriptionManager& subscriptionManager) : 
+    Connect(IDevice& dm, bool authz, const std::string& jwsToken, ISubscriptionManager& subscriptionManager) : 
         dm_{dm}, 
         subscriptionManager_{subscriptionManager},
         detailLevel_{catena::Device_DetailLevel_UNSET} {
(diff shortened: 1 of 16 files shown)


sdks/cpp/common/include/rpc/Connect.h
```diff
@@ -132,39 +132,45 @@ class Connect : public IConnect {
...
                 }},
                 {catena::Device_DetailLevel_NONE, [&]() {
                     // Don't send any updates
-                    return false;
+                    std::cout << "Updating for NONE detail level" << std::endl;
+                    should_update = false;
                 }}
             };
 
             auto it = detailLevelMap.find(this->detailLevel_);
```
mejohnnaylor, line 164: why aren't you using `detailLevelMap.contains(this->detailLevel_)`?

mejohnnaylor (changes requested):

sdks/cpp/common/include/rpc/Connect.h
ZuhayrSarkerRV: Because I didn't know we could do that. I'll update it to do so.

sdks/cpp/common/include/rpc/Connect.h
```diff
@@ -132,39 +132,44 @@ class Connect : public IConnect {
             bool should_update = false;
             
             // Map of detail levels to their update logic
-            const std::unordered_map<catena::Device_DetailLevel, std::function<bool()>> detailLevelMap {
+            const std::unordered_map<catena::Device_DetailLevel, std::function<void()>> detailLevelMap {
                 {catena::Device_DetailLevel_FULL, [&]() {
                     // Always update for FULL detail level
-                    return true;
+                    std::cout << "Updating for FULL detail level" << std::endl;
```
mejohnnaylor, line 138: drop the couts

mejohnnaylor (changes requested):

PR #602, opened 2025-05-08
Added a listener for commands in Connect (gRPC/REST)

ZuhayrSarkerRV:
Any time a command is executed, if in the COMMANDS detail level within connect, it should now update. This was implemented in both gRPC and REST.


ZuhayrSarkerRV:
Don't merge this in yet, I'm curious about something regarding it.

ZuhayrSarkerRV:
I'm closing this because I'm very confused with how to implement it properly and I've allocated too much time to figure it out. I may return back to it after.

Issue #580, opened 2025-05-02
Update response codes across all the REST controllers


ZuhayrSarkerRV:
Self-explanatory, just need to make it so that when an error occurs in a controller it forwards the correct code to the client.

BenWhitten-RV:
On this note SubscriptionManager.cpp can currently return a REST specific response code
```c++
    // Check for duplicate subscription
    if (!isWildcard(oid) && subscriptions_.find(oid) != subscriptions_.end()) {
        rc = catena::exception_with_status("Subscription already exists for OID: " + oid,
                                         catena::StatusCode::NO_CONTENT);
        subscriptionLock_.unlock();
        return false;
    }
```
As subscription manger is a common class, we should fix this to be a code shared between both APIs.

PR #606, opened 2025-05-09
Refactor Socket/SSEWriter to just have one write function

ZuhayrSarkerRV:
I also renamed it to be called "sendResponse". I don't know going back to calling it write is more ideal - if that is the case, I can make that change. I just thought sendResponse was more fitting for REST.

Worth noting that the response codes being sent might be incorrect/unimplemented, but that's going to be tackled in #580 

Additions from Ben's review:
- Remove response_ var from SocketWriter/SSEWriter
- Removed "multi" logic from SocketWriter as it is handled in SSE
- ServiceImpl sends 204 status code if in OPTIONS
- Response in writers are written if < 300
- Headers are only written once in SSE at the start (although this is how it originally was, anyway)
- Removed CORS_, no need for it anymore
- Reverse msg/error so that you can just do sendResponse(rc) if theres no message to write (msg defaults to Catena::Empty())


BenWhitten-RV (changes requested):

sdks/cpp/connections/REST/src/SocketWriter.cpp
```diff
@@ -1,15 +1,14 @@
...
     // Converting the value to JSON.
     std::string jsonOutput;
     google::protobuf::util::JsonPrintOptions options; // Default options
     auto status = MessageToJsonString(msg, &jsonOutput, options);
-    
+    auto httpStatus = codeMap_.at(err.status);
+
     // Adding JSON obj to response_ if conversion was successful.
     if (status.ok()) {
         if (response_.empty()) {
```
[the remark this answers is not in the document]
BenWhitten-RV: I couldn't comment on the core below this is not about httpStatus lol

sdks/cpp/connections/REST/src/SocketWriter.cpp
```diff
@@ -19,22 +18,12 @@ void SocketWriter::write(const google::protobuf::Message& msg) {
...
+        // If conversion failed, set the error status to invalid argument, 
+        // which maps to bad request in the codeMap_
+        httpStatus = codeMap_.at(catena::StatusCode::INVALID_ARGUMENT);
     }
     
-    // Finishes the writing process with an error message.
-    auto httpStatus = codeMap_.at(err.status);
-    
     //If response is empty but status is ok, set rc to 204
     if (response_.empty() && err.status == catena::StatusCode::OK) {
```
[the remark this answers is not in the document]
BenWhitten-RV: also because I couldn't comment on the code below

sdks/cpp/connections/REST/src/SocketWriter.cpp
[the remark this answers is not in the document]
BenWhitten-RV: You can just print jsonOutput if we initialize it with like "". Maybe even have the writer check that the status is 200 before converting to stringified json so that we don't write "{}" with everything

sdks/cpp/connections/REST/src/SocketWriter.cpp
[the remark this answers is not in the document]
ZuhayrSarkerRV: I was going to deal with the 204 thing in a separate branch but I guess I can get this one overwith here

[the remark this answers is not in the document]
ZuhayrSarkerRV: Fixed by doing < 300

sdks/cpp/connections/REST/src/SocketWriter.cpp
```diff
@@ -1,96 +1,66 @@
...
+        jsonOutput = "";
     }
 
-    if (!response_.empty() && multi_) {
-        response_ = "{\"response\":[" + response_ + "]}";
+    // Only write response if we have valid data
+    if (httpStatus.first < 300 && !jsonOutput.empty()) {
+        response << "Content-Length: " << jsonOutput.length() << "\r\n\r\n"
+                << jsonOutput;
+        boost::asio::write(socket_, boost::asio::buffer(response.str()));
```
BenWhitten-RV, line 33: The call to boost::asio::write needs to be outside of the if statement to write errors.

If statement should also pass if the status is 204 (No Content)

sdks/cpp/connections/REST/src/SocketWriter.cpp
```diff
@@ -1,96 +1,66 @@
-
 #include <SocketWriter.h>
 using catena::REST::SocketWriter;
 using catena::REST::SSEWriter;
 
-void SocketWriter::write(const google::protobuf::Message& msg) {
-    // Adds a JSON message to the end of the response.
-    // Converting the value to JSON.
-    std::string jsonOutput;
+void SocketWriter::sendResponse(const catena::exception_with_status& err, const google::protobuf::Message& msg) {
```
BenWhitten-RV, line 5: Small issue with the order the events are in,

It should go declare variables -> convert message to JSON -> write Headers -> appendJson -> write to Socket

sdks/cpp/connections/REST/src/controllers/DeviceRequest.cpp
```diff
@@ -51,16 +51,16 @@ void DeviceRequest::proceed() {
             std::lock_guard lg(dm_.mutex());
             component = serializer_->getNext();
             }
-            writer_.write(component);
+            writer_.sendResponse(rc, component);
         }
-        writer_.finish(catena::Empty(), catena::exception_with_status("", catena::StatusCode::OK));
+        writer_.sendResponse(rc);
```
BenWhitten-RV, line 56: Do we need to write OK at the end? I think we should be fine to remove this line

sdks/cpp/connections/REST/src/controllers/BasicParamInfoRequest.cpp
```diff
@@ -176,29 +171,24 @@ void BasicParamInfoRequest::proceed() {
...
         }
     } catch (catena::exception_with_status& err) {
         rc_ = std::move(err);
-        finish();
     } catch (...) {
         rc_ = catena::exception_with_status("Unknown error in BasicParamInfoRequest", catena::StatusCode::UNKNOWN);
-        finish();
     }
+
+    finish();
```
BenWhitten-RV, line 182: This is called in the serviceImpl

BenWhitten-RV (changes requested):

sdks/cpp/connections/REST/src/SocketWriter.cpp
ZuhayrSarkerRV: Moved the write and made it check if the original message was empty in order to suppress the response

sdks/cpp/connections/REST/src/SocketWriter.cpp
ZuhayrSarkerRV: The headers have nothing to do with the message, you can write the headers first and then convert the message later, it doesn't matter. They are done separately and then the response is just appended to the final response. This makes the code easier to maintain and follow, imo.

NR-RV (approved):

PR #670, opened 2025-05-21
Fix REST BPIR no-response bug + additional error handling

ZuhayrSarkerRV:
The main fix was in finish() where I added the following:

``` c++
    if (responses_.empty()) {
        // If no responses, send at least one response with the error status
        writer_.sendResponse(rc_);
    }
```

Before, if responses didn't even exist because of some error, it would just not send anything at all (not even a response code, which broke unit testing)

Additionally, I went in and added extra error handling.

### DIFF
--- a/sdks/cpp/connections/REST/src/controllers/BasicParamInfoRequest.cpp
+++ b/sdks/cpp/connections/REST/src/controllers/BasicParamInfoRequest.cpp
@@ -65,9 +65,14 @@ BasicParamInfoRequest::BasicParamInfoRequest(tcp::socket& socket, ISocketReader&
         } else {
             oid_prefix_ = "/" + oid_prefix_value;
         }
-    // Parse error
+    } catch (const catena::exception_with_status& err) {
+        rc_ = catena::exception_with_status(err.what(), err.status);
+        finish();
+    } catch (const std::exception& e) {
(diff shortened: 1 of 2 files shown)


sdks/cpp/connections/REST/src/controllers/BasicParamInfoRequest.cpp
```diff
@@ -185,10 +250,15 @@ void BasicParamInfoRequest::finish() {
     std::cout << "BasicParamInfoRequest[" << objectId_ << "] finished\n";
     
     writer_lock_.lock();
```
BenWhitten-RV, line 252: Writer is not async so this lock is unnecessary. If it was necessary you should use std::lock_guard instead to be extra safe in the case of an error within the writer_ exiting the function (i.e. .unlock() is never triggered).

ZuhayrSarkerRV: the writer lock is something that carried over from grpc, it seems its not needed in REST

ZuhayrSarkerRV:
> Also I havent looked at your test cases but for the most part I've found it's better to get field params and stuff in proceed() so you can change them between the call to setUp() and proceed().

Yes, I have to do it in proceed, it doesn't even work if I only do it off setup.

BenWhitten-RV (approved):

Issue #673, opened 2025-05-22
Fix REST endpoints


ctwarog-ross:
Remove /device and change http to https

ZuhayrSarkerRV:
Closed by #677

PR #677, opened 2025-05-22
Fixed front end and assets endpoints

ctwarog-ross:
Issue #673

Removed /device prefix from yaml
Fixed fqoid regex checking
Fixed params and comamnds definition in device_schema SocketReader now correctly handles args in request url ServiceImpl asset request matches schema
fqoid variable and getter added to SocketReader


NR-RV (commented):
Failed unit test
RESTSocketReaderTests.SocketReader_DetailLevelUnset needs to be updated

PR #688, opened 2025-05-27
gRPC MultiSetValue and SetValue Unit Tests

BenWhitten-RV:
THIS IS A BRANCH OF PR #684 , EITHER REVIEW THAT PR FIRST OR IGNORE IT ENTIRELY AND JUST REVIEW THIS PR

**Changelog:**
- Rearranged the MultiSetValue proceed() function to better match preferred style of having only 1 return (or in this case write).
- Created unit tests for gRPC::MultiSetValue with 93% line coverage.
    - The remaining 8% cannot be accessed as it is code locked behind a state that _should_ be impossible to reach without manual manipulation of a private variable. Would could make the MultiSetValue class a friend of the testFixtureClass but from what I know friendships are bad practice.
- Created 2 unit test for gRPC::SetValue with 100% line coverage (just tests overloaded functions in creation and normal case).
- Updated MockServer to contain additional function for what to expect in new CallData object creation. This reduces line repetition across test files.
- Updated GetValue_test to better matched the improved MultiSetValue_test file. This has two notable improvements:
    - Less line repetition as values are now tested in testRPC (formerly makeRPC).
    - Function input testing to ensure the process is passing in the correct values depending on normal and authz cases.
- See PR #684 for additional changes.


sdks/cpp/connections/gRPC/src/controllers/MultiSetValue.cpp
```diff
@@ -130,7 +127,10 @@ void MultiSetValue::proceed(bool ok) {
             std::cout << typeName << "[" << objectId_ << "] finished\n";
             service_->deregisterItem(this);
             break;
-        // default: Error, end process.
+        /*
+         * default: Error, end process.
+         * This should be impossible to reach.
+         */
         default:
             status_ = CallStatus::kFinish;
```
NR-RV, line 135: added Exclusion Marker for the unreachable default state
```suggestion
        default:// GCOVR_EXCL_START
            status_ = CallStatus::kFinish;
            grpc::Status errorStatus(grpc::StatusCode::INTERNAL, "illegal state");
            responder_.FinishWithError(errorStatus, this);
            // GCOVR_EXCL_STOP
```

PR #699, opened 2025-05-29
Implement Common Connect Unit Tests

ZuhayrSarkerRV:
## Intro
- I meant to work on REST Connect, but this needed attention first, so I will tackle that in a different branch.
- This PR did make me change some of the implementation, particularly with changing language pack to use ILanguagePack instead of the Protobuf component (more info at end of changes section).

## 🔍 Coverage Statistics

|   | Status | Exec/Total |  Coverage |
| ------------- | ------------- | ------------- | ------------- |
| 🟩 LINES | ✅ Good | 65/70  | 92.9  |
| 🟩 FUNCTIONS  | ✅ Full Coverage | 10/10 | 100  |
| 🟨 BRANCHES  | ❗Needs work | 70/108 | 64.8 |

#### Main focus: 

image
- It is not possible to cover the lambdas themselves, but this is fine, the logic itself is still covered line-to-line

image
- To cover the throw in language pack required me to make slight interface changes (refer to end of changes section)

#### Additionally handled:

image

Work needs to be done on covering all the branches. I can look into this further when writing tests for Connect in gRPC/REST. 

#### Class Setup
- Using a monitor and operator token that I borrowed from Authorization_test.cpp in order to help test authorization 
- Set the base detail level to FULL
- **Helper Functions**
  - `void setupCommonExpectations(mockparam, mockdescriptor)`
    - Set up scope, detail level, descriptor, minimal set and subscription expectations for use in relevant tests
  - `void setupMockParam(mockparam, oid, mockdescriptor)`
    - Sets up the param's oid, descriptor, and if it is an array, then calls setupCommonExpectations against it.
  - `unique_ptr<ILanguagePack> setupLanguagePack()`
    - Just sets up a basic mock language pack for testing, doesn't need to be fully implemented

The following unit tests determine if an update is or isn't expected from Connect (on a parameter or language pack)

#### Authorization Tests
- 1.1 🔴 FALSE: Param updateResponse readAuthz check, NO UPDATE
- 1.2 🟢 TRUE: Param updateResponse auth check, no authz, UPDATES
- 1.3 🔴 FALSE: Param updateResponse auth check, authz, NO UPDATE
- 1.4 🟢 TRUE: Param updateResponse auth check, authz, UPDATES
- 1.5 🟢 TRUE: LangPack updateResponse auth check, no authz, UPDATES
- 1.6 🔴 TRUE: LangPack updateResponse auth check, missing monitor permissions, NO UPDATE
[...]

Issue #697, opened 2025-05-29
UT Refactor


ctwarog-ross:
Refactoring REST Unit tests to abstract helper methods behind a unified UT class

PR #723, opened 2025-06-05
REST Response Code and Endpoint Patch Job

ZuhayrSarkerRV:
## 📃 Intro

Okay, so I'm labelling this PR as a "patch job" because I'm thoroughly convinced our work isn't done.

Now I know what you're thinking... "40 FILES?!?!?" (maybe even more as I progress in this pr).

HEAR ME OUT.

Most of it is literally just copy-paste simple fixes. But for what isn't...

This started as me trying to fix response codes. This required me to update unit tests. Then I realized I needed to test these new changes, so the when I went over to Postman/Swagger, I realized the endpoints were broken. I noticed inconsistencies in the open api definition, and linking was broken altogether. 

What started as a somewhat large issue transformed into a behemoth bred from slight mismanagement (not pointing fingers).

I also think there needs to be some additional testing done, because I'm still not sure everything in Open API is correctly linked.

As such, I think it's best to leave this to be the last PR to merge in. In fact, I'm quite literally going to label it as DO NOT MERGE so that someone doesn't accidentally merge it in - *cough cough* #699. I need to wait for every other PR, particularly #709 and #697 to go in.

There were also other unit tests that got updated after I updated some response codes -I made sure to check that they all had 100% line coverage:

image

So now for the actual changes! 

--- 

#### Redundant file removal
- Deleted old files that are no longer used:
  - DELETE AddLanguage.yaml, covered in languages.yaml
  - DELETE DeviceRequest.yaml, covered in Devices.yaml
  - DELETE ExecuteCommand.yaml, covered in Command.yaml
  - DELETE GetParam.yaml, covered in Param.yaml
  - DELETE GetValue.yaml, covered in Value.yaml
  - DELETE UpdateSubscriptions.yaml, covered in Subscriptions.yaml

####  The repetitive fixes:
- Added `400 Bad Request` to a bunch of the yaml files, this was missing
- Some YAMLs were missing response codes 400, 401, 403, 404, 500, 503

#### openapi.yaml changes
- Removed "content" from any response codes that are > 300, there should not be a content section for these, it violates RESTful api design

####  Device.cpp changes
- Handle getValue internal or unknown
- Handle getLanguagePack error handling with empty language ID, internal, or unknown

####  ServiceImpl.cpp
- Fix endpoints to match OpenAPI spec

####  Other REST source files 
- AddLanguage, add std::exception case that throws internal status code
  - Update UT to match
- AssetRequest, add additional error handling with files, this may conflict with @ctwarog-ross 's work
- Connect, add error handling in constructor for fields (catena, std, ... )
  - Also added some response handling in proceed()
  - Organized flow for disconnect to be a little less "scuffed" 
  - Update UT to match
- DeviceRequest, add authorization checks, also updated UT
  - Added additional error handling like throwing if socket is closed during serializing 
- The following source files all had additional response handling in the form of adding an std exception check and in some case catena exception (instead of just generic catch (...) )
  - ExecuteCommand
  - GetParam
  - GetValue
  - LanguagePackRequest
  - ListLanguages
- Update all the unit test files to cover the newly added response codes

####  Other fixes
- Fixed JSON error in Command,yaml
- Added comment in GetPopulatedSlots.yaml, this is NOT implemented, perhaps discuss this or open an issue ?

---

## UPDATES 6/16/2025
- Pushed a fix to correct the build commands in tasks.json
- Fixed a few endpoints, namely basicparaminforequest, to communicate with open api def
- There were a few small fixes done during lifespan of this PR (not directly in this one) such as by @BenWhitten-RV to fix multi-set and setvalue
- Updated some unit tests to match some of the endpoint fixes

### SUMMARY OF ENDPOINT PROGRESS:

The following were tested with authz off (as I have no idea how to operate authz on OpenAPI)

**Legend**:
❗- Needs Attention
❓- Confusion/Question
⚠️ - Unimplemented
✅ - Works 
❌ - Fails

## Discovery
-❗**GET/devices**
  - Works fine, but I don't think the slot # works.

- ❗**GET/{slot}**
  - I don't understand how this is meant to work. IT doesn't listen to whatever slot I Set it to, but it still sends all the devices related to the slots they've been set to.
    - If I do slot 3, but theres a device on slot 1, it sends slot 1's device
    - Oddly, if I do GET/Devices, it says this device is on slot 0 - I think there's some missing work with slots overall

- ❗**GET/{slot}/stream**
  - Seems to work aside from being the incorrect slot # to the one device I requested it for

## Commands
-❗**POST/{slot}/commands/{fqoid}**
  - Works, though can't test response because we don't have it implemented in our examples

## Assets
- ❓**GET/{slot}/asset/{fqoid}**
  - Seemingly works, I wasn't able to convert the utf to an image, though.
- ⚠️ **POST/{slot}/asset/{fqoid}** - UNIMPLEMENTED
- ⚠️ **PUT/{slot}/asset/{fqoid}** - UNIMPLEMENTED
- ⚠️ **DELETE/{slot}/asset/{fqoid}** - UNIMPLEMENTED
- ⚠️ **GET/{slot}/asset/{fqoid}/stream** - UNIMPLEMENTED

## Parameters
- ✅**GET /{slot}/basic-param/{fqoid}**
- ✅ **GET /{slot}/asset/{fqoid}/stream**
- ✅ **GET /{slot}/value/{fqoid}**
- ✅ **PUT /{slot}/value/{fqoid}**
- ✅ **PUT /{slot}/values**
- ❗ **GET /{slot}/subscriptions**, **PUT /{slot}/subscriptions**, **GET /{slot}/subscriptions/stream**
  - Covered in PR #741
- ✅ **PUT /{slot}/subscriptions/stream**

## Subscriptions
- ✅ All of these are already covered in parameters, should this even exist as a section or should we organize it better

## Updates
-❗ **GET /connect**
  - I can't test subscriptions detail level without PR #741, but it otherwise seems to work.

## Languages
-❗All of these are dependent on PR #746


sdks/cpp/common/src/Device.cpp
```diff
@@ -159,20 +159,34 @@ catena::exception_with_status Device::getValue (const std::string& jptr, catena:
...
         ans = catena::exception_with_status(why.what(), why.status);
+    } catch (const std::exception& e) {
+        ans = catena::exception_with_status(e.what(), catena::StatusCode::INTERNAL);
+    } catch (...) {
+        ans = catena::exception_with_status("Unknown error", catena::StatusCode::UNKNOWN);
     }
     return ans;
 }
 
 catena::exception_with_status Device::getLanguagePack(const std::string& languageId, ComponentLanguagePack& pack) const {
```
BenWhitten-RV, line 170: The if statement checking if the id is empty is fine but the other changes are unnecessary and bloat the function. The try catch surrounds code that can not fail if the language pack is present and ans will already be OK on line 183 if it makes it that far, so setting it again is redundant.

I would prefer this to be a single if -> else if -> else.

sdks/cpp/common/src/Device.cpp
```diff
@@ -207,6 +221,12 @@ std::unique_ptr<IParam> Device::getParam(const std::string& fqoid, catena::excep
     } catch (const catena::exception_with_status& why) {
         status = catena::exception_with_status(why.what(), why.status);
         return nullptr;
+    } catch (const std::exception& e) {
```
BenWhitten-RV, line 224: Do we really need to handle std::exceptions? Also this should be updated to have a single return statement.

ZuhayrSarkerRV:
My next task here is to test every endpoint in OpenAPI and mark the ones that are unimplemented or fix any that are incorrectly linked.

sdks/cpp/connections/REST/src/controllers/Connect.cpp
```diff
@@ -11,13 +11,24 @@ Connect::Connect(tcp::socket& socket, ISocketReader& context, IDevice& dm) :
...
+        // Parsing fields and assigning to respective variables.
+        userAgent_ = context.fields("user_agent");
+        forceConnection_ = context.hasField("force_connection");
 
-    // Set detail level from context
-    detailLevel_ = context_.detailLevel();
-    dm_.detail_level(detailLevel_);
+        // Set detail level from context
+        detailLevel_ = context_.detailLevel();
+        dm_.detail_level(detailLevel_);
```
BenWhitten-RV, line 22: This code cannot fail so the try catch is not necessary, Also dm_ should not govern the detail level of API calls, it should (and I believe it is already) handled by the individual API call to avoid it a race condition between clients.

sdks/cpp/connections/REST/src/controllers/Connect.cpp
```diff
@@ -61,17 +80,54 @@ void Connect::proceed() {
         cv_.wait(lock, [this] { return hasUpdate_; });
         hasUpdate_ = false;
         writeConsole_(CallStatus::kWrite, true);
+        
+        if (!socket_.is_open() || shutdown_) {
```
BenWhitten-RV, line 84: Unnecessary if statement. Errors close the socket, which means the while loop will finish while unlocking the lock at the end.

sdks/cpp/connections/REST/src/controllers/Connect.cpp
```diff
@@ -61,17 +80,54 @@ void Connect::proceed() {
...
+
         try {
-            if (socket_.is_open() && !shutdown_) {
-                res_.set_slot(dm_.slot());
-                writer_.sendResponse(catena::exception_with_status("", catena::StatusCode::OK), res_);
+            res_.set_slot(dm_.slot());
+            writer_.sendResponse(catena::exception_with_status("", catena::StatusCode::OK), res_);
+        } catch (const std::exception& e) {
+            // For errors, just send the status code without a response body
+            try {
```
BenWhitten-RV, line 94: We really don't need this heavy duty a try catch system. If an error occurs in socketWriter its handled internally. If it fails to write then the socket is closed.

sdks/cpp/connections/REST/src/controllers/Connect.cpp
```diff
@@ -61,17 +80,54 @@ void Connect::proceed() {
...
+                // If we can't send the error response, then the socket is truly dead
+            }
             socket_.close();
+            break;
         }
         lock.unlock();
     }
+
+    // If we get here, the connection is ending
+    if (socket_.is_open()) {
```
BenWhitten-RV, line 119: This code does nothing in the context of SSE.  By this point the headers will already be written by the sendResponse on line 61

sdks/cpp/connections/REST/src/controllers/Connect.cpp
```diff
@@ -83,9 +139,14 @@ void Connect::finish() {
         dm_.languageAddedPushUpdate.disconnect(languageAddedId_);
     // Listener not yet initialized.
     } catch (...) {}
-    // Finishing and closing the socket.
+    
+    // Only attempt to send response if socket is still open
     if (socket_.is_open()) {
```
BenWhitten-RV, line 144: I believe this is also a holdover from the pre SSE socketWriter when you needed to send an finished code to the client so they don't get an error. You can probably remove this.

sdks/cpp/connections/REST/src/controllers/DeviceRequest.cpp
```diff
@@ -24,9 +24,16 @@ void DeviceRequest::proceed() {
         
         // Setting up authorizer object.
         if (context_.authorizationEnabled()) {
```
BenWhitten-RV, line 26: I don't like these changes to device request. A bunch of unnecessary try catches when the flow was fine before.

sdks/cpp/connections/REST/src/controllers/ExecuteCommand.cpp
```diff
@@ -54,9 +54,17 @@ void ExecuteCommand::proceed() {
                 } else {
                     while (responder->hasMore()) {
                         writeConsole_(CallStatus::kWrite, socket_.is_open());
-                        catena::CommandResponse res = responder->getNext();
-                        if (respond) {
-                            writer_.sendResponse(rc, res);
+                        try {
```
BenWhitten-RV, line 57: Again the flow was fine before we don't need the extra try catch.

BenWhitten-RV (changes requested):

BenWhitten-RV:
> My next task here is to test every endpoint in OpenAPI and mark the ones that are unimplemented or fix any that are incorrectly linked.

I'm currently working on a socketReader fix so not all endpoints will work on openAPI atm.

ZuhayrSarkerRV:
> > My next task here is to test every endpoint in OpenAPI and mark the ones that are unimplemented or fix any that are incorrectly linked.
> 
> I'm currently working on a socketReader fix so not all endpoints will work on openAPI atm.

Yep, I realized that mid-test.

sdks/cpp/connections/REST/src/controllers/Connect.cpp
ZuhayrSarkerRV: Yep, that's a good call on the detail level check, this was a side-effect of me struggling to set detail level in a seperate issue, but it should not be handled here.

ZuhayrSarkerRV:
I need #724 merged before I can continue with this PR.

ZuhayrSarkerRV:
BasicParamInfoRequest's url/endpoint is not parsing correctly which is causing this PR to stagnate, and because this PR involved testing that endpoints work, I need to fix that before I can continue.

ZuhayrSarkerRV:
I fixed BPIR's endpoints and implementation as well the vscode build commands. 

I need to return to testing/documenting the remaining endpoints and then I can finally mark this ready for review.

sdks/cpp/connections/REST/src/controllers/Connect.cpp
```diff
@@ -61,17 +70,13 @@ void Connect::proceed() {
...
-            if (socket_.is_open() && !shutdown_) {
-                res_.set_slot(dm_.slot());
-                writer_.sendResponse(catena::exception_with_status("", catena::StatusCode::OK), res_);
-            }
-        // A little scuffed but I have no idea how else to detect disconnect.
-        } catch (...) {
-            socket_.close();
-        }
+
+        res_.set_slot(dm_.slot());
```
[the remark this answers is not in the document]
BenWhitten-RV: This still applies

PR #737, opened 2025-06-10
Fixed a number of errors flagged when running examples using valgrind.

BenWhitten-RV:
**Overview:**
- Brought valgrind errors when running gRPC examples down to a consistent 2 (used to be upwards of 60), which can be assumed to be issues on the side of the gRPC and protobuf libraries.
- Brought valgrind errors when running REST examples down to a consistent 4 (imagine it used to be much more), which can be assumed to be issues on the side of the boost and protobuf libraries.

**Changelog:**
- Updated detached threads in example services to instead join() after the server has shut down.
    - Before the threads would sometimes finish after global scope resulting in calls to a non existent device and service.
- Updated connect endpoints signal ids to have default values that are checked before disconnecting.
    - Namely an issue on the gRPC side resulting in the unused Connect RPC to disconnect from signals it is not connected to.
- Updated gRPC service implementation's processEvents() function to wait for all CallData objects in the registry to get deregistered before returning.
    - Similar to the thread issue listed above, the CallData objects would sometimes be destroyed after the service resulting in a call to a non-existent DeregisterItem() function.

### DIFF
--- a/sdks/cpp/connections/REST/examples/structs_with_authz_REST/structs_with_authz_REST.cpp
+++ b/sdks/cpp/connections/REST/examples/structs_with_authz_REST/structs_with_authz_REST.cpp
@@ -124,7 +124,9 @@ void RunRESTServer() {
             std::string front = jptr.front_as_string();
             jptr.pop();
 
-            handlers[front](jptr.toString(), p, idx);
+            if (handlers.contains(front)) {
+                handlers[front](jptr.toString(), p, idx);
+            }
         });
         
         api.run();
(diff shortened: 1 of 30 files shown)


sdks/cpp/connections/REST/examples/structs_with_authz_REST/structs_with_authz_REST.cpp
```diff
@@ -124,7 +124,9 @@ void RunRESTServer() {
             std::string front = jptr.front_as_string();
             jptr.pop();
 
-            handlers[front](jptr.toString(), p, idx);
+            if (handlers.contains(front)) {
```
ctwarog-ross, line 127: good catch

PR #738, opened 2025-06-10
Updated Java SDK to fix breaking changes

johndanenRV:
Java SDK now uses server-side command streaming instead of bidirectional command streaming. Also fixed a few other small breaking changes.

This pull request does not fix the more recent breaking changes like removing StructField


sdks/java/src/main/java/com/rossvideo/catena/example/client/CommandResponseHandler.java
rukaury, line 1: I see a lot off system error prints, we should use a logger like log4j. Maybe something to consider creating a ticket for.

johndanenRV: It probably would be better practice to use a logger but so far, the codebase only uses system prints and I'm just trying to get the breaking changes fixed. If we want to add logging we can, but that should probably be handled as a separate issue.

PR #741, opened 2025-06-11
Update Subscriptions Fixes

BenWhitten-RV:
A large PR containing changes and fixes for Update Subscriptions

### **Changelog:**
**Common**
- Updated flow of SubscriptionManager functions to have 1 return statement.
- Changed out manual lock in SubscriptionManager functions for a lockguard (cleaner and safer)
- Fixed wildcard support for SubscriptionManager::addSubscriptions() (would previous break from not finding a parameter)
- Added read authz checking to SubscriptionManager::addSubscriptions()
- Added isSubscribed() function to SubscriptionManager for use in Connect
- Changed getAllSubscribedOids to return a hard copy rather than the actual object.
    - Hypothetically the subscribedOids could be changed while updateSubs or deviceReq are returning params, so I think this is safer. **_Up for discussion_**.

**gRPC**
- Suppressed "oid not found" errors in updateSubscriptions.
- Updated flow of gRPC::updateSubscriptions::proceed() to have 1 return and better utilize the state machine.
- Removed unused updateSubscription member variables and functions

**REST**
- Suppressed "oid not found" errors in updateSubscriptions.
- Renamed updateSubscriptions to Subscriptions.
- Updated flow of REST::updateSubscriptions::proceed() and implemented GET.
    - More testable now.
- Added support for /stream flag to GET Subscriptions.
- Removed unused updateSubscription member variables and functions (mostly artifacts from converting from gRPC to REST)


BenWhitten-RV:
The change to return a hard copy is costly for connect since you would need to return a copy for every call to Connect. I just had the idea to implement an isSubbed function to get around this though so I'm going to go do that.

ZuhayrSarkerRV:
Also, is it ideal to name the file Subscriptions instead of updateSubscriptions? What was the thought process behind this? Did it get past someone ?

NR-RV (approved):

PR #752, opened 2025-06-16
Refactor gRPC Unit Tests

ctwarog-ross:
Issue # 751

Server now starts and shuts down for every test case setup/teardown code inherited from common GRPCTest class


BenWhitten-RV (approved):
Looks good. Tested 20 times without issue.

PR #763, opened 2025-06-23
REST Unit Test Refactor

BenWhitten-RV:
**Changelog:**
- Created RESTEndpointTest class which (similar to the gRPCTest class) contains a number of variables and expectations used across all REST unit tests.
- Updated REST unit tests to use this new class.
- Cleaned up many of the REST unit tests.
- Removed unnecessary unit tests from DeviceRequest and added one which tests a stream response.
- Refactored REST Param Info Request unit tests saving ~200 lines.
- Refactored REST Param Info Endpoint slightly to achieve 100% coverage.
    - We previously had many redundant, nested try-catches where the same behaviour could be achieved by just throwing to the outer try-catch.
- Included ExecuteCommand REST unit tests.
- Moved MockDeviceSerializer into its own file (similar to MockCommandResponder)


BenWhitten-RV:
Doing a quick quality check first don't review yet

NR-RV (approved):

PR #774, opened 2025-06-24
Small UpdateSubs bug fix + other misc. things

ZuhayrSarkerRV:
This is a really small bug fix just to make it so it correctly handles update subscriptions setting both a add and remove simultaneously. I also updated the unit test files.

In the REST subscriptions unit test file I also explicitly test adding and removing. The old method of hardcoding 0 and 2 isn't really modular or good unit test design, in my opinon (feel free to disagree). 

I also added .gcov to gitignore because it was weird trying to sift my git changes whilst seeing the gcov files during the coverage build process.

Then there's me just tweaking the tasks.json which is just my own preferential thing. I'm honestly with the idea that .vscode shouldn't even be shared on the codebase, but whatever.

### DIFF
--- a/sdks/cpp/connections/REST/src/controllers/Subscriptions.cpp
+++ b/sdks/cpp/connections/REST/src/controllers/Subscriptions.cpp
@@ -39,7 +39,7 @@ Subscriptions::Subscriptions(tcp::socket& socket, ISocketReader& context, IDevic
     : socket_(socket), context_(context), dm_(dm) {
     
     // GET (stream)
-    if (context.method() == "GET" && context.stream()) {
+    if (context.method() == Method_GET && context.stream()) {
         writer_ = std::make_unique<SSEWriter>(socket_, context_.origin());
     // GET (no stream) or PUT
     } else {
@@ -95,13 +95,25 @@ void Subscriptions::proceed() {
                     rc = catena::exception_with_status("Failed to parse JSON Body", catena::StatusCode::INVALID_ARGUMENT);
             
                 } else {
-                    // Processing removed OIDs
+                    // Processing removed OIDs first
                     for (const auto& oid : req.removed_oids()) {
                         context_.getSubscriptionManager().removeSubscription(oid, dm_, supressErr);
                     }
-                    // Processing added OIDs
+                    // Processing added OIDs, but skip any that are also in removed_oids
                     for (const auto& oid : req.added_oids()) {
-                        context_.getSubscriptionManager().addSubscription(oid, dm_, supressErr, *authz);
+                        // Check if this OID is also in removed_oids
+                        bool isAlsoRemoved = false;
+                        for (const auto& removedOid : req.removed_oids()) {
+                            if (oid == removedOid) {
+                                isAlsoRemoved = true;
+                                break;
+                            }
+                        }
+                        
+                        // Only add if it's not also being removed
+                        if (!isAlsoRemoved) {
+                            context_.getSubscriptionManager().addSubscription(oid, dm_, supressErr, *authz);
+                        }
                     }
                 }
 
--- a/unittests/cpp/REST/tests/Subscriptions_test.cpp
+++ b/unittests/cpp/REST/tests/Subscriptions_test.cpp
@@ -15,7 +15,7 @@
  * contributors may be used to endorse or promote products derived from this
  * software without specific prior written permission.
  *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS “AS IS”
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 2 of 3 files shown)


sdks/cpp/connections/REST/src/controllers/Subscriptions.cpp
```diff
@@ -95,13 +95,25 @@ void Subscriptions::proceed() {
                     rc = catena::exception_with_status("Failed to parse JSON Body", catena::StatusCode::INVALID_ARGUMENT);
             
                 } else {
-                    // Processing removed OIDs
+                    // Processing removed OIDs first
                     for (const auto& oid : req.removed_oids()) {
                         context_.getSubscriptionManager().removeSubscription(oid, dm_, supressErr);
                     }
-                    // Processing added OIDs
+                    // Processing added OIDs, but skip any that are also in removed_oids
```
BenWhitten-RV, line 102: This does not account for wildcards. Something like:

Removed: {"/\*"}
Added: {"/counter"}

Will still add counter to subscriptions.

If we want remove to have priority we can just move it to be after added_oids.

---

### **Below this is kinda confusing so read at your own risk**
If we want to ignore duplicate entries it gets a little more complex due to things like nested wildcards and already subscribed oids. For example, if we're subscribed to {"/counter", "/counter/0"} and call:

Removed: {"/\*}
Added {"/counter/\*"}

I subscribed oids at the end should be {"/counter/0"}.

Maybe some sort of sanitize function could accomplish this by comparing three lists: added_oids, removed_oids, and existing_oids

Removed_oids is out to-remove params.

Existing_oids is our currently subscribed params. If it exists on removed_oids and not added_oids then remove it, otherwise do nothing.

Added_oids is our to-add params. After doing what we need to with existing_oids, if the oid does not exist in removed_oids then we add it.

unittests/cpp/REST/tests/Subscriptions_test.cpp
```diff
@@ -159,18 +159,26 @@ class RESTSubscriptionsTests : public RESTEndpointTest {
...
-                    EXPECT_EQ(addedOids_, 0);
-                } else {
-                    EXPECT_EQ(addedOids_, 2);
-                }
-                // We should expect either 0 or 2 removed oids.
-                if (inVal_.removed_oids().empty()) {
-                    EXPECT_EQ(removedOids_, 0);
-                } else {
-                    EXPECT_EQ(removedOids_, 2);
+                // Calculate expected counts based on Subscriptions.cpp logic
```
BenWhitten-RV, line 162: I initially added the counters to make sure x number of things were/weren't added but for some reason didn't think to just put it in the errParam and make it dynamic. Good catch.

sdks/cpp/connections/REST/src/controllers/Subscriptions.cpp
ZuhayrSarkerRV: I'm already aware that this doesn't account for wildcards. I think this is fine to merge in for now, but we're already missing some wildcard testing and even implemention. This is going to get modified regardless when I figure out the bug I'm dealing with right now.

ZuhayrSarkerRV: The fact of the matter is when I demo'd this API call to Nathan around a month or so ago, he made me test a bunch of edge cases and it's simply not sound. Wildcards are an entirely different beast.

BenWhitten-RV: How does it not take priority if you put if after? For clarification I'm not talking about in the request but in the code.

If you add first then remove you would end up removing any duplicates in the remove step.

ZuhayrSarkerRV:
I have taken this off review because I want to investigate the core issue, not work around it. It doesn't make sense why even with a lock existing in the manager there are concurrency issues. Checking if an oid has been added/removed is "expensive", but even if we don't care about the performance issues, its just weird code design.

sdks/cpp/connections/REST/src/controllers/Subscriptions.cpp
```diff
@@ -93,18 +93,17 @@ void Subscriptions::proceed() {
...
-                    }
-                    // Processing added OIDs
+                    // Process added OIDs
                     for (const auto& oid : req.added_oids()) {
-                        context_.getSubscriptionManager().addSubscription(oid, dm_, supressErr, *authz);
+                        bool added = context_.getSubscriptionManager().addSubscription(oid, dm_, supressErr, *authz);
+                    }
+                    // Process removed OIDs
+                    for (const auto& oid : req.removed_oids()) {
+                        bool removed = context_.getSubscriptionManager().removeSubscription(oid, dm_, supressErr);
```
ZuhayrSarkerRV, line 103: These booleans aren't actually used, but are there to make debugging easier. I'm going to keep them in there for now.

sdks/cpp/connections/REST/src/controllers/Subscriptions.cpp
ZuhayrSarkerRV: This issue has been resolved for specific params but not wildcards, which I am hoping to look into in a further issue. Marking as resolved.

BenWhitten-RV (approved):

PR #780, opened 2025-06-24
Refactor to remove RestTestHelpers 

ZuhayrSarkerRV:
This is a really tiny "fix", all I did was move the helpers over to the original unit test. I guess the reason why I didn't bundle it in another PR was because there is no work left to be done in ParamInfoRequest, it's complete.


ctwarog-ross (commented):
I don't think we should be doing this, we want to keep these methods open to other tests and keep unit test files to having just unit test code

PR #790, opened 2025-06-26
Fix ParamInfoRequest Unit Test

ZuhayrSarkerRV:
Refactor all instances of Path to use () instead of {}. This enforces regex validation. REST Unit Tests are now 100%, and I have tested ParamInfoRequest to check.

### 🛠️ Bug fixes:
- ParamVisitor Path{} is now replaced with Path()
  - Uses toString(true) to add a leading slash
- Removed pointless path variable from ParamVisitor 
  - This was the main culprit and caused the path to fail building due to REGEX failure (lack of leading slash)

### ⚒️ Other Changes:
- getTopLevelParams within device now uses Path(), but requires a leading slash to be added to each Param
  - This change has nothing to do with the bug but is an enhancement

There is more context in the comments

### DIFF
--- a/sdks/cpp/common/include/IDevice.h
+++ b/sdks/cpp/common/include/IDevice.h
@@ -404,5 +404,8 @@ class IDevice {
     vdk::signal<void(const std::string&, const IParam*)> valueSetByServer;
 };
 
+using SlotMap = std::unordered_map<uint32_t, IDevice*>;
+using SignalMap = std::unordered_map<uint32_t, uint32_t>;
+
 }  // namespace common
 }  // namespace catena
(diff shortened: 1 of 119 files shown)


ZuhayrSarkerRV:
I think the top level params unit test needs re-write. It doesn't catch poor path building somehow.

ZuhayrSarkerRV:
In this PR's current state everything is functioning, but I don't like that the unit test didn't catch such a critical error that quite literally broke the implementation (the path not being built correctly and sending a bad request).

ZuhayrSarkerRV:
So turns out the issue in the unit test was a false flag due to an internal failure that needs to be tested in #796 . This PR can be approved and closed.

ctwarog-ross (approved):

PR #806, opened 2025-07-03
Common Unit Tests - PolyglotText and Constraints

BenWhitten-RV:
**Changelog:**
- PolyglotText
    - Added missing doxygen comments to PolyglotText files
    - Replaced PolyglotText::toProto's catena::MessageLite input var with catena::PolyglotText
        - Only place in the codebase we use catena::MessageLite instead of the actual proto object.
     - Simplified and refactored various functions across PolyglotText to have 1 return statement.
     - Created 100% coverage unit tests for PolyglotText
     - Cleaned up headers
 - NamedChoiceConstraint
     - Added type check to NamedChoiceConstraint constructor to ensure it is not constructed with a not supported type.
     - Refactored NamedChoiceConstraint with IDevice constructor to reuse existing constructors.
     - Simplified and refactored NamedChoiceConstraint::satisfied to have 1 return statement.
     - Created 100% coverage unit tests for NamedChoiceConstraint
     - Cleaned up headers
     - Updated NamedChoiceConstraint::toProto to support STRING_CHOICE in addition to STRING_STRING_CHOICE
 - PicklistConstraint
     - Removed PickListConstraint and replaced code generation to create a NamedChoiceConstraint with an empty PolyglotText List Initializer.
 - RangeConstraint
     - Refactored RangeConstraint constructors to reuse existing constructors
     - Simplified and Refactored RangeConstraint::satisfied to have 1 return statement.
     - Fixed issue with RangeConstraint where apply would attempt to mod a float if step was set to 0 (undefined)
     - Created 100% coverage unit tests for RangeConstraint
     - Cleaned up headers
 
_PicklistConstraint and NamedChoiceConstraint are functionally the exact same, the only difference is the code generator uses PickListConstraint for strings and NamedChoiceConstraint for string arrays (and int arrays). Is there anything stopping us from just removing PickListConstraint and having the codegen instead use NamedChoiceConstraint for both?_

Update: Removed PickListConstraint


NR-RV:
@mejohnnaylor i think  NamedChoiceConstraint for both sounds good to me

ctwarog-ross (approved):

PR #816, opened 2025-07-07
Fixed param json schema structField issue

BenWhitten-RV:
**Changlog:**
- Removed StructField from the param JSON schema.
- Removed deprecated "Value" parameter from Struct definition in the param JSON schema.
- Updated code gen and examples to reflect this change.

Essentially, whereas previously a struct param within a device model might look like this:
```yaml
struct_value:
    type: STRUCT
    fields:
        name: {type: STRING}
        id: {type: INT}
    value:
        struct_value:
            fields:
                name: {value: {string_value: "name"}}
                id: {value: {int32_value: 1}}
```
It will now look like:
```yaml
struct_value:
    type: STRUCT
    fields:
        name: {type: STRING}
        id: {type: INT}
    value:
        struct_value:
            fields:
                name: {string_value: "name"}
                id: {int32_value: 1}
```

### DIFF
--- a/schema/catena.param_schema.json
+++ b/schema/catena.param_schema.json
@@ -642,11 +642,7 @@
             "struct_values": {
               "type": "array",
               "items": {
-                "properties": {
-                  "fields": {
-                    "$ref": "#/$defs/struct_fields"
-                  }
-                }
+                "$ref": "#/$defs/struct_value"
               }
             },
             "additionalItems": false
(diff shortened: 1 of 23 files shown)


ctwarog-ross (approved):

johndanenRV (approved):

mejohnnaylor:
I'm sorry but this won't work. The reason is that it's the proto message definition that drives the JSON schema. When a value is serialized from proto to JSON, because it's polymorphic, protobuf has to put in information to say what oneOf option is being used.

when the value is a StructValue, it puts "struct_value" into the serialization. StructValue is just a map of <string, Value>, so once the string has been used as a key name, proto has to serialize a Value object, which is again polymorphic, and needs it's type to be in the serialization. 

before merging, please serialize a protobuf binary structure object to JSON to confirm that the schema changes are correct. I fear that they are not

mejohnnaylor:
gah! too late!

PR #829, opened 2025-07-09
remove the version form the toolchain evn for git

NR-RV:
removed the version from the toolchain for git as 2.43 isnt a thing anymore 
(also the only real way to test this is to push it into dev and watch the action)


abrown-RV (approved):
Ya its just git. Best practice for ci/cd is to lock versions of things so its repeatable.... but its just git. Less concerned.

PR #847, opened 2025-07-17
Refactor Unit Tests to use a Helper Function for Tokens

ZuhayrSarkerRV:
There isn't really much to comment on this PR - I just added a helper to CommonTestHelpers to return a token based on a provided scope. At the moment, the order in which scopes are provided is hardcoded - unless we literally have a token generator in the codebase or put in every single possible combination (there are 8 scopes aside from kUndefined), this is kind of the only way to go about it. 

I also renamed two UT files so that they dont conflict with the renaming in #814 when that eventually gets merged.


ctwarog-ross (approved):
Nice work here, much needed cleanup

PR #851, opened 2025-07-18
Master merge 20250718

NR-RV:
new merge to main
now with a branch to manage conflicts


abrown-RV (approved):
We just want everything from develop. I watched Nathan resolve the conflicts and only include the one change on main

PR #853, opened 2025-07-21
THE Connect Update Part 1 (gRPC)

### LINKED ISSUE #831: add a limit for the number of connected clients
consider making it a command-line option with a sensible default

### LINKED ISSUE #833: add support for force connection in the connect request
the user needs to have st2138:adm:w privileges to succeed, otherwise, return not authorized

BenWhitten-RV:
The Fellowship of gRPC

**Changelog:**
- Added 4 new functions to the virtual common::Connect class.
    - priority() returns the connection's priority value, calculated as = scope * 2 + write + (adm:w AND forceConnection)
    - objectId() returns the object's id. This will be lower based on creation time.
    - The < operator returns true based on the object's priority. If priorities are equal, then it instead returns true based on which object has the lower objectId (and is thus older).
    - shutdown() forcefully shuts down the connection.
- Added maxConnections input variable to the gRPC::ServiceImpl class.
- Added 2 new functions to the gRPC::ServiceImpl class.
    - RegisterConnection() registers a connection to the connectionQueue. If maxConnections has been reached, it forcefully shuts down the lowest priority connection in the queue.
        - ConnectionQueue is actually just a vector with the connections sorted by priority... Was going to use an actual priority queue but that doesn't allow for direct access to objects, which is required when deregistering connections.
    - DeregisterConnection() deregister's a connection from the ConnectionQueue.
- Refactored gRPC::Connect and updated to now use register and deregister connection.
    - Refactor changes are based on various things learned from developing tests for gRPC::Connect
- Added maxConnections flag.
- Created MockConnect object
- Updated relevant unit tests.

### DIFF
--- a/sdks/cpp/common/include/rpc/Connect.h
+++ b/sdks/cpp/common/include/rpc/Connect.h
@@ -81,7 +81,7 @@ class Connect : public IConnect {
     /**
      * @brief Returns true if this has less prioirty than otherConnection.
      */
-    bool operator<(const IConnect& otherConnection) override {
+    bool operator<(const IConnect& otherConnection) const override {
         return priority_ < otherConnection.priority() ||
                (priority_ == otherConnection.priority() && age_ >= otherConnection.age());
     }
--- a/sdks/cpp/common/include/rpc/Connect.h
+++ b/sdks/cpp/common/include/rpc/Connect.h
@@ -52,7 +52,6 @@
 
 #include <string>
 #include <condition_variable>
-#include <chrono>
 
 namespace catena {
 namespace common {
@@ -75,15 +74,15 @@ class Connect : public IConnect {
      */
     uint32_t priority() const override { return priority_; }
     /**
-     * @brief Returns the creation time.
+     * @brief Returns the object Id.
      */
-    const system_clock::time_point& age() const override { return age_; }
+    uint32_t objectId() const override { return objectId_; }
     /**
      * @brief Returns true if this has less prioirty than otherConnection.
      */
     bool operator<(const IConnect& otherConnection) const override {
         return priority_ < otherConnection.priority() ||
-               (priority_ == otherConnection.priority() && age_ >= otherConnection.age());
+               (priority_ == otherConnection.priority() && objectId_ >= otherConnection.objectId());
     }
     /**
      * @brief Forcefully shuts down the connection.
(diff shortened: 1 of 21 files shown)


sdks/cpp/common/include/rpc/Connect.h
```diff
@@ -169,38 +193,62 @@ class Connect : public IConnect {
...
-                return;
-            }
-            // Returning if authorization is enabled and the client does not have monitor scope.
-            if (authz_ != &catena::common::Authorizer::kAuthzDisabled
-                && !authz_->hasAuthz(Scopes().getForwardMap().at(Scopes_e::kMonitor))) {
-                return;
+            } else if (authz_ == &catena::common::Authorizer::kAuthzDisabled
+                || authz_->hasAuthz(Scopes().getForwardMap().at(Scopes_e::kMonitor))) {
+                // Updating res_'s device_component and pushing update.
+                res_.Clear();
```
ctwarog-ross, line 199: why add this line in?

sdks/cpp/common/include/rpc/Connect.h
```diff
@@ -169,38 +193,62 @@ class Connect : public IConnect {
...
+     * priority_ = scope * 2 + write + (adm:w && forceConnection)
+     * 
      * @param jwsToken The jwsToken to use for authorization.
      * @param authz true if authorization is enabled, false otherwise.
      */
     void initAuthz_(const std::string& jwsToken, bool authz = false) override {
         if (authz) {
             sharedAuthz_ = std::make_shared<catena::common::Authorizer>(jwsToken);
             authz_ = sharedAuthz_.get();
+            // Setting up their connection priority.
```
ctwarog-ross, line 224: why are we doing this inside initAuthz?

sdks/cpp/connections/gRPC/src/ServiceImpl.cpp
```diff
@@ -97,6 +98,39 @@ void CatenaServiceImpl::processEvents() {
     }
 }
 
+bool CatenaServiceImpl::registerConnection(catena::common::IConnect* cd) {
```
ctwarog-ross, line 101: so the only time this returns false is if the new connection has lower priority then every current connection?

sdks/cpp/connections/gRPC/src/ServiceImpl.cpp
```diff
@@ -97,6 +98,39 @@ void CatenaServiceImpl::processEvents() {
...
+    }
+    // Adding the connection if possible.
+    if (canAdd) {
+        connectionQueue_.insert(it, cd);
+    }
+    return canAdd;
+}
+
+void CatenaServiceImpl::deregisterConnection(catena::common::IConnect* cd) {
+    std::lock_guard<std::mutex> lock(connectionMutex_);
```
ctwarog-ross, line 125: are we sure this is all we need to do to deregister a connection

sdks/cpp/common/include/rpc/Connect.h
BenWhitten-RV: bug fix

ctwarog-ross (approved):
Good work

BenWhitten-RV:
Waiting to give a demo at the meeting tmr before I merge.

PR #856, opened 2025-07-21
Added POST AssetRequest

ctwarog-ross:
Issue #837

now supports POST AssetRequests
example BL shifted from GET to POST section
TODO: Implement system that lets BL check authz and params to allow/deny post


sdks/cpp/connections/REST/src/controllers/AssetRequest.cpp
```diff
@@ -73,94 +110,170 @@ void AssetRequest::proceed() {
...
 
-        if (file_data.empty()) {
-            std::string error = "AssetRequest[" + std::to_string(objectId_) + "] file is empty: " + context_.fqoid() + "\n";
-            throw catena::exception_with_status(error, catena::StatusCode::INVALID_ARGUMENT);
+        // GET/asset
+        else if (context_.method() == Method_GET) {
+            // Locking device and parsing object data.
+            DEBUG_LOG << "sending asset: " << context_.fqoid();
+            std::string path = context_.EOPath();
+            path.append(context_.fqoid());
```
BenWhitten-RV, line 131: No idea if this is relevant but just a warning append will not split fqoid, so if you have a fqoid containing multiple "/" the path object will resemble smt like this: ["some", "EO", "path", "/some/fqoid"]

sdks/cpp/connections/REST/src/controllers/AssetRequest.cpp
```diff
@@ -73,94 +110,170 @@ void AssetRequest::proceed() {
...
+
+            //check for any read access
+            //TODO: move to BL
+            if (authz != &catena::common::Authorizer::kAuthzDisabled
```
BenWhitten-RV, line 135: Not a big thing but all of the hasAuthz functions check for kAuthzDisabled. 

Otherwise, Scopes_e::kOperate == "st2138:op", so I think you're scope check here would then be seeing if the client has the scope "st2138:opst2138" which is not real.

This is incorrect in a few other places throughout the codebase but if looking for a specific read scope you need to check for the scope either with or without the ":w" suffix since write implies the client can also read. We should probably add an additional overload to readAuthz and writeAuthz and use that instead of hasAuthz.

BenWhitten-RV (changes requested):

BenWhitten-RV (approved):

PR #863, opened 2025-07-22
Authorizer updates and cleanup

BenWhitten-RV:
**Changelog:**
- Switched Authorizer::Scopes from vector to set to allow for use of .contains() function.
- Cleaned up Authorizer functions.
- Added readAuthz(scope_e) and writeAuthz(scope_e) for convenience.
    - Kept the string impl too in case clients have custom scopes to implement in business logic 
- Made hasAuthz(scope) private so use of readAuthz() and writeAuthz() is forced (hard to mess up).
- Updated Authorizer unit test accordingly. Also made it more memory safe (I was NOT deallocating memory before my bad)

### DIFF
--- a/sdks/cpp/connections/REST/src/controllers/AssetRequest.cpp
+++ b/sdks/cpp/connections/REST/src/controllers/AssetRequest.cpp
@@ -157,7 +157,7 @@ void AssetRequest::proceed() {
 
             //check for any read access
             //TODO: move to BL
-            if (!(authz->readAuthz(catena::common::Scopes().getForwardMap().at(catena::common::Scopes_e::kOperate)))) {   
+            if (!(authz->readAuthz(catena::common::Scopes_e::kOperate))) {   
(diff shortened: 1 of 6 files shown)


sdks/cpp/connections/REST/src/controllers/AssetRequest.cpp
```diff
@@ -157,7 +157,7 @@ void AssetRequest::proceed() {
 
             //check for any read access
             //TODO: move to BL
-            if (!(authz->readAuthz(catena::common::Scopes().getForwardMap().at(catena::common::Scopes_e::kOperate)))) {   
+            if (!(authz->readAuthz(catena::common::Scopes_e::kOperate))) {
```
ZuhayrSarkerRV, line 160: I like this change and never understood why I cargo culted the forward map nonsense.

abrown-RV (approved):

PR #865, opened 2025-07-22
THE Connect Update Part 3 (gRPC Unit Tests)

BenWhitten-RV:
### PART 1: The Fellowship of gRPC

**Changelog:**
- Added 4 new functions to the virtual common::Connect class.
    - priority() returns the connection's priority value, calculated as = scope * 2 + write + (adm:w AND forceConnection)
    - objectId() returns the object's id. This will be lower based on creation time.
    - The < operator returns true based on the object's priority. If priorities are equal, then it instead returns true based on which object has the lower objectId (and is thus older).
    - shutdown() forcefully shuts down the connection.
- Added maxConnections input variable to the gRPC::ServiceImpl class.
- Added 2 new functions to the gRPC::ServiceImpl class.
    - RegisterConnection() registers a connection to the connectionQueue. If maxConnections has been reached, it forcefully shuts down the lowest priority connection in the queue.
        - ConnectionQueue is actually just a vector with the connections sorted by priority... Was going to use an actual priority queue but that doesn't allow for direct access to objects, which is required when deregistering connections.
    - DeregisterConnection() deregister's a connection from the ConnectionQueue.
- Refactored gRPC::Connect and updated to now use register and deregister connection.
    - Refactor changes are based on various things learned from developing tests for gRPC::Connect
- Added maxConnections flag.
- Created MockConnect object
- Updated relevant unit tests.

### PART 2: The Two APIs

**Changelog:**
- Added maxConnections input variable to the REST::ServiceImpl class.
- Added 2 new functions to the REST::ServiceImpl class along with some parity getters.
    - RegisterConnection() registers a connection to the connectionQueue. If maxConnections has been reached, it forcefully shuts down the lowest priority connection in the queue.
        - ConnectionQueue is actually just a vector with the connections sorted by priority... Was going to use an actual priority queue but that doesn't allow for direct access to objects, which is required when deregistering connections.
    - DeregisterConnection() deregister's a connection from the ConnectionQueue.
- Updated SocketReader to take in and save a ICatenaServiceImpl ptr at construction (so we can use the above functions).
- Updated REST::Connect to now use register and deregister connection.
    - Refactor changes are based on various things learned from developing tests for gRPC::Connect
- Created REST::MockServiceImpl object
- Updated relevant unit tests.

### PART 3: The Return of the Test

**Changelog:**
- Connect gRPC Unit Tests providing 100% coverage.


sdks/cpp/connections/REST/include/interface/IServiceImpl.h
```diff
@@ -29,7 +29,7 @@
 */
 
 /**
- * @file IServiceImpl.h
+ * @file ICatenaServiceImpl.h
```
abrown-RV, line 32: Do we want to actually rename file?

BenWhitten-RV: My bad, ctrl + h mistake. I had updated the class name to make it consistent and accidentally changed this too. Fixed now

sdks/cpp/connections/gRPC/src/ServiceImpl.cpp
```diff
@@ -97,6 +98,40 @@ void CatenaServiceImpl::processEvents() {
     }
 }
```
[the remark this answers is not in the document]
BenWhitten-RV: For the most part the logic, like the main flow, is so different between the two APIs that they can't possibly be duplicated. I think the only case where where the code actually is 1-1 duplicated is the new registerConnection and deregisterConnection functions.

I had considered making a shared parent class to inherit this functionality, but decided against it because I don't think the tradeoff of sharing these two functions is worth the added complexity. Especially for the gRPC service impl which requires an additional function to get the gRPC::CompleteQueue, as this would make the inheritance diagram something like:

Common::IService -> Common::Service -> gRPC::IService -> gRPC::Service

It would also make mocking the service impl for the gRPC unit tests a pain.

sdks/cpp/connections/REST/include/interface/IServiceImpl.h
BenWhitten-RV: ServiceImpl not matching the class within is kind of just an artifact from how it was implemented before I joined. I don't see why we shouldn't update it. 

We could drop the "Catena" part of the class name. It used to just be on its own hence the need to distinguish. A few months ago I moved it into the catena::gRPC/catena::REST namespace though, so it being the Catena gRPC/REST ServiceImpl is implied.

Aside from serviceImpl what other files don't have the same name as the class within?

sdks/cpp/connections/gRPC/src/ServiceImpl.cpp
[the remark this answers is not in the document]
abrown-RV: You could make a ConnectionQueue class that each service has as a member. Test it in isolation and then mock its behaviour in the service tests, would save you from manually filling it up in the tests.

abrown-RV (approved):

PR #871, opened 2025-07-24
ServiceImpl Config Class

BenWhitten-RV:
Implemented a config class for each ServiceImpl constructor as per the suggestion in the linked issue. I'm personally a big fan since I've been updating the constructors a fair bit recently and its a big pain to have to update every instance whenever something new is added. I'm curious what everyone else thinks though.

### DIFF
--- a/sdks/cpp/connections/REST/include/ServiceImpl.h
+++ b/sdks/cpp/connections/REST/include/ServiceImpl.h
@@ -68,6 +68,8 @@ using catena::REST::SocketReader;
 using catena::REST::SocketWriter;
 using catena::REST::SSEWriter;
 
+using namespace catena::common;
+
 namespace catena {
 /**
  * @brief Namespace for classes relating to handling REST API requests.
@@ -75,15 +77,36 @@ namespace catena {
 namespace REST {
 
 /**
+ * @brief Config object for the CatenaServiceImpl to streamline creation.
+ */
+class ServiceConfig {
+  public:
+    /**
+     * @brief A map of slots to ptrs to their corresponding device.
+     */
+    std::vector<IDevice*> dms = {};
+    /**
+     * @brief The path to the external object.
+     */
+    std::string EOPath = "";
+    /**
+     * @brief Flag to enable authorization.
+     */
+    bool authz = false;
+    /**
+     * @brief The port to listen on.
+     */
+    uint16_t port = 443;
+    /**
+     * @brief The maximum number of connections allowed to the service.
+     */
+    uint32_t maxConnections = 16;
+};
+
+/**
  * @brief Implements Catena REST API request handlers.
  */
 class CatenaServiceImpl : public catena::REST::IServiceImpl {
-
-  // Specifying which Device and IParam to use (defaults to catena::...)
-  using IDevice = catena::common::IDevice;
-  using IParam = catena::common::IParam;
-  using SlotMap = catena::common::SlotMap;
-
   public:
     /**
      * @brief Constructor for the REST API.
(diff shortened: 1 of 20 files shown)


ctwarog-ross:
Alot of the instances of ServiceConfig are assigning the same values to members. What if we implemented a constructor in ServiceConfig with default values, would save some repetition

abrown-RV:
> Alot of the instances of ServiceConfig are assigning the same values to members. What if we implemented a constructor in ServiceConfig with default values, would save some repetition

In python they let you do something called keyword arguments where you can call any function (or ctor) with named args instead of indexed args. So you can have your defaults set in the ctor, and then override the ones you want to change.

```python
config = ServiceConfig(
    dms, # indexed since its always required
    max_connection=10,
    authz=True,
)
```
or C# lets you do
```csharp
ServiceConfig config = new ServiceConfig(dms) {
    MaxConnections = 10,
    Authz = true,
};
```

It doesn't seem like cpp supports this yet. So what I've seen is the function chaining pattern. Which is you have a bunch of setter functions for each config option that can be overridden that return a reference to itself so that you can chain the functions together.

Something like
```cpp
ServiceConfig config = ServiceConfig(dms)
    .maxConnections(10)
    .authz(true);
```

https://levelup.gitconnected.com/how-to-implement-method-chaining-in-c-3ec9f255972a

abrown-RV:
Sorry looking more at the PR Christian's point is about the repetitive code. My response to that would be that all the repetition is to pull the flag values in, so to take out that repetition we'd need a method that initializes a config from the flags.

My point about the function chaining is more quality of life improvement. Maybe not so necessary. I just think it looks clean when you don't have the repeated `config.` at the start of every line.

abrown-RV:
> C++ does have ctor, Im thinking that's how we should have this. A constructor with args and default values

Well that's no different than using the ServiceImpl ctor. The whole idea here is that the ServiceImpl ctor is growing. Even if there are nice default values, you can't take advantage of them if you want to have any non default customization.

For example with the ctor as it is on develop
```cpp
CatenaServiceImpl(std::vector<IDevice*> dms, std::string& EOPath, bool authz = false, uint16_t port = 443);
```
if you want to set the port to anything but the default, you have to set all the other args, even if you just want their defaults. And then the issue I'm raising about that is then someone has hardcoded a value to a default that may change in the future, and they don't know it.

sdks/cpp/connections/REST/include/ServiceImpl.h
```diff
@@ -77,6 +77,94 @@ namespace catena {
...
+/**
+ * @brief Config object for the ServiceImpl to streamline creation.
+ */
+class ServiceConfig {
+  public:
+    /**
+     * @brief Sets EOPath, authz, port, and maxConnections via command line
+     * flags.
+     */
+    ServiceConfig& set_flags() {
```
abrown-RV, line 89: Forgive my unfamiliarity with the purpose of examples. My understanding so far is that a developer working for a company actually developing a device that implements catena would create their own code that looks like our examples, but does not *use* our examples. As such I think it would be best if we did not dictate that they do flags in the same way as we do, and do not force `absl` as a dependency on them. So while I think the `set_flags` is a great idea, it would be best if specifics of our examples are separated from the sdk code. Could you do some kind of shared helper function that either builds a config from the flags and returns it, or takes an existing config and imposes the flags upon it. Looking quickly at the code base I'm not sure a good place for such a thing exists, you would know that one better.

BenWhitten-RV:
Removed set_flags() and just changed it back in the examples to use the setters. Clients can choose which flags to include/exclude in the business logic.

abrown-RV (approved):

Issue #777, opened 2025-06-24
Implement gRPC ParamInfoRequest test


ZuhayrSarkerRV:
I think this unit test is safe to do without being impacted by the subscription traversal bugs (#769, #767).

PR #881, opened 2025-07-29
Ut #777 zs

ZuhayrSarkerRV:
# 📃 Intro
- This was a bit of a weird test because I had to change parts of the base class to handle errors better and more closely match the REST implementation - it looks like I did a lot of changing, but I mostly moved things around (a lot of the moving is literally just centered around me switching mode 2 and 3 to match the orientation in REST)
- File has 100% coverage and was mostly a simple match from REST implementation, but when it came to error handling I was fighting to figure out how to reach certain blocks of code, only to realize the error handling needed to be improved
- I also made a few changes to the helpers

# ✨ Main Additions/Changes

### Unit Test Legend
| |  |
| ------------- | ------------- |
| ✅ | Expects eq/true |
| ☑️ | Expects false/null/no-eq/etc., but for a success case |
| ⭕ | Expects eq/true, but covers an error/failure case |
| 🔴 | Expects false |

For ✅/☑️ expect a default status of Catena Status OK 

Note: I am going to abbreviate ParamInfoRequest as PIR

### Fixtures/Helpers
- A lot of this was just shamelessly copied from Ben's setup, however, I refactored StreamReader to be a template inside GRPCTestHelpers.h
- makeOne - creates PIR handler object
- initPayload - initializes PIR payload
- testRPC - makes an async RPC call, waits for a response, and then compares output
- Members for input values, output values and expected values

Literally everything in this uses testRPC() which tests EQ/TRUE, so I won't bother specifying for every test.

### Section 0 Tests: Basic Setup
- ✅ **0.0**: TRUE create a PIR object
- ✅ **0.1**: Authorization test with valid token
- ⭕ **0.2**: Authorization test with invalid token
  - Expects Catena `Invalid JWS Token (UNAUTHENTICATED)`
- ⭕ **0.3**: Authorization test with invalid slot
  - Expects Catena `Device not found in slot (NOT_FOUND)`

### Mode 1 Tests: Get all top-level params without recursion
- ✅ **1.1**: Get all top-level params without recursion
- ✅ **1.2**: Get top-level params with array type
- ⭕ **1.3**: Get top-level params with empty list returned from getTopLevelParams 
  - Expects Catena `No top-level parameters found (NOT_FOUND)`
- ⭕ **1.4**: Get Top-level params with error status in returned params
  - Expects Catena `Error processing parameter (INTERNAL)`
- ⭕ **1.5**: Get Top-level param with exception during param processing
  - Expects Catena `Error getting top-level parameters (INTERNAL)`

### Mode 2 Tests:  Get all top-level params WITH recursion
- ✅ **2.1**: Get top-level params with recursion and deep nesting
- ✅ **2.2**: Get top-level params with recursion and arrays
- ⭕ **2.3**: Get top-level params with recursion and error in child processing
  - Expects Catena `Error processing child parameter (INTERNAL)`
- ⭕ **2.4**: Get top-level params with empty list and recursion
  - Expects Catena `No top-level parameters found (INTERNAL)`

### Mode 3 Tests: Get a specific parameter and its children if recursive
- ✅ **3.1**: Get specific parameter without recursion
- ✅ **3.2**: Get specific parameter WITH recursion
- ⭕ **3.3**: Parameter not found
  - Expects Catena `Parameter not found: missing_param (NOT_FOUND)`
- ⭕ **3.4**: Catena exception in getParam
  - Expects Catena `Error processing parameter (INTERNAL)`

### Section 4 Tests: Additional error cases at end of proceed()
- ⭕ **4.1**: Catena exception in proceed()
  - Expects Catena `Test catena exception (INTERNAL)
- ⭕ **4.2**: Std exception in proceed()
  - After throwing an std exception, expects Catena `Failed due to unknown error in ParamInfoRequest: Test std exception (UNKNOWN)`
- ⭕ **4.3**: Unknown exception in proceed()
  - After throwing an int, expects Catena `Failed due to unknown error in ParamInfoRequest (UNKNOWN)`

## 🟡 ParamInfoRequest.cpp
- Removed every single throw in the kProcess step and instead switched to assigning return code variable, which is more testable and matches other implementations
  - Note the rc scope brackets
- Removed redundant error testing in kWrite stage as any errors would be caught beforehand during process 
  - For example, if we had null/no responses, this would be caught beforehand
- Added more detailed error handling per-mode and removed vague "no more responses" error code
  - Old implementation was incredibly unclear on the specific error and would often resort to this old basic error code without detailing what the exact error was.
- Add default case to gcov exclude, impossible to reach
- Switched modes 2 and 3 to match orientation in REST
  - Does not change any kind of functionality, just looks cleaner (one of the modes was added afterwards hence the orientation weirdness)

# 🛠️ Other changes
- Renamed setupMockParam to setupMockParamInfo to avoid confusion
- Refactored StreamReader helper as a single template in GRPCTestHelpers to be used in all the relevant gRPC streaming unit tests
- Got rid of redundant test on REST-side, checked for 100% coverage

### DIFF
--- a/sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
+++ b/sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
@@ -80,23 +80,19 @@ void ParamInfoRequest::proceed(bool ok) {
                 if (!dm) {
                     throw catena::exception_with_status("device not found in slot " + std::to_string(req_.slot()), catena::StatusCode::NOT_FOUND);
                 }
-
                 if (service_->authorizationEnabled()) {
                     sharedAuthz = std::make_shared<Authorizer>(jwsToken_());
                     authz = sharedAuthz.get();
                 } else {
                     authz = &Authorizer::kAuthzDisabled;
                 }
-
                 // Mode 1: Get all top-level parameters
                 if (req_.oid_prefix().empty() && !req_.recursive()) {
                     std::vector<std::unique_ptr<IParam>> top_level_params;
-
                     {
                     std::lock_guard lg(dm->mutex());
                     top_level_params = dm->getTopLevelParams(rc, *authz);
                     }
-
                     if (rc.status == catena::StatusCode::OK && !top_level_params.empty()) {
                         std::lock_guard lg(dm->mutex());                    
                         responses_.clear();  
@@ -111,8 +107,7 @@ void ParamInfoRequest::proceed(bool ok) {
                                     updateArrayLengths_(top_level_param->getOid(), array_length);
                                 }
                             }   
-                        }
-                        
+                        }                        
                         // Begin writing responses back to the client
                         writer_lock_.lock();
                         status_ = CallStatus::kWrite;
@@ -125,55 +120,13 @@ void ParamInfoRequest::proceed(bool ok) {
                         throw catena::exception_with_status("No top-level parameters found", catena::StatusCode::NOT_FOUND);
                     }
 
-                // Mode 2: Get a specific parameter and its children
-                } else if (!req_.oid_prefix().empty()) { 
-                    {
-                    std::lock_guard lg(dm->mutex());
-                    param = dm->getParam(req_.oid_prefix(), rc, *authz);
-                    }
-
-                    if (rc.status == catena::StatusCode::OK && param) {
-                        responses_.clear();
-                        
-                        // Add the main parameter first 
-                        responses_.emplace_back();
-                        param->toProto(responses_.back(), *authz);
-                        
-                        // If the parameter is an array, update the array length
-                        if (param->isArrayType()) {
-                            uint32_t array_length = param->size();
-                            if (array_length > 0) {
-                                updateArrayLengths_(param->getOid(), array_length);
-                            }
-                        }
-                        
-                        // If recursive is true, collect all parameter info recursively through visitor pattern
-                        if (req_.recursive()) {
-                            ParamInfoVisitor visitor(*dm, *authz, responses_, *this);
-                            ParamVisitor::traverseParams(param.get(), req_.oid_prefix(), *dm, visitor, *authz);
-                        }
-
-                        // Begin writing responses back to the client
-                        writer_lock_.lock();
-                        status_ = CallStatus::kWrite;
-                        writer_.Write(responses_[0], this); //Write the first response
-                        writer_lock_.unlock();
-                        break;
-                    } else if (rc.status != catena::StatusCode::OK) {
-                        throw catena::exception_with_status(rc.what(), rc.status);
-                    } else {
-                        throw catena::exception_with_status("Parameter not found: " + req_.oid_prefix(), catena::StatusCode::NOT_FOUND);
-                    }
-
-                // Mode 3: Get ALL parameters recursively
+                // Mode 2: Get ALL parameters recursively
                 } else if (req_.oid_prefix().empty() && req_.recursive()) {
                     std::vector<std::unique_ptr<IParam>> top_level_params;
-
                     {
                         std::lock_guard lg(dm->mutex());
                         top_level_params = dm->getTopLevelParams(rc, *authz);
                     }
-
                     if (rc.status == catena::StatusCode::OK && !top_level_params.empty()) {
                         std::lock_guard lg(dm->mutex());                     
                         responses_.clear();  
@@ -188,12 +141,10 @@ void ParamInfoRequest::proceed(bool ok) {
                                     updateArrayLengths_(top_level_param->getOid(), array_length);
                                 }
                             }
-                            
                             // Collect all parameter info recursively through visitor pattern
                             ParamInfoVisitor visitor(*dm, *authz, responses_, *this);
                             ParamVisitor::traverseParams(top_level_param.get(), "/" + top_level_param->getOid(), *dm, visitor, *authz);
-                        }
-                        
+                        }                        
                         // Begin writing responses back to the client
                         writer_lock_.lock();
                         status_ = CallStatus::kWrite;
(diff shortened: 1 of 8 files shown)


unittests/cpp/gRPC/GRPCTestHelpers.h
```diff
@@ -0,0 +1,217 @@
...
+ */
+
+/**
+ * @brief Generic template for ClientReadReactor that can be used across different test files
+ * @tparam ResponseType The type of response being streamed
+ * @tparam PayloadType The type of request payload
+ * @tparam MakeCallFunc Function type for making the async call
+ */
+template<typename ResponseType, typename PayloadType, typename MakeCallFunc>
+class StreamReader : public grpc::ClientReadReactor<ResponseType> {
```
BenWhitten-RV, line 66: I think this class is better suited to either its own file or the GRPCTest.h file where we also have the code for the gRPC server.

sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
```diff
@@ -62,202 +62,182 @@ void ParamInfoRequest::proceed(bool ok) {
...
-                        // Add the main parameter first 
-                        responses_.emplace_back();
-                        param->toProto(responses_.back(), *authz);
-                        
-                        // If the parameter is an array, update the array length
-                        if (param->isArrayType()) {
-                            uint32_t array_length = param->size();
-                            if (array_length > 0) {
-                                updateArrayLengths_(param->getOid(), array_length);
+                        if (rc.status == catena::StatusCode::OK) {
```
BenWhitten-RV, line 98: There's this weird section inbetween where the dm lock unlocks before locking again. I would just use one lock_guard for this whole section.

sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
```diff
@@ -62,202 +62,182 @@ void ParamInfoRequest::proceed(bool ok) {
...
+                                    // For array types, calculate and update array length
+                                    if (top_level_param->isArrayType()) {
+                                        uint32_t array_length = top_level_param->size();
+                                        if (array_length > 0) {
+                                            updateArrayLengths_(top_level_param->getOid(), array_length);
+                                        }
+                                    }   
+                                }                        
+                                // Begin writing responses back to the client
+                                writer_lock_.lock();
```
BenWhitten-RV, line 118: use of lock in ParamInfoRequest is unnecessary since it's a linear process. It's only needed in Connect since we have a condition variable. Even if it was necessary, you should use lockGuard() since it's safer.

sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
```diff
@@ -62,202 +62,182 @@ void ParamInfoRequest::proceed(bool ok) {
...
+                                status_ = CallStatus::kWrite;
+                                writer_.Write(responses_[0], this);  //Write the first response
```
BenWhitten-RV, line 120: The old version also just fell through into write if status was ok, which I think it better than having writes interspersed within the kprocess step.

This is only needed in Connect where a signal could possible be received before the client has a change in kWrite to send the connected slots.

sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
```diff
@@ -62,202 +62,182 @@ void ParamInfoRequest::proceed(bool ok) {
...
+                                break;  
                             }
                         }
-                        
-                        // If recursive is true, collect all parameter info recursively through visitor pattern
-                        if (req_.recursive()) {
-                            ParamInfoVisitor visitor(*dm, *authz, responses_, *this);
-                            ParamVisitor::traverseParams(param.get(), req_.oid_prefix(), *dm, visitor, *authz);
+                    // Mode 2: Get ALL parameters recursively
+                    } else if (req_.oid_prefix().empty() && req_.recursive()) {
```
BenWhitten-RV, line 126: Same notes as above

sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
```diff
@@ -62,202 +62,182 @@ void ParamInfoRequest::proceed(bool ok) {
...
+                                // Begin writing responses back to the client
+                                writer_lock_.lock();
+                                status_ = CallStatus::kWrite;
+                                writer_.Write(responses_[0], this);  //Write the first response
+                                writer_lock_.unlock();
+                                break;  
+                            }
+                        }
+                    // Mode 3: Get a specific parameter and its children
+                    } else if (!req_.oid_prefix().empty()) {
```
BenWhitten-RV, line 163: Same notes as above

sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
```diff
@@ -62,202 +62,182 @@ void ParamInfoRequest::proceed(bool ok) {
...
-                    }
-                    writer_lock_.unlock();
-                    break;
-                }        
-            } catch (const std::exception& e) {
-                status_ = CallStatus::kFinish;
-                writer_.Finish(grpc::Status(grpc::StatusCode::INTERNAL, 
-                    "Error writing response: " + std::string(e.what())), this);
-            }
+                writer_lock_.lock(); //Lock the writer for writing
```
BenWhitten-RV, line 218: response scope? Also lock is not needed.

BenWhitten-RV: Just thought maybe this is a case similar to connect where its writing stuff too quickly for the client, in which case use lockguard, or if its just a unittest thing use an await after every read.

sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
```diff
@@ -62,202 +62,182 @@ void ParamInfoRequest::proceed(bool ok) {
...
+                if (current_response_ >= responses_.size()-1) {
```
BenWhitten-RV, line 219: Instead of this, it's better to have it check after the last write if there's anything remaining. If there is, transition to kPostWrite where you call writer_.Finish(), and then transition to kFinish where you deregister.

See DeviceRequest or ExecuteCommand.

sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
```diff
@@ -62,202 +62,182 @@ void ParamInfoRequest::proceed(bool ok) {
...
-                break;
+                rc = catena::exception_with_status(err.what(), err.status);
+            } catch (const std::exception& e) {
+                rc = catena::exception_with_status("Failed due to unknown error in ParamInfoRequest: " + std::string(e.what()), catena::StatusCode::UNKNOWN);
             } catch (...) {
+                rc = catena::exception_with_status("Failed due to unknown error in ParamInfoRequest", catena::StatusCode::UNKNOWN);
+            }
+            
+            // If there was an error, finish the process and write the error to the client
+            if (rc.status != catena::StatusCode::OK) {
```
BenWhitten-RV, line 208: Have it fall through if status is OK.

BenWhitten-RV (changes requested):

sdks/cpp/connections/gRPC/src/controllers/ParamInfoRequest.cpp
BenWhitten-RV: Await after each read doesn't work since the response speed isn't dictated by signals... scrap that.

Use lockguard tho

unittests/cpp/gRPC/GRPCTestHelpers.h
ZuhayrSarkerRV: I put it in a different file, check latest commit.

Issue #614, opened 2025-05-12
REST API Tests - ExternalObjectRequest


BenWhitten-RV:

BenWhitten-RV:
This is done and can be closed.

PR #883, opened 2025-07-29
Asset Request REST UT

ctwarog-ross:
Issue #614

AssetRequest REST UT 90% coverage
Exposed UNIT_TESTING cmake arg as macro to mask private keyword behind macro value for testing
Added to_base64 and from_base64 to convert to and from byte arrays


sdks/cpp/connections/REST/src/controllers/AssetRequest.cpp
```diff
@@ -157,7 +165,11 @@ void AssetRequest::proceed() {
 
             //check for any read access
             //TODO: move to BL
-            if (!(authz->readAuthz(catena::common::Scopes_e::kOperate))) {   
+            bool check = authz->readAuthz(catena::common::Scopes_e::kMonitor);
```
BenWhitten-RV, line 168: I assume you forgot to remove this from testing?

sdks/cpp/connections/REST/src/controllers/AssetRequest.cpp
```diff
@@ -177,18 +189,13 @@ void AssetRequest::proceed() {
...
             std::vector<uint8_t> file_data((std::istreambuf_iterator<char>(file)), std::istreambuf_iterator<char>());
 
-            if (file_data.empty()) {
-                std::string error = "AssetRequest[" + std::to_string(objectId_) + "] file is empty: " + context_.fqoid() + "\n";
-                throw catena::exception_with_status(error, catena::StatusCode::INVALID_ARGUMENT);
-            }
-
             // Set the payload encoding and compress the data accordingly
-            if (context_.fields("compression") == "GZIP") {
+            if (context_.hasField("compression") && context_.fields("compression") == "GZIP") {
```
BenWhitten-RV, line 193: It should be safe to remove the context_.hasField()

This is because context_.fields("...") will return "" if the field is not there.

abrown-RV (dismissed):

sdks/cpp/connections/REST/src/controllers/AssetRequest.cpp
ctwarog-ross: whoops, good catch

BenWhitten-RV (approved):

PR #889, opened 2025-07-30
Fixed code generation for Params without values

BenWhitten-RV:
**Changelog:**
- Params defined without a value, namely for use as templates, are now generated as ParamDescriptor objects within device.
- All top-level param descriptors are added to an additional vector in Device called descriptors_.
- DeviceRequest will now call toProto on all ParamDescriptors in descriptors_ instead of all the Params in params_. If the param exists in params_, it will instead call toProto on the Param object to preserve param's value.


BenWhitten-RV:
> Realized I need to update the ParamDescriptor unittests first give me a sec

fixed

BenWhitten-RV:
The template params are separate objects at the moment, optimization can be done sometime in the future.

Right now the only way to get these descriptors is via a call to DeviceRequest. Should we also allow calls like getParam and ParamInfoRequest to retrieve them?

If we did, easiest fix would probably just be to also initialize a pramWithValue with these descriptors and an EMPTY value.

PR #898, opened 2025-07-31
Common Example Updates

BenWhitten-RV:
Some of the common/examples were outdated/were just wrong. This PR fixes them.

It also improves readability lost in the logger update and updates all main() + StartLogging() functions to match.


sdks/cpp/common/examples/hello_world/hello_world.cpp
```diff
@@ -136,17 +136,21 @@ int main (int argc, char** argv) {
     }
     auto& phonetic_alphabetParam = *dynamic_cast<ParamWithValue<std::vector<std::string>>*>(ip.get());
     std::vector<std::string>& phonetic_alphabetValue = phonetic_alphabetParam.get();
-    DEBUG_LOG << "phonetic alphabet initial value: ";
+    std::ostringstream outStream;
+    outStream << "phonetic alphabet initial value: ";
     for (const auto& s : phonetic_alphabetValue) {
-        DEBUG_LOG << s << " ";
+        outStream << s << " ";
```
ZuhayrSarkerRV, line 142: This is a good change since that isn't exactly debug information, although that makes me a little concerned for where else console prints might have gotten wrongly replaced with logs.

sdks/cpp/common/examples/factory/factory.cpp
```diff
@@ -224,6 +222,7 @@ int main() {
     (*jackRussell)();
     rabid->operator()();  // alternative syntax
 
+    std::cout<<std::endl;
```
ctwarog-ross, line 225: we use DEBUG_LOG for this now, it prints to std as well as log files

sdks/cpp/common/examples/hello_world/hello_world.cpp
ctwarog-ross: logs also print to console on debug builds

ctwarog-ross (approved):

PR #903, opened 2025-08-01
Fixed logging folder creation

ctwarog-ross:
CMake now checks for the logging folder on build
if not present it is created
UNIT_TESTING macro exposed to entire codebase

### DIFF
--- a/sdks/cpp/CMakeLists.txt
+++ b/sdks/cpp/CMakeLists.txt
@@ -56,7 +56,6 @@ add_compile_options(--coverage)
 add_link_options(--coverage)
 
 option(UNIT_TESTING "Build the unit tests" ON)
-add_compile_definitions(UNIT_TESTING="${UNIT_TESTING}")
 
 # enable testing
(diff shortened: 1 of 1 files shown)


abrown-RV (approved):
Nice and clean

ZuhayrSarkerRV:
Why are we creating the logs in the root directory? And if so, shouldn't we add it to gitignore?

ctwarog-ross:
> Why are we creating the logs in the root directory? And if so, shouldn't we add it to gitignore?

it is handled by gitignore. I decided it would be best to have logs in the root dir, is there a better place for them?

ZuhayrSarkerRV:
> > Why are we creating the logs in the root directory? And if so, shouldn't we add it to gitignore?
> 
> it is handled by gitignore. I decided it would be best to have logs in the root dir, is there a better place for them?

No, it's fine. Probably easier to access from there. Just assumed it needed to be in build for some reason due to cmake stuff.

PR #904, opened 2025-08-01
Add a Subscription Limit + Small UT Fixes

ZuhayrSarkerRV:
### 🟢 (NEW) Add a subscription limit
- Created a new function `maxSubscriptions` which calculates the maximum number of possible subscriptions based on the parameters given in a device (i.e. if a device has a total of 30 parameters, the limit is 30)
  - The limit is calculated during subscription manager creation, so subscription manager now has an explicit constructor that requires the device and authorization on creation
  - image
- For regular params, it adds the single subscription, checks if it surpasses limit, and then blocks it from being added
- For wildcards, I added a function `countWildcards` which counts the number of subscriptions that could occur from expanding a wildcard
  - A reminder that wildcards themselves are NOT stored in the subscriptions (i.e. param/*) , wildcards are first processed and then stored as individual OIDs (i.e. param/subparam1, param/subparam2).
- Both of these functions use the existing ParamVisitor implementation to make it easier to traverse and count 
- I also improved the flow of SubscriptionManager a little bit removing some repetitive if statements and also accommodating for the new resource exhausted check
- Updated unit tests to cover new functionality for three new error cases

#### JUST TO CONFIRM THIS WORKS:
I've saved a response trying "/*" (all top-level params) and it's 58 responses here:

  image
--

And if you look at the limit through debugger:

image

---

### 🛠️ Other fixes/changes

- Removed pure virtual from visitArray in IParamVisitor as it is only used in ParamInfoRequest in order to calculate the length of arrays - this would remove the need to override it in other classes (where it isn't testable because it isn't used!)
- Added GCOV_EXCL_LINE to the closing brackets in Device to skip gcov bracket bug and show true coverage
- Manager now has to be initialized in service 
  - Currently it is assigned the first device, I will later update it to support multiple (so this constructor change was needed anyway)
  - image
- Fixed socket reader test to take mock subscription manager, not implementation (whcih also fixes a warning for no call to maxSubscriptions())
- Fixed no logging in ConnectionQueue test as well as Connect test 
- Updated Device test to correctly use shallow copy for top-level parameters (this was a mistake on my part) 
  - image
  - For reference, this is what it says in device to confirm this:
    - image
- Updated GRPCTest.h to expect maxSubscriptions during subscription manager creation
  - Also updated ServiceImplTest for the same expectation

### DIFF
--- a/sdks/cpp/common/include/SubscriptionManager.h
+++ b/sdks/cpp/common/include/SubscriptionManager.h
@@ -172,7 +172,7 @@ class SubscriptionManager : public ISubscriptionManager {
      * @param authz The authorizer for permission checking
      * @return The number of parameters that would be subscribed to
      */
-    uint32_t countWildcardSubscriptions(IParam* param, const std::string& path, IDevice& dm, Authorizer& authz);
+    uint32_t countWildcards(IParam* param, const std::string& path, IDevice& dm, Authorizer& authz);
 };
 
 } // namespace common
--- a/sdks/cpp/common/src/SubscriptionManager.cpp
+++ b/sdks/cpp/common/src/SubscriptionManager.cpp
@@ -78,7 +78,7 @@ bool SubscriptionManager::addSubscription(const std::string& oid, IDevice& dm, e
     // Add wildcard subscription.
     else if (wildcard && param) {
         // Count potential subscriptions before adding them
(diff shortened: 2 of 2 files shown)


ZuhayrSarkerRV:
On hold until I fix merge conflicts.

ZuhayrSarkerRV:
On hold because after discussing with Ben, we can't link a device to subscription manager, it gets convoluted - use getters to device using slot instead.

PR #905, opened 2025-08-01
Upgraded one_of_everything UI

johndanenRV:
One of everything UI now has embedded menus with multiple tabs. All constrained params now have a widget to display.

image


NR-RV (approved):
looks good

PR #932, opened 2025-08-06
Added JWS token expiration function to Authorizer

BenWhitten-RV:
**Changelog:**
- Authorizer now extracts exp from the token.
    - Default exp of 0 indicates that token does not have an exp.
- Added function Authorizer::isExpired() which returns true if the token is expired.
    - Ignores if exp is 0 (default value for unset).
- Connect now disconnects user if token is expired.
- ExecuteCommand will not cancel if token becomes expired during longer commands.
- Updated unit tests to reflect this.
    - AuthorizerTest was also updated to use a TestAuthorizer function which allows for manually getting and setting of scopes and exp.

Token refresh and such to be done at a later date.


abrown-RV (approved):
retracting request

PR #939, opened 2025-08-08
Add missing coverage for Device UT

ZuhayrSarkerRV:
Added missing coverage for the following Device.h functions:
- hasLanguage
- unhandled_exception
- slot
- mutex
- default max length
- default total length
- set default max length
- set default total length

Also added coverage for signals (called and then tested to see if they were blocked):
- GetValueSetByClient
- GetLanguageAddedPushUpdate
- GetValueSetByServer
- GetDownloadAssetRequest
- GetDeleteAssetRequest

I also added gcov exclude lines to the closing brackets on some of the functions in Device.cpp (this is a gtest artifact). Both Device.cpp and Device.h now have 100% coverage

### DIFF
--- a/sdks/cpp/common/src/Device.cpp
+++ b/sdks/cpp/common/src/Device.cpp
@@ -135,7 +135,7 @@ catena::exception_with_status Device::commitMultiSetValue (catena::MultiSetValue
         }
     }
     return ans;
-}
+} //GCOV_EXCL_LINE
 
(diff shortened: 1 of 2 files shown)


NR-RV (approved):
not a fan of `//GCOV_EXCL_LINE` being added as it can mean missed branches and funcitons but will allow as its the closing `}`

PR #958, opened 2025-08-12
Some UT and subscription logic tweaks from closed PR

ZuhayrSarkerRV:
These are just some tweaks/fixes from PR #904 which was abandoned due to the complexity of adding resource exhaustion to subscriptions without having first implemented the multi-device/client subscription logic first (ref: #832 )

I'm quite literally just going to copy-paste the relevant parts of that changelog here.

- Added const qualifier to IDevice in traverseParams
- Improved the flow of SubscriptionManager a bit, removing some repetitive if statements
- Removed pure virtual from visitArray in IParamVisitor as it is only used in ParamInfoRequest in order to calculate the length of arrays - this would remove the need to override it in other classes (where it isn't testable because it isn't used!)
- Fixed socket reader test to take mock subscription manager, not implementation
- Fixed no logging in ConnectionQueue test as well as gRPC Connect test 
- Updated Device test to correctly use shallow copy for top-level parameters (this was a mistake on my part) 
  - image
  - For reference, this is what it says in device to confirm this:
    - image

Before I delete the branch associated with #904 I will try to document some of it's logic somewhere since some of it can be re-used for the future multi-device/client subscription support.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -245,7 +245,7 @@
         //"api_only", // unimplemented
         "asset_request_REST",
         "status_update_REST",
-        "status_update_REST_JSON",
+        "status_update_JSON_REST",
         "use_commands_REST",
         "use_menus_REST",
         "structs_with_authz_REST",


sdks/cpp/common/src/SubscriptionManager.cpp
```diff
@@ -42,48 +42,40 @@ bool SubscriptionManager::addSubscription(const std::string& oid, IDevice& dm, e
     std::set<std::string>& dmSubscriptions = subscriptions_[dm.slot()];
 
     // Getting base Oid if wildcard.
-    if (wildcard) {
-        baseOid = oid.substr(0, oid.length() - 2);
-    } else {
-        baseOid = oid;
-    }
+    baseOid = wildcard ? oid.substr(0, oid.length() - 2) : oid;
```
NR-RV, line 45: in line if bad for readability even if cool

NR-RV (approved):

NR-RV (approved):

PR #974, opened 2025-08-19
Added menu groups to device request

### LINKED ISSUE #973: MenuGroups don't get sent with device request
MenuGroups aren't sent as part of a Device Request. This really only matters for getting the names of the MenuGroups, since the menus get streamed later anyways. I remember a while back deciding with John that menuGroups should be included in the first DeviceComponent sent in the Stream. I guess it just never got implemented.

johndanenRV:
closes #973


abrown-RV (changes requested):
Are there unit tests to be updated?

abrown-RV (approved):

PR #976, opened 2025-08-20
SMPTE done

### LINKED ISSUE #909: Migrate to the SMPTE hosted Interface Spec
There are a few breaking changes in here that can be found using: `https://github.com/smpte/st2138-a/compare/eee92f8...main` which compares the first commit of each file to it's current version on main. The first commit was mainly a faithful copy of our files.

## Proto changes

- device.proto ComponentCommand - uses command as the name of the Param object, instead of Param, this prevents the schemas for Component Param and Component Command being identical.
- menu.proto MenuGroup and Menu - added an integer to show their ordering
- param.proto Import - removed the ability to select a digest algorithm and forced everything to use SHA256
- added access management RPCs

## openapi changes

I bundle the openapi spec into a single file which fixed all the resolver errors we were seeing: https://petstore.swagger.io/?url=https://smpte.github.io/st2138-a/docs/openapi.yaml#/

I also added get /access, and post/access endpoints to match the access management RPCs

## tooling changes
There's a script, `build-openapi.sh` to do the bundling, and another `install-tooling.sh` to install the tooling used by the build script.

The changes required by this issue amount to:

- deleting our copies of `interface`, `schema`, `examples` at the repo root. Let's nix nab2024 while we're at it too.
- mounting the st2138-a repo as a submodule
- changing the cmake files to pull the .proto files from there instead of their current location
- a mix of prayers and curses until it all builds and runs again ;-)

The release tag is: https://github.com/SMPTE/st2138-a/releases/tag/v1.0.0-pre-fcd

ctwarog-ross:
Closes issue #909

Removed interface, schema and example_device_models, nab2024
Converted all relevant fields from catena:: to st2138::
Added schema repo as subodule
updated codegen and examples to validate against new schema
index.js reworked to use validator script from smpte repo
tools/package.json changed to include smpte validator as package

smpte/build-openapi.sh runs on toolchain creation
smpte/openapi.sh runs post creation on relevant containers
cmake also runs post creation of container, ninja is now ran by user
Correcting github workflow to recursively pull smpte submodule when necessary
Changed to curl from wqget in healthcheck

NOTE: Toolchain will only build once merged. It was tested here and built successfully https://github.com/rossvideo/Catena/actions/runs/17270446143/job/49013442802


abrown-RV (changes requested):
Earmarking for Monday: changes are needed in the examples dockerfile as well. The proto path needs to change to the new smpte submodule path.

.devcontainer/toolchain-cpp.dockerfile
```diff
@@ -74,4 +75,10 @@ RUN . /root/toolchain.env \
 RUN sudo apt-get update && sudo apt-get install -y python3-pip \
     && sudo pip3 install --no-cache-dir gcovr --break-system-packages \
     && sudo apt-get clean \
-    && sudo rm -rf /var/cache/apt/archives/ /var/lib/apt/lists/*
\ No newline at end of file
+    && sudo rm -rf /var/cache/apt/archives/ /var/lib/apt/lists/*
+
+# Build OpenAPI
+RUN echo "INSTALLING OPENAPI"
+COPY smpte /root/smpte
```
abrown-RV:
```suggestion
COPY smpte /root/smpte/install-tooling.sh
```
So that only the script is left in the image

package.json
```diff
@@ -1,11 +1,12 @@
 {
     "dependencies": {
-        "ajv": "^8.8.2",
+        "ajv": "^8.17.1",
         "ajv-formats": "^2.1.1",
         "catena-codegen": "file:tools/codegen",
         "commander": "^9.2.0",
         "json-source-map": "^0.6.1",
         "node": "^20.7.0",
+        "nvm": "^0.0.4",
```
abrown-RV:
```suggestion
```
Not needed? nvm is only used by the `install-tooling.sh` which is handled in the toolchain now

.devcontainer/devcontainer.json
```diff
@@ -1,27 +1,36 @@
...
+      "IMAGE_VERSION": "1.0.0",
+      "IMAGE_VENDOR": "Ross Video",
+      "CMAKE_BUILD_TYPE": "Debug",
+      "CONNECTIONS": "gRPC;REST",
+      "BUILD_TARGET": "build/cpp"
+    }
+  },
+  "postCreateCommand": {
+    "dockerInit": "sudo groupdel docker || true && DOCKER_GID=$(stat -c '%g' /var/run/docker.sock) && sudo groupadd -forg $DOCKER_GID docker && sudo usermod -aG docker ${USER_NAME}",
+    "buildOpenAPI": "cd /home/${USER_NAME}/Catena/smpte && ./build-openapi.sh",
```
abrown-RV, line 21:
```suggestion
```
Might not actually need this? the submodule actually already commits the product of the script. So its redundant, I guess it doesn't really harm anything

.devcontainer/toolchain-cpp.dockerfile
```diff
@@ -6,6 +6,7 @@ LABEL org.opencontainers.image.vendor=$IMAGE_VENDOR
 
 # Set environment variables
 ENV DEBIAN_FRONTEND=noninteractive
+ENV PATH=$PATH:~/.local/bin
```
abrown-RV:
```suggestion
```

abrown-RV (changes requested):
- `docker/example.dockerfile` needs the `COPY` on ln 32 updated with the new path to the protos
- workflow in docker_builer.yml needs the submodule checked out.
- suggest updating the `.devcontainer/README.md` to include the new `CMAKE_COMMAND` shortcut

.devcontainer/devcontainer.json
abrown-RV: ignore this for now

ctwarog-ross: as discussed we're keeping this

.devcontainer/toolchain-cpp.dockerfile
ctwarog-ross: do we not still  need this?

abrown-RV: No that was the beginnings of an idea of how we'd handle installation of yq. Putting `install-tooling.sh` in the toolchain takes care of it cause the toolchain runs as root, so it can write to `/usr/local/bin`.

docker/example-healthcheck.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 if [ $CATENA_CONNECTION = "gRPC" ]; then
-    exec grpcurl -plaintext -proto st2138/service.proto -import-path st2138 localhost:$CATENA_PORT catena.CatenaService/GetPopulatedSlots >/dev/null 2>&1 || exit 1
+    exec curl -f -X POST --http2-prior-knowledge http://localhost:$CATENA_PORT || exit 1
 elif [ $CATENA_CONNECTION = "REST" ]; then
     exec wget -q --spider http://localhost:$CATENA_PORT/st2138-api/v1/health || exit 1
```
abrown-RV:
```suggestion
    exec curl -f http://localhost:$CATENA_PORT/st2138-api/v1/health || exit 1
```
Forgot to update the REST healthcheck cause we took out wget, my bad

abrown-RV (changes requested):

abrown-RV (dismissed):

abrown-RV (approved):

PR #988, opened 2025-08-22
ExecuteCommand now checks for WriteAuthz

### LINKED ISSUE #982: Check for Write Authz before executing a command
Right now we do not do that, meaning executing a command needs nothing but a valid token.

#975 addresses adding ReadAuthz checks to device's getParam function (Previously a TODO item). I suggest adding a writeAuthz check to ParamDescriptor's ExecuteCommand function which returns catena::exception_with_status(PERMISSION_DENIED) and a nullptr if it fails.

BenWhitten-RV:
Had an hour to kill and was bored so I decided to fix this as my one last hoorah.

ExecuteCommand now checks for WriteAuthz before returning the commandResponder. This had been tested and the unittests have been updated for coverage.

This will probably interfere with Christian's SMPTE PR. The work here is all done though. Feel free to do with this what you like.


abrown-RV (approved):
well done on the merge conflicts

PR #989, opened 2025-08-22
[Example] im a teapot

ZuhayrSarkerRV:
This is an example PR that show how to make a good pr

I want to implement the 418 http response on the `/coffee` endpoint

---

## 🎯 Goal / Outcome
When hitting the `/coffee` endpoint with GET i should get a 418 responce

---

## ✅ Definition of Done (DoD)
- [x] make a test that expects the 418 responce
- [x] see that the test fails
- [x] make code changes
- [ ] ask for help from sr, devs if needed
- [x] update docs
- [x] run tests and see the 418 test now passes
- [x] make a nice pr using the template  

---

## 🛠️ Implementation Notes

This is just an example and shouldnt be merged 
@NR-RV is keeping this UpToDate with expectations and standereds

---

## 🔗 Related Issues / Links

> can add # of issues if this closed somthing

### DIFF
--- a/sdks/cpp/common/include/Status.h
+++ b/sdks/cpp/common/include/Status.h
@@ -213,6 +213,9 @@ enum StatusCode {
   /// The server was acting as a gateway/proxy and did not receive a timely
   /// response from the upstream server.
   GATEWAY_TIMEOUT = 504,
+
+  /// I'm a teapot.
+  I_AM_A_TEAPOT = 418,
 };
 
 /**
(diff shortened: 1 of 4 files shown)


abrown-RV (changes requested):
The implementation hardcodes the param `/coffee` as returning the 418, so it introduces the quirk in the device that no one can ever have their own param named `/coffee`, at least if they're connecting through REST. For this reason I think we can't accept as is.

KeonF-ross:
## 📖 Description

Addresses Issue \#989 by adding a 418 http response to the /coffee endpoint. A unit test for the endpoint was also implemented.

---

## ✅ Definition of Done (DoD)


- [x] Implement a 418 http response for the /coffee endpoint

---

## 🛠️ Implementation Notes


- Added I_AM_A_TEAPOT status code
- Modified Device::getParam to update status with the new status code if endpoint is /coffee
- Created unit test to ensure correct status code is generated
---

PR #1009, opened 2025-09-08
Issue mandatory params2

JChen2005:


abrown-RV (changes requested):
You've got changes for other issues that don't belong in here. The `.devcontainer`, `.vscode` and the controllers code. Recommend dealing with those first, will make your merge with develop easier.

PR #1012, opened 2025-09-09
938 disconnect issues

### LINKED ISSUE #938: disconnect issues
John has noticed that long running connections with no traffic are terminated by the gateway on our k8s cluster. At exactly 1 minute the `Connect` stream ends with an RST_STREAM code 2

Image

Nathan has confirmed that nginx is indeed terminating a stale connection. Nginx would probably allow for a longer timeout, but it will likely always be set to something, so it would probably be best not to rely on solving from that angle.

This issue is also compounded by another issue (#931) reported by John where stale connections are not thrown away by the connection manager. I'm assuming (@johndanenRV can confirm) that Dashboard automatically reconnects if it sees the disconnect, so the connection queue fills up after 16 minutes with state connections and no other connections are allowed.

A good solution for this would be some form of heartbeat/keepalive. GRPC itself does have keepalive capability built in. https://grpc.io/docs/guides/keepalive/ It looks like either the client or the server can be configured to send keepalive pings at a set interval. So we could set that up in our gRPC service with keepalive times. OR require that clients connecting should set up keepalive on their end.

Alternatively we could also implement a heartbeat/keepalive into just our connect stream. Something where if there are no updates for a certain amount of time it just sends an empty `PushUpdates` message. Maybe the ConnectPayload has a flag to turn on this behaviour or something.

abrown-RV:
Closes #938 

**My first real PR please be brutal**

Adds new Heartbeat timer, fires signals on the configured timeout.
Device now has `sendHeartbeat` function to send a "update" of the given param
Device `start/stop Heartbeat` to start an internally managed heartbeat timer on a hardcoded 5 second.
Device has `HeartbeatParam` setter and getter that must be set to something, `sendHeartbeat` reads this string.
More customized Heartbeats can create their own Heartbeat class and call `sendHeartbeat` manually.
Updated all examples to send heartbeats.


sdks/cpp/common/src/Device.cpp
```diff
@@ -355,6 +356,26 @@ std::unique_ptr<IParam> Device::getCommand(const std::string& fqoid, catena::exc
...
+        heartbeat_ = std::make_unique<Heartbeat>();
+        heartbeat_->getHeartbeatSignal().connect([this, &param]() {
+            this->sendHeartbeat(param);
+        });
+    }
+}
+
+void Device::startHeartbeat(const IParam& param) {
+    initHeartbeat(param);
+    heartbeat_->start(5000);
```
ctwarog-ross: Id rather startHeartbeat have an additional arg with default value set to 5000 in the declaration

sdks/cpp/common/src/Device.cpp
```diff
@@ -355,6 +356,26 @@ std::unique_ptr<IParam> Device::getCommand(const std::string& fqoid, catena::exc
     return command;
 }
 
+void Device::initHeartbeat(const IParam& param) {
+    if (!heartbeat_) {
+        heartbeat_ = std::make_unique<Heartbeat>();
+        heartbeat_->getHeartbeatSignal().connect([this, &param]() {
+            this->sendHeartbeat(param);
```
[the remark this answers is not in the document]
ctwarog-ross: copilot is right in its diagnosis but wrong in its fix

sdks/cpp/common/src/Heartbeat.cpp
```diff
@@ -0,0 +1,85 @@
...
+        return;
+    }
+    // create the thread
+    running_ = true;
+    thread_ = std::make_unique<std::thread>([this, milliseconds]() {
+        // the thread acquires the lock
+        std::unique_lock<std::mutex> lock(mutex_);
+        while (running_) {
+            // this is basically an interruptible sleep
+            condition_.wait_for(lock, std::chrono::milliseconds(milliseconds));
```
ctwarog-ross: condition_.wait_for(lock, std::chrono::milliseconds(milliseconds), [this]{ return !running_; });
to handle spurious wakeups

sdks/cpp/common/src/Heartbeat.cpp
```diff
@@ -0,0 +1,85 @@
...
+            // don't emit if we're shutting down
+            if (running_) {
+                try {
+                    // emit the signal
```
ctwarog-ross, line 61: unlock lock before signal emit and lock it after

ctwarog-ross (changes requested):
I didnt check UT yet, Ill review those after the requested changes

sdks/cpp/common/src/Device.cpp
abrown-RV: John was pretty firm in his statement that the timeout should be hardcoded 5 seconds. The heartbeat class itself allows you to specify an interval, so if you want a different interval that is the avenue.

sdks/cpp/common/src/Heartbeat.cpp
abrown-RV: ack

sdks/cpp/common/src/Heartbeat.cpp
abrown-RV: cool, good to know.

NR-RV (commented):
The UT look good to me

sdks/cpp/common/src/Device.cpp
ctwarog-ross: how about the class has a member var with default value 5000, and line 385 is heartbeat_->start(var). The user can change the var with a set method before starting, but the default implementation is 5000. In the current implementation the user has no way to actually specify other than 5000

unittests/cpp/common/tests/Device_test.cpp
```diff
@@ -264,6 +303,10 @@ class DeviceTest : public ::testing::Test {
 
 // ======== 0. Initial Setup ========
 
+// Static member definitions for DummyDevice
+MockHeartbeat* DeviceTest::DummyDevice::mockHeartbeat = nullptr;
```
ctwarog-ross: why are these static?

sdks/cpp/common/include/Device.h
```diff
@@ -167,7 +167,9 @@ class Device : public IDevice {
 
     void sendHeartbeat() override;
 
-    void startHeartbeat() override;
+    void startHeartbeat() override { startHeartbeat(5000); }
```
ctwarog-ross: no need to make 2 functions, just have 1 function declaration
void startHeartbeat(int32_t interval = 5000) override;
this makes it so you can call startHearbeat() and it defaults to 5000, or you can call startHeartbeat (someNumber) to set interval explicitly

ctwarog-ross (changes requested):

unittests/cpp/common/tests/Device_test.cpp
abrown-RV: don't need to be, thanks for the kick in the pants.

sdks/cpp/common/include/Device.h
abrown-RV: switched to defaults

ctwarog-ross (approved):
GTG, nice work on this Andrew!

PR #1017, opened 2025-09-11
Audio deck fix #971

### LINKED ISSUE #971: REST audio deck example needs to be updated to match gRPC.
Currently only the gPRC json example has the updated eq graph logic. Should be mapped over to REST to become more consistent. The YAML variants on both gRPC and REST need to also be matched.

nelgd:
Took a look at the gRPC JSON AudioDeck example and mimicked that for the other examples that needed to be updated or created.

Resolves issue #971


ctwarog-ross (approved):
great work here!

PR #1027, opened 2025-09-12
NMOS registration implemented and functional

### LINKED ISSUE #1025: Discovery Catena source code
Implement a NmosNode class in Catena that can be instantiated from either ServiceImpl or Device such that business logic can control it completely

ctwarog-ross:
Closes issue #1025

Created NmosNode class that connects to an instance of Sony Nmos Registry
NmosNode discovers and registers both a node and device into the registry
Avahi libraries used are network setup agnostic, work with mdns and dns-sd
discovery_REST example device that creates instance of NmosNode


.github/workflows/cpp-pull-request.yml
abrown-RV, line 1: gonna need to update the toolchain dockerfile too

sdks/cpp/common/include/NmosNode.h
abrown-RV, line 1: Have an interface class too, for mocking?

sdks/cpp/connections/REST/examples/discovery_REST/device.discovery.json
abrown-RV, line 1: don't forget the mandatory params... or maybe do forget them and let the codegen take care of it once Jason is done.

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,426 @@
...
+    auto sel = choose_registry_and_build_base(8080);
+    if (!sel) { LOG(ERROR) << "No registry discovered. Exiting.\n"; return 2; }
+
+    const char* bearerChars = std::getenv("NMOS_BEARER");
+    std::string bearer = bearerChars ? std::string(bearerChars) : std::string();
+
+    // Compose minimal resources
+    std::string node_id = random_uuid();
+    std::string dev_id = random_uuid();
+    // TODO: detect local IP to reach your Node API; using 127.0.0.1 as placeholder
```
abrown-RV: Friendly reminder machines can have multiple ips assigned. Might need some logic to decide which interface's IP to use.

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,426 @@
...
+    while (std::chrono::steady_clock::now() - start < std::chrono::seconds(2)) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(50));
+        if (stop_.load()) break;
+    }
+    avahi_simple_poll_quit(simple_poll_);
+    loop_thr.join();
+
+    // Choose a registry
+    auto sel = choose_registry_and_build_base(8080);
+    if (!sel) { LOG(ERROR) << "No registry discovered. Exiting.\n"; return 2; }
```
abrown-RV: Suggest well defined error codes with static consts defined on NmosNode

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,426 @@
...
+    auto tp = system_clock::now();
+    auto s = time_point_cast<seconds>(tp);
+    auto t = system_clock::to_time_t(s);
+    char buf[32]{};
+    std::strftime(buf, sizeof(buf), "%Y-%m-%dT%H:%M:%SZ", std::gmtime(&t));
+    return buf;
+}
+
+std::string NmosNode::random_uuid() {
+    // NOT cryptographically strong. Good enough for example code.
```
abrown-RV, line 151: Does it NEED to be cryptographically strong?

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,426 @@
...
+    auto r32 = [](){ unsigned v = (unsigned)std::rand(); return v; };
+    return fmt("%08x-%04x-%04x-%04x-%012x",
+    r32(), r32() & 0xFFFF, (r32() & 0x0FFF) | 0x4000,
+    (r32() & 0x3FFF) | 0x8000, ((unsigned long long)r32() << 16) | (r32() & 0xFFFF));
+}
+
+bool NmosNode::pick_primary_iface(std::string& ifname, std::string& mac){
+    DIR* d = opendir("/sys/class/net"); if (!d) return false;
```
abrown-RV: consider c++17 std::filesystem? (I don't know if its any better AI Overview suggested it XD)

Would this not pick the first interface with `oper == "up"` in alphabetical order? I don't think that's necessarily the *primary*

sdks/cpp/common/src/NmosNode.cpp
abrown-RV: I see you've got functions already, so nevermind!

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,426 @@
...
+        "\"id\": \"%s\","
+        "\"version\": \"%s\","
+        "\"label\": \"Catena Node\","
+        "\"description\": \"An NMOS node running Catena software\","
+        "\"href\": \"%s\","
+        "\"caps\": {},"
+        "\"tags\": {},"
+        "\"services\": ["
+            "{"
+                "\"href\": \"http://%s:%d/x-catena/status/\","
```
abrown-RV: So hold on a sec.... is this a URL on the catena service that the registry expects to be able to call?

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,426 @@
...
+    pick_primary_iface(ifname, mac);
+    if (pick_ipv4_for_iface(ifname, detected)) ip = detected;
+
+    std::string href = fmt("http://%s:%d/x-nmos/node/v1.3/", ip.c_str(), node_port);
+
+    return fmt(
+    "{"
+        "\"id\": \"%s\","
+        "\"version\": \"%s\","
+        "\"label\": \"Catena Node\","
```
abrown-RV: I feel like a bunch of these hardcoded values should be user configurable.

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,426 @@
...
+
+    pick_primary_iface(ifname, mac);
+    if (pick_ipv4_for_iface(ifname, detected)) ip = detected;
+
+    std::string href = fmt("http://%s:%d/x-nmos/node/v1.3/", ip.c_str(), node_port);
+
+    return fmt(
+    "{"
+        "\"id\": \"%s\","
+        "\"version\": \"%s\","
```
abrown-RV: weird that "version" is just the current time in a specific format, what are the nmos guys doing? XD

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,426 @@
...
+        if (stop_.load()) break;
+    }
+    avahi_simple_poll_quit(simple_poll_);
+    loop_thr.join();
+
+    // Choose a registry
+    auto sel = choose_registry_and_build_base(8080);
+    if (!sel) { LOG(ERROR) << "No registry discovered. Exiting.\n"; return 2; }
+
+    const char* bearerChars = std::getenv("NMOS_BEARER");
```
abrown-RV, line 123: Where does this come from?

sdks/cpp/common/src/utils.cpp
```diff
@@ -118,4 +118,10 @@ std::string catena::from_base64(const std::string& encoded) {
         }
     }
     return decoded;
+}
+
+std::string catena::fmt(const char* f, ...) {
```
abrown-RV, line 129: Is it worth just grabbing a json library? Might make life easier. There are "header-only" libraries, that means there's no linking right? So no need for any extra `apt install`?

sdks/cpp/common/src/NmosNode.cpp
ctwarog-ross: no, hence why I left it for now

sdks/cpp/common/src/NmosNode.cpp
ctwarog-ross: like throwing an enum code or something?

sdks/cpp/common/src/NmosNode.cpp
ctwarog-ross: yes

sdks/cpp/common/src/NmosNode.cpp
ctwarog-ross: NMOS_BEARER was implemented with the use case in mind that until we get an authorization service implemented,  you pass in an arbitrary bearer code as an env var during container creation

sdks/cpp/common/src/utils.cpp
ctwarog-ross: Im thinking at this point thats the best idea yeah

sdks/cpp/common/src/NmosNode.cpp
abrown-RV: So this node stuff needs to then make a whole http server. Listening on another port OR hook into our existing rest server for the REST half. I also suggest making a totally standalone container image for the node so we can sidecar it.

ctwarog-ross: maybe we can have that as a separate issue, making PRs too big is less than ideal

sdks/cpp/common/src/NmosNode.cpp
ctwarog-ross: version is essentially the time signature of the last time the node made any config updates with the registry. It is expected in a seconds:nanoseconds format

.github/workflows/cpp-pull-request.yml
abrown-RV: https://github.com/rossvideo/Catena-deployments/issues/4

sdks/cpp/common/include/INmosNode.h
```diff
@@ -0,0 +1,124 @@
...
+
+    virtual AvahiClient* getClient() = 0;
+
+    virtual bool get_node_iface(const std::string& reg_addr) = 0;
+
+    virtual std::string make_node_json(const std::string& node_id, int node_port) = 0;
+
+    virtual std::string make_device_json(const std::string& dev_id, const std::string& node_id, int node_port) = 0;
+
+  protected:
```
abrown-RV, line 156: I don't think protected should go on interfaces.

sdks/cpp/common/include/INmosNode.h
```diff
@@ -0,0 +1,124 @@
...
+
+    virtual std::vector<RegistryCandidate> getCandidates() = 0;
+
+    virtual AvahiClient* getClient() = 0;
+
+    virtual bool get_node_iface(const std::string& reg_addr) = 0;
+
+    virtual std::string make_node_json(const std::string& node_id, int node_port) = 0;
+
+    virtual std::string make_device_json(const std::string& dev_id, const std::string& node_id, int node_port) = 0;
```
abrown-RV: These look like internal helper functions

sdks/cpp/common/include/NmosNode.h
```diff
@@ -0,0 +1,216 @@
...
+ * @file Authorizer.h
+ * @brief Class for handling authorization information
+ * @author John R. Naylor john.naylor@rossvideo.com
+ * @author John Danen
+ * @date 2024-09-18
+ * @author Ben Whitten benjamin.whitten@rossvideo.com
+ * @author Zuhayr Sarker zuhayr.sarker@rossvideo.com
+ * @date 2025-02-20
+ * @copyright Copyright (c) 2024 Ross Video
+ */
```
abrown-RV, line 39: Clean up your copy paste XD

sdks/cpp/common/src/NmosNode.cpp
abrown-RV: Oh I see later in the code, IF there's a bearer it'll use it.

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,422 @@
...
+        (AvahiClientFlags)0, client_cb, this, &error);
+
+    if (!client_) {
+        LOG(ERROR) << "Failed to create client: " << avahi_strerror(error);
+        return false;
+    }
+
+    DEBUG_LOG << "Avahi client created successfully.";
+
+    // Browse for _http._tcp services, change as needed
```
abrown-RV:
```suggestion
    // Browse for _nmos-registration._tcp services
```

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,422 @@
...
+        LOG(ERROR) << "Failed to create service browser: " << avahi_strerror(avahi_client_errno(client_));
+        return false;
+    }
+
+    DEBUG_LOG << "Service browser created successfully.";
+
+    // Run discovery for a short window (e.g., 2 seconds) then proceed
+    auto start = std::chrono::steady_clock::now();
+    std::thread loop_thr([&](){ avahi_simple_poll_loop(simple_poll_); });
+    while (std::chrono::steady_clock::now() - start < std::chrono::seconds(2)) {
```
abrown-RV: if you calculate now + 2 seconds before starting and compare against that. Instead of subtracting start from now. You'll save +-40 subtract operations. Don't know if we care XD

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,422 @@
...
+    DEBUG_LOG << "Avahi client state changed: " << (int)state;
+    if (state == AVAHI_CLIENT_FAILURE) {
+        LOG(ERROR) << "Client failure: " << avahi_strerror(avahi_client_errno(client));
+        avahi_simple_poll_quit(node->getPoll());
+    }
+}
+
+// ---------------------- Registration + heartbeat ----------------------
+
+std::optional<NmosNode::RegistrySelection> NmosNode::choose_registry_and_build_base(int node_port) {
```
abrown-RV: `node_port` is not used.

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,422 @@
...
+    const char* bearerChars = std::getenv("NMOS_BEARER");
+    std::string bearer = bearerChars ? std::string(bearerChars) : std::string();
+
+    // Compose minimal resources
+    std::string node_id = random_uuid();
+    std::string dev_id = random_uuid();
+
+    if (!get_node_iface(candidates_[0].addr)) return false;
+
+    json node_json = make_node_json(node_id, 8080);
```
abrown-RV: This port 8080 just a hardcoded dummy for now? until we sort out the Registry -> Node stuff? Still might be good to variable it so its easy to update when we do figure it out.

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,422 @@
...
+        LOG(ERROR) << "Failed to register Device\n";
+    }
+
+    // Start heartbeat loop
+    std::thread hb(&NmosNode::heartbeat_thread, this, sel->base, node_id, bearer, heartbeatInterval);
+
+    DEBUG_LOG << "Registered Node " << node_id << ". Heartbeating. Press Ctrl+C to exit.";
+
+    // Wait for signal
+    while (!stop_.load()) std::this_thread::sleep_for(std::chrono::milliseconds(200));
```
abrown-RV: I think the idea is switch to the `catena::common::Heartbeat` which will work a little differently. But that's probably a good thing, cause won't this block the calling thread until `stop_ == true`? I feel like that's bad.

sdks/cpp/common/include/INmosNode.h
```diff
@@ -0,0 +1,124 @@
...
+
+    virtual bool init(std::chrono::milliseconds heartbeatInterval) = 0;
+
+    virtual AvahiSimplePoll* getPoll() = 0;
+
+    virtual void addCandidate(RegistryCandidate&& c) = 0;
+
+    virtual std::vector<RegistryCandidate> getCandidates() = 0;
+
+    virtual AvahiClient* getClient() = 0;
```
abrown-RV: Needs Doxygen comments

sdks/cpp/common/include/INmosNode.h
ctwarog-ross: why not?

abrown-RV: Nothing preventing it of course since they're not real interfaces, but interfaces describe the public ...interface that code uses to interact with the class. So hidden members may as well be left up to the implementors. For us I think we mainly do interfaces for mocking. And there's no point in mocking non public members.

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,422 @@
...
+    // NOT cryptographically strong. Good enough for example code.
+    auto r32 = [](){ uint32_t v = (uint32_t)std::rand(); return v; };
+    return fmt("%08x-%04x-%04x-%04x-%012llx",
+    r32(), r32() & 0xFFFF, (r32() & 0x0FFF) | 0x4000,
+    (r32() & 0x3FFF) | 0x8000, ((uint64_t)r32() << 16) | (r32() & 0xFFFF));
+}
+
+//figure out how to get subnet and match it to the current ip in the loop
+bool NmosNode::get_node_iface(const std::string& reg_addr){
+    std::string subnet = reg_addr.substr(0, reg_addr.rfind('.'));
```
abrown-RV, line 160: subnet mask is not always /24. It's probably the most common, but should probably allow for other subnet masks.

sdks/cpp/common/include/NmosNode.h
```diff
@@ -0,0 +1,216 @@
...
+
+#include <curl/curl.h>
```
abrown-RV, line 41: can this move to the cpp?

abrown-RV (commented):
Some comments, you were right! XD

unittests/cpp/common/tests/NmosNode_test.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+    bool start() {
+        listen_fd = socket(AF_INET, SOCK_STREAM, 0);
+        if (listen_fd < 0) return false;
+        int opt = 1; setsockopt(listen_fd, SOL_SOCKET, SO_REUSEADDR, &opt, sizeof(opt));
+        sockaddr_in a{}; a.sin_family = AF_INET; a.sin_port = htons(3210);
+        inet_pton(AF_INET, "127.0.0.1", &a.sin_addr);
+        if (bind(listen_fd, (sockaddr*)&a, sizeof(a)) < 0) return false;
+        if (listen(listen_fd, 16) < 0) return false;
+        thr = std::thread([this]{
```
abrown-RV, line 78: Suggest grabbing an http server lib if you ever want to mock more than a single endpoint (or even for the single to make it cleaner). Nathan found one for the web example

abrown-RV (commented):
Super cool trick with the --wrap. I guess you could also use to to totally swap out all of curl as well. But actually I think just spinning up a "mock" http server is actually less work, looking at all the different curl functions 😕 

This is still dropping the line coverage by at least a %, have to look at the nmosnode coverage to have ideas on covering more

.devcontainer/toolchain-cpp.dockerfile
```diff
@@ -28,7 +28,10 @@ RUN . /root/toolchain.env \
     ninja-build=$NINJA_VERSION \
     gdb=$GNU_DEBUGGER_VERSION \
     valgrind=$VALGRIND_VERSION \
-    curl=$CURL_VERSION \
+    curl=$CURL_VERSION libcurl4-openssl-dev\
+    libavahi-client-dev libavahi-common-dev \
+    nlohmann-json3-dev \
```
abrown-RV: Does more version locking make sense for avahi or json?

unittests/cpp/REST/CMakeLists.txt
```diff
@@ -33,7 +33,6 @@ foreach(test_file ${REST_TEST_FILES})
     get_filename_component(test_name ${test_file} NAME_WE)
     add_executable(REST_${test_name} tests/${test_file})
     target_link_libraries(REST_${test_name} PRIVATE
-        catena_proto_common
```
abrown-RV, line 36: Why?

sdks/cpp/common/src/utils.cpp
```diff
@@ -26,6 +26,12 @@
 #include <cstring>
 
 std::string catena::readFile(std::filesystem::path path) {
+    try { 
+        if (!std::filesystem::exists(path)) return ""; 
+    } catch (const std::exception& e) { 
+        LOG(ERROR) << "File does not exist: " << e.what(); 
+        return ""; 
+    }
```
abrown-RV, line 34: Did you verify the other uses of this function can handle the functionality change?

sdks/cpp/common/src/utils.cpp
```diff
@@ -118,4 +124,10 @@ std::string catena::from_base64(const std::string& encoded) {
         }
     }
     return decoded;
+}
+
+std::string catena::fmt(const char* f, ...) {
```
abrown-RV, line 129: Unit tests? kind of covered indirectly by all the other code, but still good to make some direct unit tests. @NR-RV agreed?

sdks/cpp/common/src/utils.cpp
```diff
@@ -26,6 +26,12 @@
 #include <cstring>
 
 std::string catena::readFile(std::filesystem::path path) {
+    try { 
+        if (!std::filesystem::exists(path)) return ""; 
+    } catch (const std::exception& e) { 
+        LOG(ERROR) << "File does not exist: " << e.what();
```
abrown-RV: Not what the exception means.

unittests/cpp/common/tests/NmosNode_test.cpp
```diff
@@ -0,0 +1,319 @@
...
+        if (hb.joinable()) hb.join();
+        stop_.store(false); // reset
+
+        return ok1 && ok2;
+    }
+};
+class NmosNodeTest : public ::testing::Test {
+protected:
+    void SetUp() override {
+        g_avahi_test_control.reset();
```
abrown-RV, line 167: glog init? and teardown?

unittests/cpp/common/mocks/WrapAvahiClient.h
abrown-RV, line 1: Would it be possible to still use mocks here?

the `MOCK_METHOD` macro does seem to resolve to normal function definition, albeit with some weird templating around the return and param types. That could mess with the wrap linking part.

But if that doesn't work could you make it so the struct is a collection of the `MOCK_METHOD`s and then the `extern "C"` part just calls those mocks? Then you could do all the regular and excellent `EXPECT_CALL` awesomeness.

unittests/cpp/common/tests/NmosNode_test.cpp
```diff
@@ -0,0 +1,319 @@
...
+TEST_F(NmosNodeTest, NmosNodeTest_DiscoveredRegistryBrowseFail) {
+    TestableNmosNode node;
+
+    g_avahi_test_control.discovered_service = true;
+    g_avahi_test_control.has_iface = true;
+    g_avahi_test_control.browse_event = AVAHI_BROWSER_FAILURE; // force browse failure
+
+    EXPECT_FALSE(node.init());
+}
+
```
abrown-RV, line 279: Throw in one with `AVAHI_BROWSER_REMOVE` `AVAHI_BROWSER_CACHE_EXHAUSTED` or `AVAHI_BROWSER_ALL_FOR_NOW` for free extra coverage in `browse_cb` for those other switch cases

unittests/cpp/common/tests/NmosNode_test.cpp
```diff
@@ -0,0 +1,319 @@
...
+    // Register + heartbeat against the fake server
+    bool ok = node.register_against(sel->base, std::chrono::milliseconds(120));
+    EXPECT_TRUE(ok);
+    EXPECT_GE(reg.resource_posts.load(), 2); // node + device
+    EXPECT_GE(reg.heartbeats.load(), 1);
+
+    reg.stop_and_join();
+}
+
```
abrown-RV, line 369: - Add a manual call to `resolve_cb` with `event != AVAHI_RESOLVER_FOUND` to get free coverage on the else of the if in `resolve_cb`.
- Add a test with a couple of manual calls to `resolve_cb` with different candidates to test the sorting of candidates and get coverage on the sort callback. Could even be a couple of tests to cover all the permutations of the sorting logic.

unittests/cpp/common/tests/NmosNode_test.cpp
```diff
@@ -0,0 +1,319 @@
...
+                std::string buf; buf.resize(4096);
+                ssize_t r = recv(fd, buf.data(), buf.size(), 0);
+                if (r <= 0) { close(fd); continue; }
+                buf.resize(size_t(r));
+                auto sp1 = buf.find(' ');
+                auto sp2 = buf.find(' ', sp1 + 1);
+                std::string method = buf.substr(0, sp1);
+                std::string path = buf.substr(sp1 + 1, sp2 - sp1 - 1);
+
+                if (method == "POST" && path == "/x-nmos/registration/v1.3/resource") {
```
abrown-RV, line 93: If this could succeed on the first call, and then fail on the 2nd we could cover the 2nd post failure in `sendRequests`. There some way to get a mock in here? It supports that behaviour, AND counting number of invokes.

abrown-RV (changes requested):
Hopefully this doesn't hide my earlier comments. That wrap linker thing is super cool. Some comments on getting to 100% coverage in `NmosNode.cpp`, I think its doable.

sdks/cpp/common/include/INmosNode.h
ctwarog-ross: you're right, I had them like this to give myself a blank slate for mocking. Bad design, removing from interface

sdks/cpp/common/src/NmosNode.cpp
ctwarog-ross: have it in the arg with default value of some sort?

abrown-RV: I assume its something you can determine from the network settings. Or just like take the IP with the highest number of matching bits or something...

unittests/cpp/REST/CMakeLists.txt
ctwarog-ross: not needed

unittests/cpp/common/tests/NmosNode_test.cpp
ctwarog-ross: is that a case worth covering? Its alot of wizardry to confirm handling of very simple logging and return code

abrown-RV: true.... but 100% coverage! XD

sdks/cpp/common/src/utils.cpp
ctwarog-ross: There is one test for it, I changed from EXPECT_THROW to EXPECT_EQ. In the areas of the code that use it, they are just more robust since now you callers arent responsible for catching the error

sdks/cpp/common/src/utils.cpp
abrown-RV: [`std::filesystem::exists()`](https://en.cppreference.com/w/cpp/filesystem/exists.html) returns false if it doesn't exist. The exceptions are from other OS errors. So coverage doesn't hit the catch.

sdks/cpp/common/src/utils.cpp
abrown-RV: `ServiceCredentials` popping up again. I mean, the `readFile` function is also used here. Does it still work with the changed `readFile` behaviour?

sdks/cpp/common/src/utils.cpp
ctwarog-ross: I wasn't thinking when writing this, you're right needs rewrite

.devcontainer/toolchain-cpp.dockerfile
ctwarog-ross: I dont think at this time

unittests/cpp/common/mocks/WrapAvahiClient.h
ctwarog-ross: The avahi lib canot be mocked as the functions are not virtual. The AvahiTestControl struct could contain mocked versions of the avahi classes, unsure what I would get from that though as theyre mostly structs that just contain data. Is that what you meant?

sdks/cpp/common/src/utils.cpp
ctwarog-ross: The only thing that comes to mind is if somewhere up the calling chain was handling an error from the file read if it didnt exist, that block of code will never be executed now.

abrown-RV: Exactly, and also now code that maybe was once expecting to not run if the file didn't exist, might now run with the empty string. So that code needs to be able to not die if given `""`

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -0,0 +1,451 @@
...
+            std::string scheme = c.https ? "https" : "http";
+            std::string base = fmt("%s://%s:%u/x-nmos/registration/%s", scheme.c_str(), c.host.c_str(), c.port, ver.c_str());
+            DEBUG_LOG << "Chosen registry base: " << base;
+            return NmosNode::RegistrySelection{base, ver};
+        }
+    }
+    return std::nullopt;
+}
+
+void NmosNode::run_heartbeat(std::string base, std::string node_id, std::string bearer, std::chrono::milliseconds interval) {
```
abrown-RV: Not using the Heartbeat class?

abrown-RV (commented):
Not too sure where the segfault is coming from. Maybe using the heartbeat class will fix it?

sdks/cpp/common/src/NmosNode.cpp
ctwarog-ross: No9 we went over it, I had implemented it but it meant NmosNode was more complicated to set up and not self contained. Heartbeat is part of the Nmos spec so I thought it best to keep that self contained

abrown-RV (approved):
💯

PR #1029, opened 2025-09-15
Validate-Generate-Mandatory-Param

### LINKED ISSUE #1000: Validate/Generate manadatory params
Since the mandatory device params are... mandatory, it would be good to enforce their inclusion. We could add to the code generator the ability to automatically generate the params into the device. Suggest either a single flag pointing at a variables file with the values or individual flags for each. Benefit with the single flag is if new mandatory params are added, it's slightly less work to reflect that change in the codegen.

I think that any automatic generation should be merged with anything the user has done themselves in the device definition file. Then after the merging with the device definition, fail the generation if they don't have the mandatory params. There should also be a flag to disable this behaviour if someone really means it.

For reference the mandatory params:
Image

Specifically for the `/product/catena_sdk` and `/product/catena_sdk_version` params, that's something that can be determined from the catena sdk, so we should try to automatically provide the value for those.

JChen2005:
Resolved issue #1000 

Updated the codegen to generate and validate the mandatory parameters. 
Added an areAllRequiredParamsPresent function that checks if all parameters and their values (except catena_sdk & catena_sdk_version) are present. 
Added a disableMandatoryEnforcement command. 
Added product structs to all example JSON/YAML file in the common directory.


tools/codegen/index.js
```diff
@@ -24,15 +24,16 @@ program //Defines cmd line options (Default values are provided for each)
     .option('-s, --schema <string>', 'path to schema definitions', '../../smpte/interface/schemata/device.json') //Schema file path
     .option('-d, --device-model <string>', 'Catena device model to process', '../../example_device_models/device.minimal.json') //JSON device model input file path
     .option('-l, --language <string>', 'Language to generate code for', 'cpp') // code generation language
-    .option('-o, --output <string>', 'Output folder for generated code', '.') // where to output the generated code
+    .option('-o, --output <string>', 'Output folder for generated code', '../../build/cpp') // where to output the generated code
     .option('-m, --mandatory-params <string>', 'patht to files')
-    .option('disable-mandatory-enforcement', 'Disable enforcement of mandatory parameters during code generation', false)
+    .option('disable-mandatory-enforcement', 'Disable enforcement of mandatory parameters during code generation' , true)
```
NR-RV: this should be off by default

tools/codegen/index.js
```diff
@@ -24,15 +24,16 @@ program //Defines cmd line options (Default values are provided for each)
     .option('-s, --schema <string>', 'path to schema definitions', '../../smpte/interface/schemata/device.json') //Schema file path
     .option('-d, --device-model <string>', 'Catena device model to process', '../../example_device_models/device.minimal.json') //JSON device model input file path
     .option('-l, --language <string>', 'Language to generate code for', 'cpp') // code generation language
-    .option('-o, --output <string>', 'Output folder for generated code', '.') // where to output the generated code
+    .option('-o, --output <string>', 'Output folder for generated code', '../../build/cpp') // where to output the generated code
```
NR-RV: this dosnt need a default

tools/codegen/index.js
NR-RV, line 1: there is scope change :D
ill run you though it i a sec

NR-RV (changes requested):

tools/codegen/cpp/cppgen.js
abrown-RV, line 1: unused code, can just take it back out

tools/codegen/index.js
```diff
@@ -19,16 +19,25 @@
...
+    .option('-o, --output <string>', 'Output folder for generated code', '.') 
+    .option('-m, --mandatory-params <string>', 'patht to files')
+    .option('--disable-mandatory-enforcement', 'Disable enforcement of mandatory parameters during code generation')
+    // Individual mandatory parameter flags
+    .option('--product-name <string>', 'Catena')
+    .option('--product-vendor <string>', 'Ross Video')
+    .option('--product-version <string>', '1.0.0')
+    .option('--catena-sdk <string> ' , '')
+    .option('--catena-sdk-version <string> ' , '')
+    .option('--product-serial <string>', 'SN-7K9M-2024-XR485-BLU')
```
abrown-RV, line 36: no longer needed with scope change

tools/codegen/index.js
```diff
@@ -19,16 +19,25 @@
...
+const { program } = require('commander');  
+program 
     .option('-s, --schema <string>', 'path to schema definitions', '../../smpte/interface/schemata/device.json')
-    .option('-d, --device-model <string>', 'Catena device model to process', '../../example_device_models/device.minimal.json')
-    .option('-l, --language <string>', 'Language to generate code for', 'cpp')
-    .option('-o, --output <string>', 'Output folder for generated code', '.');
+    .option('-d, --device-model <string>', 'Catena device model to process', '../../example_device_models/device.minimal.json') 
+    .option('-l, --language <string>', 'Language to generate code for', 'cpp') 
+    .option('-o, --output <string>', 'Output folder for generated code', '.') 
+    .option('-m, --mandatory-params <string>', 'patht to files')
```
abrown-RV, line 28: No longer needed with scope change

tools/codegen/index.js
```diff
@@ -110,47 +127,128 @@ class DeviceModel {
...
-//         const codeGen = new CodeGen(dm, options.output);
-//         codeGen.generate();
-//     }
-
-// } catch (why) {
-//     console.log(why.message);
-//     process.exit(typeof why.error === 'number' ? why.error : 1);
-// }
+const deviceName = path.parse(options.deviceModel).name.split('.')[0]; //Extracts the name prefix (eg device.minimal.json -> device)
+console.log(`Validating device model '${deviceName}' from file '${options.deviceModel}' against schema file '${options.schema}'...`); //Begins validation
```
abrown-RV, line 131:
```suggestion
//Extracts the name prefix (eg device.minimal.json -> device)
const deviceName = path.parse(options.deviceModel).name.split('.')[0];
//Begins validation
console.log(`Validating device model '${deviceName}' from file '${options.deviceModel}' against schema file '${options.schema}'...`);
```
I prefer comments like this on their own lines so that it is obvious the code hasn't actually changed.

tools/codegen/index.js
```diff
@@ -110,47 +127,128 @@ class DeviceModel {
...
+  const REQUIRED_PARAMS = [
+    "name",
+    "vendor",
+    "version",
+    "catena_sdk",
+    "catena_sdk_version",
+    "serial_number"
+  ];
+
+  //console.log('🔍 DEBUG areAllRequiredParamsPresent called - hasProduct:', !!(deviceParams && deviceParams.product));
```
abrown-RV, line 143: Don't leave in commented out code.

tools/codegen/index.js
```diff
@@ -40,16 +41,21 @@ if (options.language) {
 if (options.output) {
     console.log(`output: ${options.output}`);
 }
+if(options.mandatoryParams) {
+    console.log(`mandatoryParams: ${options.mandatoryParams}`);
+}
```
abrown-RV, line 46: Not needed anymore.

tools/codegen/index.js
```diff
@@ -24,7 +24,8 @@ program
     .option('-s, --schema <string>', 'path to schema definitions', '../../smpte/interface/schemata/device.json')
     .option('-d, --device-model <string>', 'Catena device model to process', '../../example_device_models/device.minimal.json')
     .option('-l, --language <string>', 'Language to generate code for', 'cpp')
-    .option('-o, --output <string>', 'Output folder for generated code', '.');
+    .option('-o, --output <string>', 'Output folder for generated code', '../../build/cpp')
+    .option('disable-mandatory-enforcement', 'Disable enforcement of mandatory parameters during code generation' , true)
```
abrown-RV, line 28: should have `--` in front of the option

tools/codegen/index.js
```diff
@@ -59,6 +64,7 @@ const yaml = require('yaml')
  * @class DeviceModel
  * @brief Holds the information parsed from a json device model file
  */
+
```
abrown-RV, line 67:
```suggestion
```

tools/codegen/index.js
```diff
@@ -111,44 +117,119 @@ class DeviceModel {
...
-//     }
-
-// } catch (why) {
-//     console.log(why.message);
-//     process.exit(typeof why.error === 'number' ? why.error : 1);
-// }
+const deviceName = path.parse(options.deviceModel).name.split('.')[0]; //Extracts the name prefix (eg device.minimal.json -> device)
+console.log(`Validating device model '${deviceName}' from file '${options.deviceModel}' against schema file '${options.schema}'...`); //Begins validation
+
+function areAllRequiredParamsPresent(deviceParams, disableMandatoryEnforcement = false) {
```
abrown-RV, line 123: function needs a doc comment block

tools/codegen/index.js
```diff
@@ -111,44 +117,119 @@ class DeviceModel {
...
+  const REQUIRED_PARAMS = [
+    "name",
+    "vendor",
+    "version",
+    "catena_sdk",
+    "catena_sdk_version",
+    "serial_number"
+  ];
+
+  if (!deviceParams || !deviceParams.product) {
```
abrown-RV, line 133: should also check `product.type` is "STRUCT"

tools/codegen/index.js
abrown-RV, line 1: Be consistent with the existing indentation. The file was originally 4 space indentation, most of your additions are 2. Hot tip: you can highlight sections of code, right click, and "Format Selection". Do it only in the selection so you don't format code that isn't part of the PR.

tools/codegen/index.js
```diff
@@ -111,44 +117,119 @@ class DeviceModel {
...
+    if (!disableMandatoryEnforcement) {
```
abrown-RV, line 134: The function should just not do any checks at all if `disabledMandatoryEnforcement`. I would just put
```js
if (disableMandatoryEnforcement) {
    return;
}
```
at the start to skip everything. Then you don't do any checks if the checks are disabled.

tools/codegen/index.js
```diff
@@ -111,44 +117,119 @@ class DeviceModel {
...
+  const productValue = deviceParams.product.value; 
+  const missing = [];
+  const emptyValues = [];
 
+  REQUIRED_PARAMS.forEach(key => {
+    const param = productParams[key];
+    
+    if (!param) {
+      missing.push(key);
+    } else if (key !== 'catena_sdk_version' && key !== 'catena_sdk') {
```
abrown-RV, line 150: I don't like hardcoding these here. Should put them in a const with the other one, or integrate the information into the existing `REQUIRED_PARAMS`.

tools/codegen/index.js
```diff
@@ -111,44 +117,119 @@ class DeviceModel {
...
+    const param = productParams[key];
+    
+    if (!param) {
+      missing.push(key);
+    } else if (key !== 'catena_sdk_version' && key !== 'catena_sdk') { 
+    let stringValue;
+
+    if (param.value && param.value.string_value) {
+        stringValue = param.value.string_value;
+    }
```
abrown-RV, line 155: Don't let them get away with putting the value here, should only be in the `struct_value.fields`.

But you should check that the param is type `STRING`.

tools/codegen/index.js
```diff
@@ -111,44 +117,119 @@ class DeviceModel {
...
+            console.log('❌ Validation failed.');
+            if (!validationPassed) {
+                console.log('❌ Code generation stopped due to schema validation errors.');
+            } else {
+                console.log('❌ Code generation stopped due to mandatory parameter errors.');
+            }
+            process.exit(1);
+        }
+        
+        // Continue with code generation even if mandatory params are missing (build compatibility)
```
abrown-RV, line 220: All this error logic is overkill. I'd be happy with just the thrown `Error` getting caught in the existing catch at the end and call it a day.

abrown-RV (changes requested):
You've got the core validation logic pretty much down. 👍  Gonna scope creep on you a bit and I've asked for enforcing the types on the params as well as checking their existence. Most of my comments are styling and simplification related.

tools/codegen/cpp/cppgen.js
NR-RV: the file came back

tools/codegen/index.js
```diff
@@ -40,11 +40,12 @@ if (options.language) {
...
     console.log(`output: ${options.output}`);
 }
+if (options.disableMandatoryEnforcement) {
+    console.log(`Mandatory parameter enforcement disabled`);
+}
 
-// import the fs libraries
 const fs = require('fs');
 
-// verify input file exists
```
abrown-RV, line 47: Restore these comments.

tools/codegen/index.js
```diff
@@ -77,23 +78,23 @@ class DeviceModel {
...
+      * @todo add support for other import types 
+      */
+     importParam(importArg) {
+         if ("file" in importArg) {
             const importDir = path.dirname(this.filePath);
             const importPath = `${importDir}/${importArg.file}`;
             if (!fs.existsSync(importPath)) {
               throw new Error(`Cannot open file at ${importArg.file}`);
             }
-            // Determining if the file is yaml or json and parsing accordingly.
```
abrown-RV, line 96: Why are you removing perfectly good comments from code you're not touching?

tools/codegen/index.js
```diff
@@ -104,51 +105,97 @@ class DeviceModel {
...
+    const REQUIRED_PARAMS = {
+        "name": true,
+        "vendor": true,
+        "version": true,
+        "catena_sdk": false,
+        "catena_sdk_version": false,
+        "serial_number": true
+    };
+
+    if (!deviceParams || !deviceParams.product) {
```
abrown-RV, line 136: This still needs a product type is `STRUCT` check

tools/codegen/index.js
```diff
@@ -104,51 +105,97 @@ class DeviceModel {
...
             throw new Error(`Unsupported import type: ${importArg}`);
-          }
-      }
+         }
+     }
 }
 
 // extract schema name from input filename
 const deviceName = path.parse(options.deviceModel).name.split('.')[0];
-console.log(`Validating device model '${deviceName}' from file '${options.deviceModel}' against schema file '${options.schema}'...`);
```
abrown-RV, line 115: Restore this console.log

tools/codegen/index.js
```diff
@@ -104,51 +105,97 @@ class DeviceModel {
...
+    const allIssues = [...missing.map(p => `${p} (missing field)`), ...emptyValues];
+
+    if (allIssues.length > 0 && !disableMandatoryEnforcement) {
+        throw new Error(`Invalid mandatory product parameters: ${allIssues.join(', ')}`);
+    }
+}
 
 try {
     const validator = new Validator(options.schema);
-    console.log(`Applying schema '${deviceName}' to file '${options.deviceModel}'`);
```
abrown-RV, line 142: Restore this console.log

tools/codegen/index.js
```diff
@@ -104,51 +105,97 @@ class DeviceModel {
...
+    const missing = [];
+    const emptyValues = [];
+
+    Object.entries(REQUIRED_PARAMS).forEach(([key, checkValue]) => {
+        const param = productParams[key];
+
+        if (!param) {
+            missing.push(key);
+        } else if (checkValue) {
+            if (param.type !== 'STRING') {
```
abrown-RV, line 151: this check needs to happen even if `checkValue` is false.

tools/codegen/index.js
```diff
@@ -103,52 +104,98 @@ class DeviceModel {
...
-//         const CodeGen =  require(`./${options.language}/${options.language}gen.js`);
-//         const dm = new DeviceModel(options.deviceModel, validator, data);
-//         const codeGen = new CodeGen(dm, options.output);
-//         codeGen.generate();
-//     }
-
-// } catch (why) {
-//     console.log(why.message);
-//     process.exit(typeof why.error === 'number' ? why.error : 1);
-// }
```
abrown-RV, line 137: Don't just delete stuff.

tools/codegen/cpp/cppgen.js
```diff
@@ -394,4 +394,4 @@ class CppGen {
   };
 }
 
-module.exports = CppGen;
+module.exports = CppGen;
```
abrown-RV, line 397: just missing a newline at the end of the file now.

abrown-RV (changes requested):

tools/codegen/index.js
```diff
@@ -137,18 +141,95 @@ console.log(`Validating device model '${deviceName}' from file '${options.device
...
+
+    if (allIssues.length > 0 && !disableMandatoryEnforcement) {
+        throw new Error(`Invalid mandatory product parameters: ${allIssues.join(', ')}`);
+    }
+}
+
 try {
     const validator = new Validator(options.schema);
     console.log(`Applying schema '${deviceName}' to file '${options.deviceModel}'`);
     const isValid = validator.validate(deviceName, options.deviceModel);
```
abrown-RV, line 216: This one actually *should* be removed cause you have changed the logic below

abrown-RV (approved):

NR-RV (approved):

PR #1032, opened 2025-09-16
added the locks and maked the names better

### LINKED ISSUE #1031: lock bypass bug
there is a way to bypass the lock in `CallStatus::kWrite`
and update `res_` wile it is writing to the stream buffer making the service die

NR-RV:
added the locks and made the names better
moved the `mtx_` to the parent I*.h and removed them form the children

Closes #1031


abrown-RV (approved):
We did this together so I hope we all approve! XD

PR #1038, opened 2025-09-17
Struct validation #1004

### LINKED ISSUE #1004: read only doesn't apply to children of a struct
Verified with REST. When you have a struct that is declared read only, its children fields can still be modified.

Start by making unit tests that will initially fail. They should modify values and then assert that you can't, but then since the bug exists they will fail. Make tests for both gRPC and REST and all nested param types as well.

THEN you can go in and actually fix the bug. The unit tests will help you discover how widespread the issue is

nelgd:
Resolves issue #1004


sdks/cpp/common/src/Device.cpp
```diff
@@ -72,6 +72,30 @@ bool Device::tryMultiSetValue (st2138::MultiSetValuePayload src, catena::excepti
                         index = path.back_as_index();
                         path.popBack();
                     }
+
```
ctwarog-ross: I like that it's in its own code block, so all the variables created are cleaned up after

abrown-RV:
Friendly reminder to take off draft as well when done.

ctwarog-ross (commented):
On the right track, just gotta finish the unit tests. Nice work

abrown-RV:
I'm concerned about the efficiency of this potentially high traffic endpoint in setting values. The fact that it might have to walk up a potentially deep nested param tree structure every call worries me. I think there are cases where params will be changing frequently and we want it to be *fast*. Basically where I'm going is that since the param structure cannot change, and the readonly also cannot change. I'm wondering if it would be better to:
- Walk the entire param tree during initialization and resolve all the readonlys then.
- Have a field on the Param/ParamDescriptor obj that lets you cache the answer so it doesn't have to keep checking.
- Modify getParam to track readonly as it walks down the param path so you don't have to walk back up it afterwards.

unittests/cpp/common/tests/Device_test.cpp
```diff
@@ -276,6 +283,87 @@ TEST_F(DeviceTest, Device_Create) {
...
+// add expect calls
+// child param is not mock
+
+// 1.0: Error Case - Fails to write to a child when parent struct ("product") is read-only
+TEST_F(DeviceTest, TryMultiSetValue_ReadOnlyStruct) {
+auto parentParam = std::make_shared<MockParam>();
+    auto parentInfo  = std::make_shared<MockParamDescriptor>();
+    auto& parentDesc = *parentInfo;
+
+    setupMockParam(*parentParam, "/product", parentDesc, /*isReadonly*/false, /*access*/0, adminScope_);
```
abrown-RV:
```suggestion
    setupMockParam(*parentParam, "/product", parentDesc, /*isArray*/false, /*size*/0, adminScope_);
```
```
/**
 * @brief Helper function to set up common mock parameter expectations
 * @param param The mock parameter to set up
 * @param oid The OID to return
 * @param descriptor The descriptor to return
 * @param isArray Whether this is an array type (defaults to false)
 * @param size The array size if it's an array type (defaults to 0)
 * @param scope The scope to return (defaults to monitor)
 */
inline void setupMockParam
```
4th param is `isArray` not `isReadonly`
5th param is `size` not `access`

unittests/cpp/common/tests/Device_test.cpp
```diff
@@ -276,6 +283,87 @@ TEST_F(DeviceTest, Device_Create) {
...
+
+    EXPECT_CALL(*parentParam, getDescriptor())
+        .WillRepeatedly(testing::ReturnRef(parentDesc));
+
+    // --- Child: /product/name (STRING, nominally writable)
+    auto childParam = std::make_shared<MockParam>();
+    auto childInfo  = std::make_shared<MockParamDescriptor>();
+    auto& childDesc = *childInfo;
+
+    setupMockParam(*childParam, "/product/name", childDesc, /*isReadonly*/false, /*access*/0, adminScope_);
```
abrown-RV: same comment as https://github.com/rossvideo/Catena/pull/1038/files#r2359640061

unittests/cpp/gRPC/tests/SetValue_test.cpp
```diff
@@ -138,3 +138,39 @@ TEST_F(gRPCSetValueTests, SetValue_Normal) {
...
+
+TEST_F(gRPCSetValueTests, SetValue_StructReadOnlyFail) {
+    // Start with a known "before" value for /structy_readonly/child
+    initPayload(0, "/struct_readonly/child", "new_value");
+    // Setting expectations.
+    EXPECT_CALL(dm0_, tryMultiSetValue(::testing::_, ::testing::_, ::testing::_)).Times(1)
+        .WillOnce(::testing::Invoke([this](st2138::MultiSetValuePayload src, catena::exception_with_status &ans, const IAuthorizer &authz) {
+            EXPECT_EQ(src.SerializeAsString(), expMultiVal_.SerializeAsString());
+            EXPECT_EQ(!authzEnabled_, &authz == &Authorizer::kAuthzDisabled);
+            return true;
```
abrown-RV: When attempting to set a readonly struct, the real `tryMultiSetValue` returns false, and sets the `ans` to an exception with `StatusCode::PERMISSION_DENIED`. The mock should replicate this, to make sure the controller handles it ok.

unittests/cpp/REST/tests/SetValue_test.cpp
```diff
@@ -131,3 +131,33 @@ TEST_F(RESTSetValueTests, SetValue_FailParse) {
...
+    initPayload(0, "/struct_readonly/child", "new_value");
+
+    // Setting expectations for the first call to set the value to "OriginalName"
+    EXPECT_CALL(dm0_, tryMultiSetValue(testing::_, testing::_, testing::_)).Times(1)
+        .WillOnce(testing::Invoke([this](st2138::MultiSetValuePayload src, catena::exception_with_status &ans, const IAuthorizer &authz) {
+            // Checking that function gets correct inputs.
+            auto val = (*src.mutable_values())[0];
+            EXPECT_EQ(val.oid(), fqoid_);
+            EXPECT_EQ(val.value().string_value(), inVal_.string_value());
+            return true;
```
abrown-RV: Same as https://github.com/rossvideo/Catena/pull/1038/files#r2359691114

abrown-RV (changes requested):
Couple mocking points in the unit tests.

Also what we talked about expanding the permutations of where readonly appears in the struct nesting.

Nice work figuring out mocking the nested structure in the first place! 👍

unittests/cpp/gRPC/tests/SetValue_test.cpp
```diff
@@ -137,4 +137,4 @@ TEST_F(gRPCSetValueTests, SetValue_Normal) {
     EXPECT_CALL(dm1_, commitMultiSetValue(::testing::_, ::testing::_)).Times(0);
     // Sending the RPC.
     testRPC();
-}
+}
```
ctwarog-ross: Undo this, we like the newline at the end of files (so the last line would be number 141 and it's blank)

unittests/cpp/REST/tests/SetValue_test.cpp
```diff
@@ -130,4 +130,4 @@ TEST_F(RESTSetValueTests, SetValue_FailParse) {
     EXPECT_CALL(dm0_, commitMultiSetValue(testing::_, testing::_)).Times(0);
     // Calling proceed and testing the output
     testCall();
-}
+}
```
ctwarog-ross: refer to other newline comment below

unittests/cpp/common/tests/ParamDescriptor_test.cpp
```diff
@@ -377,3 +378,78 @@ TEST_F(ParamDescriptorTest, ParamDescriptor_CommandErrUnhandled) {
...
     EXPECT_FALSE(response.has_response())                             << "Response should not have a response after an error.";
 }
+
+/*
+ * TEST 11 - Testing ParamDescriptor readOnly passed down from parent where readOnly() is true and false.
+ */
+
+TEST_F(ParamDescriptorTest, ParamDescriptor_ReadOnlyTrue) {
+    // Adding sub parameters.
+    MockParamDescriptor subPd1, subPd2;
```
ctwarog-ross: I don't think these should be mocks. Generaly mocking is used to mock the objects the object your testing interacts with during execution, not for the object itself. For example, you could remove your changes to readonly in the source and this test would still pass because you hardcoded the subparams readonly to return their parents' readonly. pd readonly true, create two ParamWithValues , one with readonly false, one with true and add them to pd, set pd readonly to return true. Then verify both params return the correct value. Create another seperate test where pd readonly is false, and repeat.

unittests/cpp/common/tests/ParamDescriptor_test.cpp
```diff
@@ -377,3 +378,78 @@ TEST_F(ParamDescriptorTest, ParamDescriptor_CommandErrUnhandled) {
...
+    EXPECT_EQ(pd->readOnly(), (subPd1.readOnly() && subPd2.readOnly()));
+}
+
+/*
+ * TEST 12 - Testing ParamDescriptor readOnly passed down multiple levels.
+ */
+
+TEST_F(ParamDescriptorTest, ParamDescriptor_ReadOnlyMultipleLevels) {
+    // Adding sub parameters.
+    MockParamDescriptor subPd, subsubPd;
```
ctwarog-ross: apply same logic as outlined above

ctwarog-ross (changes requested):

unittests/cpp/common/tests/ParamDescriptor_test.cpp
abrown-RV: For something simple like this, since its a pretty basic boolean logic, you can make a truth table for all possible outcomes and then test each of them

unittests/cpp/common/tests/ParamDescriptor_test.cpp
```diff
@@ -377,3 +409,53 @@ TEST_F(ParamDescriptorTest, ParamDescriptor_CommandErrUnhandled) {
     EXPECT_FALSE(responder->hasMore())                                << "Responder should not have any more responses after an error.";
     EXPECT_FALSE(response.has_response())                             << "Response should not have a response after an error.";
 }
+
+/*
+ * TEST 11 - Testing ParamDescriptor readOnly passed down from parent where readOnly() is true and false.
+ */
+
+    TEST_F(ParamDescriptorTest, ParamDescriptor_ReadOnlyTrue) {
```
abrown-RV: fix indentation

unittests/cpp/common/tests/ParamDescriptor_test.cpp
```diff
@@ -377,3 +409,53 @@ TEST_F(ParamDescriptorTest, ParamDescriptor_CommandErrUnhandled) {
...
+    auto subPd1 = createRealParamDescriptor(subOid1, "sub1", false, pd.get());
+    auto subPd2 = createRealParamDescriptor(subOid2, "sub2", false, pd.get());
+
+    // Test readOnly is true on children 
+    EXPECT_TRUE(pd->readOnly());
+    EXPECT_TRUE(subPd1->readOnly());  
+    EXPECT_TRUE(subPd2->readOnly());  
+
+    // Swap parent to readOnly = false
+    pd->readOnly(false);
```
ctwarog-ross, line 439: great stuff! Now seperate this 2nd part of the test into its own unique test. Each test is to test a different use case in an isolated environment

unittests/cpp/common/tests/ParamDescriptor_test.cpp
abrown-RV, line 1: I'd love another test that actually does use a mock (I know we just said there shouldn't be any mocks but bear with me) do a MockParam that is the parent of pd. Then you can `EXPECT_CALL(parent.readOnly()).times(0)` when the child is already readonly true to prove that the parent isn't checked if the child is already readonly.

abrown-RV (dismissed):
echo Christian's comment about breaking up the tests

ctwarog-ross (approved):

PR #1039, opened 2025-09-22
Added command oids to menu in codegen

### LINKED ISSUE #923: Command Oids aren't added to Menus
Add command oids to menus in the codegen and check that commands oids are sent with menus in the device request.

JChen2005:
Added command oids in to the menu in codegen.


abrown-RV (approved):
Nice work Jason!

PR #1045, opened 2025-09-23
Asset req test #820

### LINKED ISSUE #820: gRPC Unit Tests - ExternalObject(Asset?)Request
A unit test needs to be written for this, but it also doesn't match the REST naming.

nelgd:
Resolves Issue #820

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -263,6 +263,7 @@
         "unittests/gRPC/gRPC_ListLanguages_test",
         "unittests/gRPC/gRPC_MultiSetValue_test",
         "unittests/gRPC/gRPC_GetValue_test",
+        "unittests/gRPC/gRPC_ExternalObjectRequest_test",
         "unittests/REST/REST_GetValue_test",
         "unittests/REST/REST_SetValue_test",
         "unittests/REST/REST_ExecuteCommand_test",
--- a/sdks/cpp/connections/gRPC/src/controllers/ExternalObjectRequest.cpp
+++ b/sdks/cpp/connections/gRPC/src/controllers/ExternalObjectRequest.cpp
@@ -160,9 +160,10 @@ void ExternalObjectRequest::proceed(bool ok) {
             break;
 
         // Throws an error if the state is not recognized
-        default:
+        default: // GCOVR_EXCL_START
             status_ = CallStatus::kFinish;
             grpc::Status errorStatus(grpc::StatusCode::INTERNAL, "illegal state");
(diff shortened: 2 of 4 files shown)


nelgd:
Made some unit tests and ran coverage. Coverage is good, except for case where Call Status state is unrecognized. Feedback would be apreciated!
image

nelgd:
Changed and simplified the unit tests to match the ones for REST

abrown-RV:
> Made some unit tests and ran coverage. Coverage is good, except for case where Call Status state is unrecognized. Feedback would be apreciated! image

That's one of those shouldn't happen chunks of code. You could theoretically make the `status_` field `protected` instead of `private` on the parent class. Then you could extend the `ExternalObjectRequest` class in the unit tests to add a `setStatus` function to allow you to set the status to an invalid value and then call `proceed` to cover that default. There's examples on the protected member subclass thing in a few other units tests like Device and ConnectionQueue I think.

All that being said, that `default:` is there to stop compiler warnings complaining about switch statements with no default. Especially since its switching on an enum, it's pretty difficult to accidently get the wrong status in there. In other controllers, they've just excluded it from coverage, which I think is fine to do here as well:

```cpp
default: // GCOVR_EXCL_START
    status_ = CallStatus::kFinish;
    grpc::Status errorStatus(grpc::StatusCode::INTERNAL, "illegal state");
    writer_.FinishWithError(errorStatus, this);
    // GCOVR_EXCL_STOP
```

abrown-RV (changes requested):
Nice work Nelson. My only requested change I addressed in: https://github.com/rossvideo/Catena/pull/1045#issuecomment-3339920385

And the minor suggestion that you put the `cleanupTestFiles()` function into the test `tearDown` (it gets called after each `TEST_F`) so you don't have to manually put it at the end of every test.

nelgd:
> Nice work Nelson. My only requested change I addressed in: [#1045 (comment)](https://github.com/rossvideo/Catena/pull/1045#issuecomment-3339920385)
> 
> And the minor suggestion that you put the `cleanupTestFiles()` function into the test `tearDown` (it gets called after each `TEST_F`) so you don't have to manually put it at the end of every test.

Got that stuff implemented!

abrown-RV (approved):
👍

ctwarog-ross:
whats the coverage of the source file like now @nelgd ?

nelgd:
> whats the coverage of the source file like now @nelgd ?

It is 100%

PR #1047, opened 2025-09-23
985 Template sub params 

### LINKED ISSUE #985: Template sub params are inaccessible
Since GetParam relies heavily on the param's value to retrieve sub params, template params (initialized with an empty value) make it impossible to access children via GetParam.

This should be fixed so calls like /eq/fader (structs_with_authz example) do not return error.

JChen2005:
Now allows user to retrieve child param even when parent param has no values.


abrown-RV (changes requested):
This solution gives the struct a value. I believe this is breaking the intention of the device. If a param does not have a value we should not give it one. This is probably not a codegen problem, but a problem with the getParam function.

Potentially a discussion for tomorrow's checkin.

sdks/cpp/common/include/ParamWithValue.h
NR-RV, line 1: update the Copyright year and file meta data
line 4 and 33-40

sdks/cpp/common/src/Device.cpp
```diff
@@ -558,5 +588,4 @@ bool Device::shouldSendParam(const IParam& param, bool is_subscribed, const IAut
     }
 
     return should_send;
-}
-
+}
```
NR-RV, line 599: watch what your modifing
dont let the AI write to the full file

NR-RV (changes requested):

NR-RV (changes requested):
check things befor asking for review :)

sdks/cpp/common/src/Device.cpp
NR-RV, line 1: dont change files you didnt touch

tools/codegen/cpp/cppgen.js
NR-RV, line 1: dont touch files you didnt edit

unittests/cpp/common/tests/ParamWithValue_test/ParamWithValue_test.cpp
```diff
@@ -270,3 +260,17 @@ TEST_F(ParamWithValueTest, ParamInfoToProto_Error) {
     EXPECT_EQ(rc_.status, catena::StatusCode::PERMISSION_DENIED)
         << "toProto should return PERMISSION_DENIED if Authorizer does not have readAuthz.";
 }
+
+TEST_F(ParamWithValueTest, EmptyChild) {
```
NR-RV: bad test name and needs a comment to explain what it dose

unittests/cpp/common/tests/ParamWithValue_test/ParamWithValue_test.cpp
```diff
@@ -270,3 +260,17 @@ TEST_F(ParamWithValueTest, ParamInfoToProto_Error) {
...
+    NestedStructParam param(nestedValue, pd_);
+    Path path = Path("/f1/f2");
+    auto foundParam = param.getParam(path, authz_, rc_);
+    // Add verification that empty child is empty
+    EXPECT_EQ(rc_.status, catena::StatusCode::OK);
+    ASSERT_TRUE(foundParam) << "Did not find a parameter when one was expected";
+    // Verify the child field has the expected empty/default value
+    EXPECT_EQ(getParamValue<int32_t>(foundParam.get()), 0) 
+        << "Empty child field should have value 0";
+}
```
NR-RV, line 355: what happend to the other test you had for testing a param that has a child with a child?

unittests/cpp/common/CommonTestHelpers.h
NR-RV, line 1: update the meta data on line 31-36

unittests/cpp/common/tests/ParamWithValue_test/ParamWithValue_test.cpp
NR-RV, line 1: update meta data on line 31-37

sdks/cpp/common/include/ParamWithValue.h
NR-RV: and add your name to the authors and update the date

NR-RV (changes requested):

PR #1066, opened 2025-10-08
toolchain build fail due to old libssl-dev version no longer exists

ctwarog-ross:
updated libssl-dev version


abrown-RV (approved):
🤞 here's hoping!

PR #1074, opened 2025-10-09
1033 Scopes On Mandatory Params

### LINKED ISSUE #1033: mandatory param scope
Just a question about the mandatory params. What scope should they take on? For example in the asset_request_REST the default scope for the device is `st2138:op` meaning that since there is no scope on the `/product` param, the client needs that scope to even see the params. Which doesn't feel right to me.

Happy to chock this up to a problem for implementors and just close.

JChen2005:
now allows the codgen to check for required scope on mandatory params


abrown-RV (changes requested):
Allow for either `product` having the right scope or the individual params.

tools/codegen/index.js
abrown-RV: Should change the name and description of the param now that it is different.

tools/codegen/index.js
```diff
@@ -117,6 +120,17 @@ function areAllRequiredParamsPresent(deviceParams, disableMandatoryEnforcement =
                 return;
             }
 
+            // Validate the scope for mandatory parameters
+            const productScope = deviceParams.params.product.access_scope;
+            const globalScope = deviceParams.access_scopes || [];
```
abrown-RV: Check against `default_scope`

tools/codegen/index.js
```diff
@@ -89,22 +89,25 @@ function areAllRequiredParamsPresent(deviceParams, disableMandatoryEnforcement =
         "serial_number": true
     };
 
-    if (!deviceParams || !deviceParams.product) {
+    const REQUIRED_SCOPES = "st2138:mon";
```
abrown-RV: not a plural

tools/codegen/index.js
```diff
@@ -117,6 +120,17 @@ function areAllRequiredParamsPresent(deviceParams, disableMandatoryEnforcement =
                 return;
             }
 
+            // Validate the scope for mandatory parameters
+            const productScope = deviceParams.params.product.access_scope;
+            const globalScope = deviceParams.access_scopes || [];
+            const hasGlobalRequiredScope = globalScope.includes(REQUIRED_SCOPES);
+            
+            if (param.access_scope && param.access_scope !== REQUIRED_SCOPES) {
```
abrown-RV: I think it would be simpler to derive the scope for each param by combining the levels and then just a single if checking if its correct.

abrown-RV (changes requested):
still got a few more newlines to restore.

tools/codegen/index.js
```diff
@@ -136,7 +171,7 @@ function areAllRequiredParamsPresent(deviceParams, disableMandatoryEnforcement =
         }
     });
 
-    const allIssues = [...missing.map(p => `${p} (missing field)`), ...emptyValues];
+        const allIssues = [...missing.map(p => `${p} (missing field)`), ...emptyValues, ...invalidScopes];
```
abrown-RV:
```suggestion
    const allIssues = [...missing.map(p => `${p} (missing field)`), ...emptyValues, ...invalidScopes];
```

abrown-RV (changes requested):
just about there, just a minor indentation fix. There's still an unresolved comment from last round as well.

tools/codegen/index.js
```diff
@@ -117,6 +144,13 @@ function areAllRequiredParamsPresent(deviceParams, disableMandatoryEnforcement =
                 return;
             }
 
+            // Derive the effective scope and check if it's correct
```
NR-RV, line 137: check if this dose what you want it to do:

image

here is a nice truth table you can fill out the outcome line with what your code is doing and see if that is what you want it to be doing

NR-RV: here is an example of filling out the outcome (may not flow what your code dose)*:
image

tools/codegen/index.js
```diff
@@ -89,22 +90,48 @@ function areAllRequiredParamsPresent(deviceParams, disableMandatoryEnforcement =
...
+        
+        if (productScope) {
+            return productScope;
+        }
+        
+        if (defaultScope) {
+            return defaultScope;
+        }
+
+        else {
```
abrown-RV: Take this else out, its not needed

tools/codegen/index.js
JChen2005: I have fixed comments to include both param and scope.

tools/codegen/index.js
JChen2005: I fixed the indenting.

tools/codegen/index.js
JChen2005: Redid the entire if statement, gave the else statment a use as it returns "Invalid" when other possible senarios are not possible.

tools/codegen/index.js
JChen2005: Also changed param description to also reference the scopes.

abrown-RV: Change the name of the param, its not the device params anymore, its the whole model.

abrown-RV (approved):
Well done!

JChen2005:
Closes issue #1033 

Updated function areAllRequiredParamsPresent by renaming it to validateRequiredParamsAndScopes.

Updated function comments.

Created a nested fucntion getDerivedScope to check and validate scopes in devices.

PR #1076, opened 2025-10-10
gcovr 8.4 issues

abrown-RV:
gcovr has updated to 8.4 and caused a bunch of havoc. Trying to fix it:

- cmake now has a COVERAGE flag to turn coverage on or off. default is off, but the $CMAKE_COMMAND env var in the dev container sets it ON
- Some progress was made on the whole MASSIVE function count for ParamWithValue.h. gcovr is now set to only scan .o and .gcno files in the NON-example dirs, so it doesn't see all of the specializations for ParamWithValue<> with the types defined in examples.
- the gocvr args were getting out of hand so I switched to putting everything in gcovr.cfg which it reads by default.
- locked gcovr to 8.4 but moved the version number to a requirements.txt so dependabot can monitor it.
- added coverage gutters to recommended extensions
- I'm tired of copying toolchain-cpp.env every time so I updated the path in the dockerfile


NR-RV (dismissed):

gcovr.cfg
```diff
@@ -0,0 +1,17 @@
...
+delete-gcov-files = yes
+# combines line hits from multiple specializied template functions into a single line
+merge-lines = yes
+# search path controls where gcovr looks for object files
+# we need to point it where all the main code is built
+# but don't include any of the examples
+search-path = build/cpp/common/CMakeFiles
+search-path = build/cpp/connections/gRPC/CMakeFiles
+search-path = build/cpp/connections/REST/CMakeFiles
+search-path = build/cpp/unittests
```
ctwarog-ross: NEWLINE

ctwarog-ross (dismissed):
TOIGHT

gcovr.cfg
abrown-RV:
```suggestion
search-path = build/cpp/unittests

```

.devcontainer/toolchain-cpp.requirements.txt
```diff
@@ -0,0 +1 @@
+gcovr==8.4
```
abrown-RV:
```suggestion
gcovr==8.4

```

abrown-RV:
Looking into `set(COVERAGE ${COVERAGE} PARENT_SCOPE)` it will pass the value of `COVERAGE` up to the parent scope, as the name implies. I think the idea here is that if cmake is called without `-DCOVERAGE` then the `option(COVERAGE "Enable code coverage flags" OFF)` will create the var (either from cache or with the default) and then the `set` is required to pass it up to the parent scope. Since it is in a function, the value of `COVERAGE` would die at the end of the function.

For completeness if cmake is called with `-DCOVERAGE` then the var exists already in the top level scope and `option...set` do nothing.

https://cmake.org/cmake/help/latest/command/set.html

NR-RV (approved):

Issue #1016, opened 2025-09-11
Log rework


ctwarog-ross:
Clean up startup message
Fix loggging directory issues
Any other suggestions to be added by others

PR #1078, opened 2025-10-17
Changed logging libs from GLOG to ABSEIL

ctwarog-ross:
Closes Issue #1016

Now uses ABSL with custom FileSink that specifies file location UT always log to ABSL_LOGGING_DIR specified in cmake Examples use relative ./logs folder as default unless specified with new --logdir arg --silent still silences all logs except errors
Release mode only prints to logs
Debug prints to std and logs
Added unit test for log coverage


sdks/cpp/common/include/Logger.h
```diff
@@ -38,71 +38,80 @@
...
+    return logger;
+  }
+
+  Logger() = default;
+  Logger(const Logger&) = delete;
+  Logger& operator=(const Logger&) = delete;
+
   std::ostringstream stream_;
+  std::unique_ptr<FileLogSink> fileLogSink_ = nullptr;
 };
```
ctwarog-ross: newline, bad christian

sdks/cpp/connections/REST/examples/asset_request_REST/asset_request_REST.cpp
```diff
@@ -191,15 +191,10 @@ void RunRESTServer() {
...
-    absl::ParseCommandLine(argc, argv);
+    Logger::init(argc, argv, "asset_request_REST");
     
     std::thread catenaRestThread(RunRESTServer);
     catenaRestThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
ctwarog-ross: again, bad

sdks/cpp/connections/REST/examples/discovery_REST/discovery_REST.cpp
```diff
@@ -108,18 +108,13 @@ void RunRESTServer() {
...
     NmosNode node("Catena Device", "Catena Node", "A Catena example Node", "discovery_REST");
     node.init();
     
     std::thread catenaRestThread(RunRESTServer);
     catenaRestThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
ctwarog-ross: buddy

sdks/cpp/connections/REST/examples/structs_with_authz_JSON_REST/structs_with_authz_JSON_REST.cpp
```diff
@@ -145,15 +145,10 @@ void RunRESTServer() {
...
-    absl::ParseCommandLine(argc, argv);
+    Logger::init(argc, argv, "structs_with_authz_JSON_REST");
     
     std::thread catenaRestThread(RunRESTServer);
     catenaRestThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
ctwarog-ross: this one wasnt me

sdks/cpp/connections/REST/examples/structs_with_authz_REST/structs_with_authz_REST.cpp
```diff
@@ -145,15 +145,10 @@ void RunRESTServer() {
...
-    absl::SetProgramUsageMessage("Runs the Catena Service");
-    absl::ParseCommandLine(argc, argv);
-    
     std::thread catenaRestThread(RunRESTServer);
     catenaRestThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
ctwarog-ross: ^

sdks/cpp/common/include/Logger.h
```diff
@@ -38,71 +38,80 @@
...
+ * @brief Logger class for logging messages to both console and file.
+ */
+class Logger {
+public:
+  /*
+  * @brief Initialize the logger.
+  * @param appName The name of the application.
+  * @param logDir The directory where log files will be created.
+  */
+  static void init(const std::string& appName, const std::string& logDir = ABSL_LOG_DIR);
```
abrown-RV: suggest using `LOGGING_DIR` so if we change logging libraries in the future, we can reuse without having to go through all the cmake files again.

sdks/cpp/common/src/Logger.cpp
```diff
@@ -0,0 +1,86 @@
...
+        char datetime[32];
+        std::strftime(datetime, sizeof(datetime), "%Y%m%d_%H%M%S", &tm_buf);
+
+        std::string log_file = logDir + "/" + datetime + std::string("_") + appName + ".log";
+        instance().fileLogSink_ = std::make_unique<FileLogSink>(log_file);
+        absl::AddLogSink(instance().fileLogSink_.get());
+    });
+}
+
+void Logger::init(int argc, char** argv, const std::string& appName) {
```
abrown-RV: want to contemplate using the `ABSL_FLAG` system? since this is all abseil now? Benefit being I think that you'd no longer need to take `argc` and `argv` so you'd only need the one `init`. `ABSL_FLAG`s seem to be able to magically just read the args.

Maybe even add to `SharedFlags` to continue with the pattern?

sdks/cpp/common/src/Logger.cpp
```diff
@@ -0,0 +1,86 @@
...
+    static std::once_flag flag;
+    std::call_once(flag, [&]() {
+        absl::InitializeLog();
+
+        // Optionally set log dir via env or here
+        if (!std::filesystem::exists(logDir)) {
+            std::filesystem::create_directories(logDir);
+        }
+
+#ifdef NDEBUG
```
abrown-RV, line 51: I think we should give sdk users the option of logging to console in prod if they want. But maybe something to look at when we actually talk about making this something others might use. So feel free to ignore, we're already talked about logging to strout/err a bit to death XD

sdks/cpp/connections/REST/examples/use_commands_REST/use_commands_REST.cpp
```diff
@@ -184,18 +184,13 @@ void defineCommands() {
...
     // commands should be defined before starting the REST server 
     defineCommands();
     
     std::thread catenaRestThread(RunRESTServer);
     catenaRestThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
abrown-RV:
```suggestion
}

```

sdks/cpp/connections/REST/examples/use_menus_REST/use_menus_REST.cpp
```diff
@@ -152,15 +152,10 @@ void RunRESTServer() {
...
-    absl::ParseCommandLine(argc, argv);
+    Logger::init(argc, argv, "use_menus_REST");
     
     std::thread catenaRestThread(RunRESTServer);
     catenaRestThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
abrown-RV:
```suggestion
}

```

sdks/cpp/connections/gRPC/examples/one_of_everything/one_of_everything.cpp
```diff
@@ -362,7 +362,5 @@ int main(int argc, char* argv[])
     std::thread catenaRpcThread(RunRPCServer, addr);
     catenaRpcThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
abrown-RV:
```suggestion
}

```

sdks/cpp/connections/gRPC/examples/status_update/status_update.cpp
```diff
@@ -241,7 +241,5 @@ int main(int argc, char* argv[])
     std::thread catenaRpcThread(RunRPCServer, addr);
     catenaRpcThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
abrown-RV:
```suggestion
}

```

sdks/cpp/connections/gRPC/examples/structs_with_authz/structs_with_authz.cpp
```diff
@@ -179,7 +179,5 @@ int main(int argc, char* argv[])
     std::thread catenaRpcThread(RunRPCServer, addr);
     catenaRpcThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
abrown-RV:
```suggestion
}

```

sdks/cpp/connections/gRPC/examples/structs_with_authz_yaml/structs_with_authz_yaml.cpp
```diff
@@ -176,7 +176,5 @@ int main(int argc, char* argv[])
     std::thread catenaRpcThread(RunRPCServer, addr);
     catenaRpcThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
abrown-RV:
```suggestion
}

```

sdks/cpp/connections/gRPC/examples/use_commands/use_commands.cpp
```diff
@@ -279,7 +279,5 @@ int main(int argc, char* argv[])
     std::thread catenaRpcThread(RunRPCServer, addr);
     catenaRpcThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
abrown-RV:
```suggestion
}

```

sdks/cpp/connections/gRPC/examples/use_menus/use_menus.cpp
```diff
@@ -178,7 +178,5 @@ int main(int argc, char* argv[])
     std::thread catenaRpcThread(RunRPCServer, addr);
     catenaRpcThread.join();
     
-    // Shutdown Google Logging
-    google::ShutdownGoogleLogging();
     return 0;
 }
```
abrown-RV:
```suggestion
}

```

sdks/cpp/connections/REST/src/controllers/DeviceRequest.cpp
```diff
@@ -87,5 +87,5 @@ void DeviceRequest::proceed() {
 
     // Writing the final status to the console.
     writeConsole_(CallStatus::kFinish, socket_.is_open());
-    DEBUG_LOG << "DeviceRequest[" << objectId_ << "] finished\n";
+    LOG(INFO) << "DeviceRequest[" << objectId_ << "] finished\n";
 }
```
abrown-RV:
```suggestion
}

```

unittests/cpp/common/tests/ConnectionQueue_test.cpp
```diff
@@ -51,11 +51,10 @@ class ConnectionQueueTest : public testing::Test {
   protected:
     // Set up and tear down Google Logging
     static void SetUpTestSuite() {
-        Logger::StartLogging("ConnectionQueueTest");
+      Logger::init("ConnectionQueueTest", ABSL_LOG_DIR);
```
abrown-RV: Why does this on get `ABSL_LOG_DIR`?

Indentation

unittests/cpp/common/tests/Logger_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+            foundLogFile = true;
+            std::ifstream logFile(entry.path());
+            std::string logContent((std::istreambuf_iterator<char>(logFile)), std::istreambuf_iterator<char>());
+            EXPECT_NE(logContent.find("This is an info log message."), std::string::npos);
+            EXPECT_NE(logContent.find("This is a warning log message."), std::string::npos);
+            EXPECT_NE(logContent.find("This is an error log message."), std::string::npos);
+            break;
+        }
+    }
+}
```
abrown-RV:
```suggestion
}

```

unittests/cpp/common/tests/Logger_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    LOG(INFO) << "This is an info log message.";
+    LOG(WARNING) << "This is a warning log message.";
+    LOG(ERROR) << "This is an error log message.";
+    
+    //verify that the log file was created and contains the log messages
+    std::string logDir = ABSL_LOG_DIR;
+    std::string logFilePattern = "LoggerTest_"; // Log file name starts with this
+    bool foundLogFile = false;
+    for (const auto& entry : std::filesystem::directory_iterator(logDir)) {
```
abrown-RV: This test verification design makes me a little uncomfortable. As written, if for whatever reason there was a failing test in the logging dir. You'd have to go an manually delete it before tests would pass in the future. I think it would be better to clear out all `LoggerTest_` log files before calling `Logger::init`, or you could make this a fixtured test and then put it in the teardown (so it runs even if the test fails).

Another though I have is to not use `ABSL_LOG_DIR`, but instead a separate logging dir for these test(s). Since its a param in the init. Would make it easier to clean up logs from past test runs. What I like to do is make the directory and put a .gitignore file with a `*` in it in the dir. This works nicely because the .gitignore file will ignore all other files in the dir, and also gets around git's limitation of not being able to commit empty dirs by giving it a file in the dir to commit!

I do appreciate the desire to test the maybe more common and full path of if `init` uses the `ABSL_LOG_DIR` as the default. So since there's only one test possible, maybe ignore my last paragraph.

I do still think cleaning up test artifacts between tests if they will effect the pass/fail of the test to be important.

unittests/cpp/common/tests/Logger_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+        if (entry.is_regular_file() && entry.path().filename().string().find(logFilePattern) == 0) {
+            foundLogFile = true;
+            std::ifstream logFile(entry.path());
```
abrown-RV, line 80: Could also do `EXPECT_THAT(entry.path().filename().string(), MatchesRegex(pattern));` to assert the filename has the expected pattern with a formatted time, etc.

abrown-RV (changes requested):
Mostly newlines at the end of files if we're going to be picky about those.

Some actual thoughts as well on the real code.

sdks/cpp/common/src/Logger.cpp
ctwarog-ross: we need all the args because of the added functionality my wrapper provides. For ex, we parse a --silent arg

unittests/cpp/common/tests/Logger_test.cpp
ctwarog-ross: No the last paragraph is valid, making a folder in the ut dir for this is a smart idea, as well as cleaning up artifacts in the teardown.

.devcontainer/devcontainer.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/rossvideo/catena-toolchain-cpp:latest
+FROM testing:latest
```
abrown-RV:
```suggestion
FROM ghcr.io/rossvideo/catena-toolchain-cpp:latest
```

.devcontainer/toolchain-cpp.dockerfile
```diff
@@ -85,11 +85,9 @@ RUN . /root/toolchain.env \
     && apt-get update \
     && apt-get install -y docker-ce-cli docker-buildx-plugin docker-compose-plugin
 
-COPY .devcontainer/toolchain-cpp.requirements.txt /root/requirements.txt
-
 # Install Python and gcovr for coverage reports
 RUN apt-get update && apt-get install --no-install-recommends -y python3-pip \
-    && pip3 install --no-cache-dir -r /root/requirements.txt --break-system-packages \
+    && pip3 install --no-cache-dir gcovr --break-system-packages \
```
abrown-RV:
```suggestion
    && pip3 install --no-cache-dir -r /root/requirements.txt--break-system-packages \
```

.devcontainer/toolchain-cpp.env
```diff
@@ -25,3 +25,4 @@ LIBCURL_VERSION=8.5.0-2ubuntu10.6
 AVAHI_CLIENT_VERSION=0.8-13ubuntu6
 AVAHI_COMMON_VERSION=0.8-13ubuntu6
 PKG_CONFIG_VERSION=1.8.1-2build1
+ABSEIL_VERSION=4a2c633
```
abrown-RV: Points to [20240116.rc2](https://github.com/abseil/abseil-cpp/releases/tag/20240116.rc2). Probably want to not be on a Release Candidate, but a full release?

Suggest putting the version tag in a comment as well since commit hashes are so opaque. 
Since you're running git checkout in the submodule it doesn't have to be a commit hash, could be the tag.

sdks/cpp/cmake/ProjectConfig.cmake
```diff
@@ -21,24 +21,15 @@ function(setup_project_directories)
 endfunction()
 
 # Setup logging configuration  
-function(setup_logging_config CATENA_ROOT_DIR)
+function(setup_logging_config)
     # Set default log directory - respect command line override
```
abrown-RV, line 25: Do we want to have a default `LOG_DIR`? What happens if it's not set, does `add_compile_definitions` still work? I guess it would inject `LOG_DIR=""`. Which the code probably is fine with?

sdks/cpp/common/src/Logger.cpp
```diff
@@ -0,0 +1,76 @@
...
+// common
+#include <Logger.h>
+#include <SharedFlags.h>
+
+void Logger::init(const std::string& appName) {
+    static std::once_flag flag;
+    std::call_once(flag, [&]() {
+        absl::InitializeLog();
+
+        auto test = absl::GetFlag(FLAGS_log_dir);
```
abrown-RV: unused? Guessing it was a gdb watch?

sdks/cpp/common/src/Logger.cpp
```diff
@@ -0,0 +1,76 @@
...
+        // Optionally set log dir via env or here
```
abrown-RV:
```suggestion
        // Create the log dir if it doesn't exist
```

sdks/cpp/connections/gRPC/examples/audiodeck/audiodeck.cpp
```diff
@@ -168,18 +168,14 @@ void RunRPCServer(std::string addr)
 
 int main(int argc, char* argv[])
 {
-    Logger::StartLogging(argc, argv);
-
     std::string addr;
-    absl::SetProgramUsageMessage("Runs the Catena Service");
     absl::ParseCommandLine(argc, argv);
+    Logger::init("structs_with_authz_REST");
```
abrown-RV:
```suggestion
    Logger::init("audiodeck");
```

sdks/cpp/connections/gRPC/examples/audiodeck_yaml/audiodeck_yaml.cpp
```diff
@@ -165,18 +165,14 @@ void RunRPCServer(std::string addr)
 
 int main(int argc, char* argv[])
 {
-    Logger::StartLogging(argc, argv);
-
     std::string addr;
-    absl::SetProgramUsageMessage("Runs the Catena Service");
     absl::ParseCommandLine(argc, argv);
-  
+    Logger::init("structs_with_authz_REST");
```
abrown-RV:
```suggestion
    Logger::init("audiodeck_yaml");
```

sdks/cpp/connections/gRPC/examples/audiodeck/audiodeck.cpp
```diff
@@ -168,18 +168,14 @@ void RunRPCServer(std::string addr)
 
 int main(int argc, char* argv[])
 {
-    Logger::StartLogging(argc, argv);
-
     std::string addr;
-    absl::SetProgramUsageMessage("Runs the Catena Service");
```
abrown-RV, line 174: Why remove?

sdks/cpp/connections/REST/examples/audiodeck_JSON_REST/audiodeck_JSON_REST.cpp
```diff
@@ -145,15 +145,11 @@ void RunRESTServer() {
 }
 
 int main(int argc, char* argv[]) {
-    Logger::StartLogging(argc, argv);
-
-    absl::SetProgramUsageMessage("Runs the Catena Service");
     absl::ParseCommandLine(argc, argv);
-    
+    Logger::init("structs_with_authz_JSON_REST");
```
abrown-RV:
```suggestion
    Logger::init("audiodeck_JSON_REST");
```

sdks/cpp/connections/REST/examples/audiodeck_REST/audiodeck_REST.cpp
```diff
@@ -145,15 +145,11 @@ void RunRESTServer() {
 }
 
 int main(int argc, char* argv[]) {
-    Logger::StartLogging(argc, argv);
-
-    absl::SetProgramUsageMessage("Runs the Catena Service");
     absl::ParseCommandLine(argc, argv);
-    
+    Logger::init("structs_with_authz_REST");
```
abrown-RV:
```suggestion
    Logger::init("audiodeck_REST");
```

unittests/cpp/common/tests/Logger_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+// gtest
+#include <gtest/gtest.h>
+
+using namespace catena::common;
+
+class LoggerTest : public ::testing::Test {
+    void SetUp() override {
+        absl::SetFlag(&FLAGS_log_dir, UNITTEST_LOG_DIR + std::string("/logger"));
+        Logger::init("LoggerTest");
```
abrown-RV, line 58: I realized I misspoke last round, I actually intentionally like to do the clean in the setup, that way, if there's a test that doesn't pass, the artifacts are left behind to poke around at them.

unittests/cpp/common/tests/Logger_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+    LOG(INFO) << "This is an info log message.";
+    LOG(WARNING) << "This is a warning log message.";
+    LOG(ERROR) << "This is an error log message.";
+    
+    //verify that the log file was created and contains the log messages
+    std::string logFilePattern = "_LoggerTest.log"; // Log file name starts with this
+    bool foundLogFile = false;
+    for (const auto& entry : std::filesystem::directory_iterator(UNITTEST_LOG_DIR + std::string("/logger"))) {
+        //get the file that ends with _LoggerTest
+        if (entry.is_regular_file() && entry.path().filename().string().find(logFilePattern) == entry.path().filename().string().length() - logFilePattern.length()) {
```
abrown-RV, line 78: Now that you're in an empty subdir for this test, don't just skip non matching files. Any extra files are a failed test, cause it should only make the one. I think you still need the loop, since you don't concretely know the name of the log file cause of the time, but the loop can just expect the regex pattern, and then also expect the count to be one. Unless there's a fancy iter way to just get the first element out of the iter without a loop. Then just do that and then expect the iter is at the end.

unittests/cpp/gRPC/GRPCTest.h
```diff
@@ -58,6 +58,7 @@
 // common
 #include <Status.h>
 #include <Logger.h>
+#include <SharedFlags.h>
```
abrown-RV, line 61:
```suggestion
```

unittests/cpp/common/tests/Logger_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+            foundLogFile = true;
+            std::ifstream logFile(entry.path());
+            std::string logContent((std::istreambuf_iterator<char>(logFile)), std::istreambuf_iterator<char>());
+            EXPECT_NE(logContent.find("This is an info log message."), std::string::npos);
+            EXPECT_NE(logContent.find("This is a warning log message."), std::string::npos);
```
abrown-RV, line 83: Check for error log message.

abrown-RV (commented):
Super cool `absl::SetFlag` exists. Great for testing!

.devcontainer/devcontainer.dockerfile
ctwarog-ross: doh!

.devcontainer/toolchain-cpp.env
ctwarog-ross: I tried the latest release, but it wouldn't build due to incompatibility with grpc. I ended up choosing this version as thats the hash that the version of grpc we use has checked out

ctwarog-ross: so essentially we're hardcoded locked into this combination of grpc and abseil

sdks/cpp/cmake/ProjectConfig.cmake
ctwarog-ross: the log dir is passed in cmd args in the dockerfile. My thought was if someone wants to go build it their own way, they can change it there. I will add that iif they don't, the default location is set to Catena/logs in the cmake files anyway.

unittests/cpp/gRPC/GRPCTest.h
ctwarog-ross: it needs this to be able to set the unit test dir flag

unittests/cpp/common/tests/Logger_test.cpp
ctwarog-ross: I could also just check if it has more than 1 file?

sdks/cpp/connections/gRPC/examples/audiodeck/audiodeck.cpp
ctwarog-ross: doesnt add anything to the log

.devcontainer/toolchain-cpp.env
abrown-RV: gotcha, can you lock it to the version tag instead of the hash for readability?

sdks/cpp/cmake/ProjectConfig.cmake
abrown-RV: I'm only familiar with using the `option` cmake directive, is there another way?

unittests/cpp/common/tests/Logger_test.cpp
abrown-RV: totally

unittests/cpp/gRPC/GRPCTest.h
abrown-RV: Where? this file doesn't touch the flag. All the files that include GRPCTest.h `#include <SharedFlags.h>` themselves.

sdks/cpp/connections/gRPC/examples/audiodeck/audiodeck.cpp
abrown-RV: Its not for logging. Its part of the absl flags system. It prints that message when the user runs the example with `--help`

unittests/cpp/gRPC/GRPCTest.h
ctwarog-ross: gotchu, I must haver forgotten to remove it then.

ctwarog-ross: nope, just looked through and none of the files that include grpctest.h include sharedflags. Leaving as is

ctwarog-ross:
BLOCKED by https://github.com/rossvideo/Catena/pull/1090

abrown-RV (approved):

Issue #1003, opened 2025-09-07
codegen fails if the param_oids or command_oids fields are missing


mejohnnaylor:
line 216, codegen.js

```js
        let paramOids = menus[menu].param_oids.map(oid => `"${oid}"`).join(", ");
```

I've got a dm with menus that don't have command or param oids defined which makes the `param_oids` object `undefined`.

Fix - either we change the schema to make the fields `required`, or we permit them to be missing, and the code tests whether they're defined before proceeding.

I could go either way.

nelgd:
Has codegen changed since this issue was made? I cant seem to find the right files

abrown-RV:
Closed #923

PR #1079, opened 2025-10-17
Handling No Params and Commands (#1003)

nelgd:
If no command or params returns empty list instead of object. Closes issue #1003


abrown-RV:
Actually fixed in https://github.com/rossvideo/Catena/issues/923 already. But code looks nice and readable.

PR #1089, opened 2025-10-23
Connect bug #943

### LINKED ISSUE #943: REST::Connect needs a PushUpdate to realize a connection has been cancelled
This is an old bug that has been fixed in the gRPC implementation. REST, or at least Boost, does not seem to have a good way of detecting when the client has been disconnected. Further research needs to be done as to a solution.

nelgd:
Closes issue #943 

Gave fixing this bug a shot based off what was done for the gRPC fix

Added a new unit test to check for the disconnect update and all other unit tests still pass

### DIFF
--- a/sdks/cpp/connections/REST/src/ServiceImpl.cpp
+++ b/sdks/cpp/connections/REST/src/ServiceImpl.cpp
@@ -45,7 +45,7 @@ ServiceImpl::ServiceImpl(const ServiceConfig& config)
       EOPath_{config.EOPath},
       port_{config.port},
       authorizationEnabled_{config.authz},
-      acceptor_{io_context_, tcp::endpoint(tcp::v4(), config.port)},
+      acceptor_{io_context_},
       router_{Router::getInstance()},
       connectionQueue_{config.maxConnections} {
 
--- a/sdks/cpp/connections/REST/src/ServiceImpl.cpp
+++ b/sdks/cpp/connections/REST/src/ServiceImpl.cpp
@@ -19,11 +19,6 @@ using catena::REST::ServiceImpl;
 
 using catena::REST::Connect;
 
-#include <sys/types.h>
-#include <sys/socket.h>
-#include <netinet/tcp.h>
-#include <netinet/in.h>
-
 // Defining the port flag from SharedFlags.h
 ABSL_FLAG(uint16_t, port, 443, "Catena REST service port");
 
@@ -95,7 +90,7 @@ void ServiceImpl::run() {
         // Waiting for a connection.
         tcp::socket socket(io_context_);
         acceptor_.accept(socket);
-    
+         // Once a connection is made, increment activeRequests and handle async.
         {
             std::lock_guard<std::mutex> lock(activeRequestMutex_);
             activeRequests_ += 1;
--- a/sdks/cpp/connections/REST/src/ServiceImpl.cpp
+++ b/sdks/cpp/connections/REST/src/ServiceImpl.cpp
@@ -23,7 +23,6 @@ using catena::REST::Connect;
 #include <sys/socket.h>
 #include <netinet/tcp.h>
 #include <netinet/in.h>
-#endif
 
 // Defining the port flag from SharedFlags.h
 ABSL_FLAG(uint16_t, port, 443, "Catena REST service port");
@@ -99,7 +98,6 @@ void ServiceImpl::run() {
         // Enable TCP keepalive to detect half-open connections
         try {
             socket.set_option(boost::asio::socket_base::keep_alive(true));
-#if !defined(_WIN32)
             int yes = 1;
             int idle = 30;     // seconds before sending keepalive probes
             int intvl = 10;    // interval between keepalive probes
@@ -108,7 +106,6 @@ void ServiceImpl::run() {
             ::setsockopt(socket.native_handle(), IPPROTO_TCP, TCP_KEEPIDLE, &idle, sizeof(idle));
             ::setsockopt(socket.native_handle(), IPPROTO_TCP, TCP_KEEPINTVL, &intvl, sizeof(intvl));
             ::setsockopt(socket.native_handle(), IPPROTO_TCP, TCP_KEEPCNT, &cnt, sizeof(cnt));
-#endif
         } catch (...) { /* best-effort keepalive */ }
         // Once a connection is made, increment activeRequests and handle async.
         {
--- a/sdks/cpp/connections/REST/src/ServiceImpl.cpp
+++ b/sdks/cpp/connections/REST/src/ServiceImpl.cpp
@@ -19,6 +19,13 @@ using catena::REST::ServiceImpl;
 
 using catena::REST::Connect;
 
+#if !defined(_WIN32)
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/tcp.h>
+#include <netinet/in.h>
+#endif
+
 // Defining the port flag from SharedFlags.h
 ABSL_FLAG(uint16_t, port, 443, "Catena REST service port");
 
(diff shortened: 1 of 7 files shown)


sdks/cpp/connections/REST/src/ServiceImpl.cpp
```diff
@@ -19,6 +19,13 @@ using catena::REST::ServiceImpl;
 
 using catena::REST::Connect;
 
+#if !defined(_WIN32)
```
abrown-RV: we don't support windows, so I think just take out WIN32 gates

abrown-RV (changes requested):
If I'm understanding correctly, this will change the sendResponse to gracefully close the connection if it tries to send and the other side has hung up. And then sets up a heartbeat to wake up and send on an interval so that the failure occurs and then gracefully shuts down the connection.

So we already have a heartbeat, so I think the resolution here is to keep the graceful shutdown part and then just use the existing heartbeat. Your unit test for this would just be to send a regular param update where the client has hung up, and check that it gracefully closes.

nelgd:
> If I'm understanding correctly, this will change the sendResponse to gracefully close the connection if it tries to send and the other side has hung up. And then sets up a heartbeat to wake up and send on an interval so that the failure occurs and then gracefully shuts down the connection.
> 
> So we already have a heartbeat, so I think the resolution here is to keep the graceful shutdown part and then just use the existing heartbeat. Your unit test for this would just be to send a regular param update where the client has hung up, and check that it gracefully closes.

Got all that done. I had an issue where coverage was passing locally but not on the repo. 

Would you be able to explain what this code that Chat got me to add to `ServiceImpl.cpp` does to fix the problem?

```
// Open and bind the acceptor with reuse_address to reduce flakiness on CI
    boost::system::error_code ec;
    acceptor_.open(tcp::v4(), ec);
    if (ec) {
        throw std::runtime_error(std::string("Failed to open acceptor: ") + ec.message());
    }
    acceptor_.set_option(boost::asio::socket_base::reuse_address(true), ec);
    // Ignore ec from set_option; continue to bind
    acceptor_.bind(tcp::endpoint(tcp::v4(), config.port), ec);
    if (ec) {
        throw std::runtime_error(std::string("Failed to bind port ") + std::to_string(config.port) + ": " + ec.message());
    }
    acceptor_.listen(boost::asio::socket_base::max_listen_connections, ec);
    if (ec) {
        throw std::runtime_error(std::string("Failed to listen on port ") + std::to_string(config.port) + ": " + ec.message());
    }
 ```

File: sdks/cpp/connections/REST/src/ServiceImpl.cpp [hunk 2]
```diff
@@ -90,6 +112,18 @@ void ServiceImpl::run() {
...
+            socket.set_option(boost::asio::socket_base::keep_alive(true));
+            int yes = 1;
+            int idle = 30;     // seconds before sending keepalive probes
+            int intvl = 10;    // interval between keepalive probes
+            int cnt = 3;       // probes before declaring dead
+            ::setsockopt(socket.native_handle(), SOL_SOCKET, SO_KEEPALIVE, &yes, sizeof(yes));
+            ::setsockopt(socket.native_handle(), IPPROTO_TCP, TCP_KEEPIDLE, &idle, sizeof(idle));
+            ::setsockopt(socket.native_handle(), IPPROTO_TCP, TCP_KEEPINTVL, &intvl, sizeof(intvl));
+            ::setsockopt(socket.native_handle(), IPPROTO_TCP, TCP_KEEPCNT, &cnt, sizeof(cnt));
+        } catch (...) { /* best-effort keepalive */ }
```
abrown-RV: Gonna say take all this out as well. We don't need multiple keepalives, and I misspoke, gRPC does not have a transport level keepalive either. It also has the ability, we've just elected to use our application level keepalive.

sdks/cpp/connections/REST/src/SocketWriter.cpp
```diff
@@ -48,7 +48,12 @@ void SocketWriter::sendResponse(const catena::exception_with_status& err, const
                  << "Access-Control-Allow-Headers: Content-Type, Authorization, accept, Origin, X-Requested-With, Language, Detail-Level\r\n"
                  << "Access-Control-Allow-Credentials: true\r\n\r\n"
                  << jsonBody_;
-        boost::asio::write(socket_, boost::asio::buffer(response.str()));
+        // Use non-throwing write; on error, close socket to signal disconnect
+        boost::system::error_code ec;
+        boost::asio::write(socket_, boost::asio::buffer(response.str()), ec);
+        if (ec) {
+            socket_.close();
```
abrown-RV, line 57: log the error at warning level: `LOG(WARNING) << message`

sdks/cpp/connections/REST/src/SocketWriter.cpp
```diff
@@ -85,10 +90,15 @@ void SSEWriter::sendResponse(const catena::exception_with_status& err, const goo
...
     if (httpStatus.first < 300 && !jsonOutput.empty()) {
         response << "data: " << jsonOutput << "\n\n";
     }
 
-    boost::asio::write(socket_, boost::asio::buffer(response.str()));
+    // Use non-throwing write; on error, close socket to signal disconnect
+    boost::system::error_code ec;
+    boost::asio::write(socket_, boost::asio::buffer(response.str()), ec);
+    if (ec) {
+        socket_.close();
```
abrown-RV, line 105: same, log it please

abrown-RV (changes requested):
I'd also like to see a unit test (if possible) that connects sends a message, disconnects the client, and then send another message that should trigger the graceful shutdown.

Nice work!

sdks/cpp/connections/REST/src/ServiceImpl.cpp
```diff
@@ -19,6 +19,11 @@ using catena::REST::ServiceImpl;
 
 using catena::REST::Connect;
 
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/tcp.h>
+#include <netinet/in.h>
+
```
abrown-RV: I think these are also no longer necessary without all the `setopts`?

sdks/cpp/connections/REST/src/ServiceImpl.cpp
```diff
@@ -90,7 +95,7 @@ void ServiceImpl::run() {
         // Waiting for a connection.
         tcp::socket socket(io_context_);
         acceptor_.accept(socket);
-        // Once a connection is made, increment activeRequests and handle async.
+
```
abrown-RV:
```suggestion
        // Once a connection is made, increment activeRequests and handle async.
```

sdks/cpp/connections/REST/src/SocketWriter.cpp
```diff
@@ -1,4 +1,5 @@
 #include <SocketWriter.h>
+#include <absl/log/log.h>
```
abrown-RV: We actually have a `Logger.h` that abstracts absl that you should include instead.

unittests/cpp/REST/tests/ServiceImpl_test.cpp
```diff
@@ -53,6 +53,12 @@
 using namespace catena::common;
 using namespace catena::REST;
 
+#include <atomic>
+
+namespace {
+    std::atomic<uint16_t> s_next_port{50050};
```
abrown-RV: Just make this a field on the `RESTServiceImplTests` class

unittests/cpp/REST/tests/SocketWriter_test.cpp
```diff
@@ -310,3 +309,41 @@ TEST_F(RESTSocketWriterTests, SSEWriter_WriteErrEnd) {
...
+    google::protobuf::util::JsonPrintOptions options;
+    auto status = google::protobuf::util::MessageToJsonString(msg1, &jsonBody1, options);
+    EXPECT_EQ(readResponse(), expectedResponse(rc, jsonBody1));
+
+    // Shutdown and close client socket to simulate unexpected client disconnect.
+    boost::system::error_code ec;
+    clientSocket_.shutdown(tcp::socket::shutdown_both, ec);
+    clientSocket_.close();
+
+    // Writing multiple messages to ensure the error is triggered.
```
abrown-RV: Why multiple? shouldn't it just always die right away? Or is there some delay after calling `clientSocket_.close()` before it actually closes?

abrown-RV (changes requested):
Very nice Nelson, just a little bit of cleanup, and a question.

sdks/cpp/connections/REST/src/ServiceImpl.cpp
nelgd: Yeah I thought so also. I removed them and then the build failed on github so I reverted it

abrown-RV: built fine for me locally, if it's failing on github, that's an issue worth investigating

nelgd: ill give it another shot and see what happens

unittests/cpp/REST/tests/SocketWriter_test.cpp
nelgd: Yeah, multiple is unneeded. Changed it just to one

nelgd: Had issues with just one. I asked chat to explain why multiple might be needed. Is this valid reasoning?
image

sdks/cpp/connections/REST/src/ServiceImpl.cpp
abrown-RV: I think this comment is still valid, especially since there's no other changes in the file now.

unittests/cpp/REST/tests/ServiceImpl_test.cpp
abrown-RV, line 1: Add yourself to `@author` here too

abrown-RV (changes requested):
Lookin good just a couple little fixes

unittests/cpp/REST/tests/ServiceImpl_test.cpp
nelgd: Added myself and added back the comment

abrown-RV (approved):
nice work! 👍

PR #1112, opened 2025-11-20
Case insensitive headers (#1105)

### LINKED ISSUE #1105: case insensitive header
Proper HTTP spec says that headers need to be case *insensitive*. Currently our header checks (or at least the `Detail-Level` header on the device request) only work with the upper cases D and L. All endpoints should be changed to allow for either, and unit tests need to be updated to check that the headers are case insensitive.

nelgd:
Headers for HTTP are now case insensitive

Added a unit test

### DIFF
--- a/sdks/cpp/connections/REST/src/SocketReader.cpp
+++ b/sdks/cpp/connections/REST/src/SocketReader.cpp
@@ -1,6 +1,5 @@
 
 #include <SocketReader.h>
-#include <algorithm>
 #include <cctype>
 using catena::REST::SocketReader;
 
--- a/sdks/cpp/connections/REST/src/SocketReader.cpp
+++ b/sdks/cpp/connections/REST/src/SocketReader.cpp
@@ -1,5 +1,7 @@
 
 #include <SocketReader.h>
+#include <algorithm>
+#include <cctype>
 using catena::REST::SocketReader;
 
 void SocketReader::read(tcp::socket& socket) {
@@ -77,28 +79,47 @@ void SocketReader::read(tcp::socket& socket) {
     // Looping through headers to retrieve JWS token and json body len.
     std::size_t contentLength = 0;
     while(std::getline(header_stream, header) && header != "\r") {
+        // Parse "<name>: <value>" and handle header name case-insensitively
+        std::size_t sep = header.find(':');
+        if (sep == std::string::npos) {
+            continue;
+        }
+        std::string nameLower = header.substr(0, sep);
+        std::transform(nameLower.begin(), nameLower.end(), nameLower.begin(),
+                       [](unsigned char c){ return static_cast<char>(std::tolower(c)); });
+        std::string value = header.substr(sep + 1);
+        // Trim leading spaces
+        while (!value.empty() && (value.front() == ' ' || value.front() == '\t')) {
+            value.erase(0, 1);
+        }
+        // Trim trailing CR and spaces
+        while (!value.empty() && (value.back() == '\r' || value.back() == ' ' || value.back() == '\t')) {
+            value.pop_back();
+        }
+
         // Getting jwsToken.
-        if (service_->authorizationEnabled() && jwsToken_.empty() && header.starts_with("Authorization: Bearer ")) {
-            jwsToken_ = header.substr(std::string("Authorization: Bearer ").length());
-            // Removing newline.
-            jwsToken_.erase(jwsToken_.length() - 1);
+        if (service_->authorizationEnabled() && jwsToken_.empty() && nameLower == "authorization") {
+            // Expect "Bearer <token>" (keep scheme check case-sensitive as before)
+            static const std::string kBearerPrefix = "Bearer ";
+            if (value.rfind(kBearerPrefix, 0) == 0) {
+                jwsToken_ = value.substr(kBearerPrefix.length());
+            }
         }
         // Getting origin
-        else if (origin_.empty() && header.starts_with("Origin: ")) {
-            origin_ = header.substr(std::string("Origin: ").length());
+        else if (origin_.empty() && nameLower == "origin") {
+            origin_ = value;
         }
         // Getting detail level from header
-        else if (detailLevel_ == st2138::Device_DetailLevel_UNSET && header.starts_with("Detail-Level: ")) {
-            std::string dl = header.substr(std::string("Detail-Level: ").length());
-            dl.erase(dl.length() - 1); // Removing newline
+        else if (detailLevel_ == st2138::Device_DetailLevel_UNSET && nameLower == "detail-level") {
+            std::string dl = value;
             auto& dlMap = catena::common::DetailLevel().getReverseMap();
             if (dlMap.contains(dl)) {
                 detailLevel_ = dlMap.at(dl);
             }
         }
         // Getting body content-Length
-        else if (contentLength == 0 && header.starts_with("Content-Length: ")) {
-            contentLength = stoi(header.substr(std::string("Content-Length: ").length()));
+        else if (contentLength == 0 && nameLower == "content-length") {
+            contentLength = stoi(value);
         }
     }
     // If body exists, we need to handle leftover data and append the rest.
--- a/sdks/cpp/connections/REST/src/SocketReader.cpp
+++ b/sdks/cpp/connections/REST/src/SocketReader.cpp
@@ -124,7 +124,7 @@ void SocketReader::read(tcp::socket& socket) {
         if (service_->authorizationEnabled() && jwsToken_.empty() && iequals_header_name(name, "authorization")) {
             // Expect "Bearer <token>" (keep scheme check case-sensitive as before)
             static const std::string kBearerPrefix = "Bearer ";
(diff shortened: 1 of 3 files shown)


sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -1,5 +1,7 @@
 
 #include <SocketReader.h>
+#include <algorithm>
```
abrown-RV: not needed?

sdks/cpp/connections/REST/src/SocketReader.cpp
abrown-RV, line 1: Because this is a high traffic section of code (every rest endpoint call starts here) I'm looking at it pretty critically from a performance perspective. I'm uncomfortable with the overhead of converting *every* header to lowercase, even when its the not right header. I would prefer a case-insensitive equals to pre converting every header name. Here's a specific equals function AI wrote us that should work, although maybe not XD. There are efficiancies to be gained over the `std::tolower` function since we don't need to do any locality, the header names should just be basic ascii. Also the b input is what we're testing against, so you already know it will be lowercase.
```cpp
inline bool iequals_header_name(std::string_view a, std::string_view b) {
    if (a.size() != b.size()) return false;

    const unsigned char* pa = reinterpret_cast<const unsigned char*>(a.data());
    const unsigned char* pb = reinterpret_cast<const unsigned char*>(b.data());
    const std::size_t n = a.size();

    for (std::size_t i = 0; i < n; ++i) {
        unsigned char ca = pa[i];
        unsigned char cb = pb[i];  // already lowercase

        if (ca == cb) continue;    // fast path

        // Fold only 'A'..'Z' to 'a'..'z'
        if (ca >= 'A' && ca <= 'Z') {
            ca = static_cast<unsigned char>(ca | 0x20);  // to lowercase
            if (ca == cb) continue;
        }

        // If we get here, they are not equal
        return false;
    }
    return true;
}
```

unittests/cpp/REST/RESTTest.h
abrown-RV, line 1: This isn't you're fault, but now that I'm looking at this file I see that the headers are actually being constructed incorrectly. HTTP spec says that every header is separated by CRLF, but our test code is just doing LF. Now that I've noticed, can you fix it?

unittests/cpp/REST/RESTTest.h
```diff
@@ -147,6 +147,69 @@ class RESTTest {
         boost::asio::write(*writeSocket_, boost::asio::buffer(request));
     }
```
abrown-RV, line 126: Would it be possible to chain `writeRequest` to `writeRequestWithHeaderNames` for less code duplication? something like this:
```cpp
void writeRequest(REST Method method,
    uint32_t slot,
    const std::string& endpoint
    // and so on
{
    writeRequestWithHeaderNames(method, slot, endpoint, fqoid, stream, fields, jwsToken, origin, detailLevel, language, jsonBody, "Origin", "Authorization", "Detail-Level", "Language", "Detail-Level");
}
```

abrown-RV (changes requested):

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -278,3 +278,45 @@ TEST_F(RESTSocketReaderTests, SocketReader_LongJsonBody) {
 TEST_F(RESTSocketReaderTests, SocketReader_HeadersUnset) {
     testCall(catena::REST::Method_GET, 1, "/test-call", "/test/oid", false, {}, false, "", "*", st2138::Device_DetailLevel_UNSET, "", "");
 }
+
+/*
+ * TEST 15 - Headers are case-insensitive (per HTTP spec).
+ */
+TEST_F(RESTSocketReaderTests, SocketReader_HeaderCaseInsensitive) {
+    // Authz false by default.
+    EXPECT_CALL(service_, authorizationEnabled()).WillRepeatedly(testing::Return(false));
```
abrown-RV: should make this return true so it parses the auth header and you can assert the jwt after the call. With this false you aren't actually proving that `AUTHORIZATION` works.

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -1,7 +1,34 @@
 
 #include <SocketReader.h>
+#include <string_view>
 using catena::REST::SocketReader;
 
+namespace {
```
abrown-RV: just make this a private static function on SocketReader

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -1,7 +1,34 @@
...
+        if (ca == cb) continue; // fast path
+
+        // Fold only 'A'..'Z' to 'a'..'z'
+        if (ca >= 'A' && ca <= 'Z') {
+            ca = static_cast<unsigned char>(ca | 0x20); // to lowercase
+            if (ca == cb) continue;
+        }
+
+        // If we get here, they are not equal
+        return false;
```
abrown-RV, line 26: unit tests never hit this line because all the headers that return false return on the size check. So there needs to be a header with the same length as one of the checked headers.....


You could change `writeRequestWithHeaderNamess` to be `writeRequestWithHeaders` and take a `std::map<std:string, std::string> headers` param. and then the `writeRequest` chain would have `writeRequestWithHeaders(..., {{"Origin", origin}, {"Detail-Level", detailLevel}, etc.});`

That way in your new test you can add a random extra header that will be ignored, but has the same number of letters as one of the real headers.

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -77,28 +104,45 @@ void SocketReader::read(tcp::socket& socket) {
     // Looping through headers to retrieve JWS token and json body len.
     std::size_t contentLength = 0;
     while(std::getline(header_stream, header) && header != "\r") {
+        // Parse "<name>: <value>" and handle header name case-insensitively
+        std::size_t sep = header.find(':');
+        if (sep == std::string::npos) {
```
abrown-RV, line 109: also really hard to hit with the current testing framework.... we can talk to Nathan about excluding it.

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -77,28 +104,45 @@ void SocketReader::read(tcp::socket& socket) {
...
+
         // Getting jwsToken.
-        if (service_->authorizationEnabled() && jwsToken_.empty() && header.starts_with("Authorization: Bearer ")) {
-            jwsToken_ = header.substr(std::string("Authorization: Bearer ").length());
-            // Removing newline.
-            jwsToken_.erase(jwsToken_.length() - 1);
+        if (service_->authorizationEnabled() && jwsToken_.empty() && iequals_header_name(name, "authorization")) {
+            // Expect "Bearer <token>" (keep scheme check case-sensitive as before)
+            static const std::string kBearerPrefix = "Bearer ";
+            if (value.rfind(kBearerPrefix, 0) == 0) {
```
abrown-RV: This should be `value.starts_with(kBearerPrefix)` (or just regular `find`). `rfind` finds the last index of the input. Which under a crazy bad situation could find "Bearer" somewhere in the base64 data of the token, which would fail when it shouldn't.

abrown-RV (changes requested):
nice work Nelson appreciate you working on optimization! 😎

sdks/cpp/connections/REST/src/SocketReader.cpp
nelgd: Will ask him tommorow

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -278,3 +279,47 @@ TEST_F(RESTSocketReaderTests, SocketReader_LongJsonBody) {
...
+    const auto detailLevel = st2138::Device_DetailLevel_NONE;
+    const std::string language = "en";
+    const std::string jsonBody = "{test_json_body}";
+    writeRequestWithHeaderNames(method, slot, endpoint, fqoid, stream, fields,
+                                jwsToken, origin, detailLevel, language, jsonBody,
+                                "origin",            // lower-case
+                                "AUTHORIZATION",     // upper-case
+                                "dEtAiL-LeVeL",      // mixed-case
+                                "LANGUAGE",          // upper-case
+                                "content-length");   // lower-case
```
abrown-RV: And a random other header that is the same length as one of the real ones so we hit that `return false` at the end of `iequals_header_name` This is where the map of header names to values comes in

abrown-RV (changes requested):
don't forget to put your name at the top of RESTTest.h as well

sdks/cpp/connections/REST/src/SocketReader.cpp
abrown-RV: Still missing coverage on this line. You need a header name with the same length as one of the checked ones.

abrown-RV:
bugbot review

abrown-RV (approved):
Nice work Nelson! 👍

PR #1117, opened 2025-11-27
Audiodeck Pan Knob Fix

### LINKED ISSUE #845: new and more "real" audio deck example
I want a more real audio deck example

cool audio level thing

here are the components:
just msg me if you need more context
```
main componets
	data:
		output: sum of channel signal * channel volume
			Note this can be more complex but dont take all day
	slider_m:
	  this will show a level between -80.0 and 10 and the step will be log 2
		float: -80.0 to 10.0
	select_m:
		bool
		command:|
			on main
				turn on the select
				set the display mode to "LR"
				set the display text to "main"
				set the display img to the Ross video icon
			on each channel
				turn off the select
				set the slider to the volume for its channel
				set the display mode to the channel number
				set the display text to the channel name
				set the display img to the volume img
	solo_m:
		bool
		command:|
			toggle the solo_m
	mute_m:
		bool
		command:|
			toggle the mute_m
	clear_m:
		command:|
			on each channel
				set the solo to false
	display_m:
		img: file
		text: mode
		text: text

channel * 8:
	data: 
		signal: run this code at 60hz
			signal = 0
			for n from 1 to 6:
				signal += (1/n) * sin(2 * 3.1459 * n * (f0+n*10) * t)
			signal *= sin(-alpha * t)
		volume: float
		array of frequencies and there value
			{int:float}
		array of sliders and what frequecny they map to
			{int:int}
		name: string
	select
		bool
		command:|
			set the select to true
			on each channel
					|channel slider| frequencie|
					|++++++++++++++|+++++++++++|
					|	1		   |		 10|
					|	2		   |		 21|
					|	3		   |		 42|
					|	4		   |		 83|
					|	5		   |		166|
					|	6		   |		333|
					|	7		   |		577|
					|	8		   |	   1000|
				set the slider to the level for the frequency
				set the display text to the frequency
				set the display mode to "Freq"
			on main
				turn off the select
				set the display mode to channel number
				set the display text to channels name
				set the display img to a signwave icon or sumthing
	solo
		bool
		command:|
			if selected
				set the solo to true
				set display mode to "SET FREQ"
				set display img to a nice img
				set display txt to frequecny
				set the range for the slider to 
					10 - 20000 : step is 1 and is rounded to int
					
				on hitting the solo buttion again
					save frequency 
					set solo to false
			else
				toggel solo
	mute
		bool
		command:|
			toggle mute
	warning
		bool: COMP
			set true
		int32: CLIP
			if signal is > 1 then mutliply the remander by 100 and round to int
		bool: Freq
			if the frequency has been modifide set to true
	slider
	  this will show a level between -80.0 and 10 and the step will be log 2
		float: -80.0 to 10.0
	display
		img: file
		text: mode
		text: subtext
```

nelgd:
- Fixed the pan knob like we discussed in yesterday's meeting (Almost done)
- Made visualizer for each channel wider
- Removed unneeded code in grid file

### DIFF
--- a/sdks/cpp/connections/gRPC/examples/simple_dashboard_audiodeck/device.simple_dashboard_audiodeck.json
+++ b/sdks/cpp/connections/gRPC/examples/simple_dashboard_audiodeck/device.simple_dashboard_audiodeck.json
@@ -35,7 +35,7 @@
         "ripple": { "type": "FLOAT32", "precision": 1, "constraint": { "type": "FLOAT_RANGE", "float32_range": { "min_value": -30.0, "max_value": 30.0, "display_min": -30.0, "display_max": 30.0, "step": 1.0 } } },
         "resp": { "type": "STRING", "constraint": { "type": "STRING_CHOICE", "string_choice": { "choices": ["Cheb I","Cheb II","Cheb III","Cheb IV","Cheb V","Cheb VI","Cheb VII","Cheb VIII","Cheb IX","Cheb X"] } } },
         "volume": { "type": "FLOAT32", "constraint": { "type": "FLOAT_RANGE", "float32_range": { "min_value": 0.0, "max_value": 100.0, "display_min": 0.0, "display_max": 100.0, "step": 1.0 } } },
-        "fader": { "type": "FLOAT32", "constraint": { "type": "FLOAT_RANGE", "float32_range": { "min_value": 0.0, "max_value": 10.0, "display_min": 0.0, "display_max": 10.0, "step": 0.1 } } }
+        "fader": { "type": "FLOAT32", "constraint": { "type": "FLOAT_RANGE", "float32_range": { "min_value": 0.0, "max_value": 10.0, "display_min": 0.0, "display_max": 10.0, "step": 0.5 } } }
       }
     },
     "audio_deck": {
--- a/sdks/cpp/connections/gRPC/examples/simple_dashboard_audiodeck/device.simple_dashboard_audiodeck.json
+++ b/sdks/cpp/connections/gRPC/examples/simple_dashboard_audiodeck/device.simple_dashboard_audiodeck.json
@@ -174,7 +174,7 @@
                   "ripple": { "float32_value": 3.0 },
                   "resp": { "string_value": "Cheb II" },
                   "volume": { "float32_value": 50.0 },
-                  "fader": { "string_value": "Center" }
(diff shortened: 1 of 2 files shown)


nelgd:
What the Audiodeck looks like now
image

abrown-RV (changes requested):
Awesome Nelson, only comment is tha you had it right before, the Pan should be a knob.

nelgd:
> Awesome Nelson, only comment is tha you had it right before, the Pan should be a knob.1

Yeahhh, I couldnt get the knob to do what I wanted. I couldn't find a way to make the knob default position be up (it's right currently). Also no way to limit the range of motion of the knob

abrown-RV:
Let me talk to a dashboard expert on this one.

abrown-RV:
Aury came through, the way to get the knob to have 0 facing up is using a config element and setting `w.zero` which controls where the zero point is. So you should be able to get it to orient properly.
```
                  <param height="30" top="70" oid=".fader" left="10" relative="true" widget="wheel" width="140">
                     <config key="w.zero">270</config>
                  </param>
```

nelgd:
> Aury came through, the way to get the knob to have 0 facing up is using a config element and setting `w.zero` which controls where the zero point is. So you should be able to get it to orient properly.
> 
> ```
>                   <param height="30" top="70" oid=".fader" left="10" relative="true" widget="wheel" width="140">
>                      <config key="w.zero">270</config>
>                   </param>
> ```

Amazing!!

nelgd:
@abrown-RV Are you able to check with Aury if it is possible to limit the range of motion the wheel has? Thats the last part I need

abrown-RV (changes requested):
```xml
<config key="w.loop">false</config>
```
- Makes the wheel not loop.
- May as well make it go from negative to positive as well while you're at it. Instead of 0 to a positive value.
- I think the `fader` param should be renamed to `pan`. Fader typically refers to the main volume.

nelgd:
> ```
> <config key="w.loop">false</config>
> ```
> 
> * Makes the wheel not loop.
> * May as well make it go from negative to positive as well while you're at it. Instead of 0 to a positive value.
> * I think the `fader` param should be renamed to `pan`. Fader typically refers to the main volume.

Thanks! Those changes are in now

abrown-RV (approved):
thanks Nelson!

PR #1120, opened 2025-12-04
Update getParam to handle array indexes.

### LINKED ISSUE #1064: getParam_ index checks?
Right now (template <EmptyValue> ParamWithValue::getParam_) can only handle structs with an empty value. However when it checks for index, it returns an error. Do we need to handle looking up an array index instead of returning an error?

JChen2005:
Change logic in getParam to allow array indexes.

Previously only allowed strings and indexes returned an error.

Also added/changed unit tests appropriate to the changed code.


unittests/cpp/common/tests/ParamWithValue_test/ParamWithValue_test.cpp
```diff
@@ -342,25 +342,30 @@ TEST_F(ParamWithValueTest, TwoLevelPathAccess) {
...
     EXPECT_TRUE(parentParam.getParam(path, authz_, rc_));
     EXPECT_EQ(rc_.status, catena::StatusCode::OK);
- }
+}
 /*
- * TEST 17 - Testing getParam with numeric index on struct parameter.
+ * TEST 17 - Testing getParam with numeric index instead.
  */
-TEST_F(ParamWithValueTest, NumericIndexOnStructParameterFails) {
+TEST_F(ParamWithValueTest, NumericIndexOnStructParameterFail) {
```
NR-RV: the name of this test is wrong as its a passing test

NR-RV (changes requested):

NR-RV (approved):

Issue #1122, opened 2025-12-04
Go environment container


ctwarog-ross:
Implement dev container for Go development

PR #1134, opened 2025-12-04
Go sdk

ctwarog-ross:
Closes Issue https://github.com/rossvideo/Catena/issues/1122

Introcuding the Go sdk
.devcontainer dir split between cpp and go
cpp toolchain and container renamed, go toolchain and container added
base level go service implemented

device package exposes a barebones device implementation
rest server maps callbacks to 3 url formats (device, param, asset)
basic_device listens annd responds to all messages supported by server

launch config setup to debug basic_device


sdks/go/Makefile
```diff
@@ -0,0 +1,13 @@
...
+all: $(CMDS)
+
+$(CMDS):
+	mkdir -p $(BIN_DIR)/$@
+	go build -o $(BIN_DIR)/$@/$@ ./examples/$@
+
+clean:
+	rm -rf $(BIN_DIR)
+
+.PHONY: all clean $(CMDS)
```
ctwarog-ross: EOF space come on christian

.devcontainer/go/devcontainer.dockerfile
```diff
@@ -0,0 +1,49 @@
...
+
+# Set the working directory
+WORKDIR /home/${USER_NAME}/Catena
+
+USER ${USER_NAME}
+
+RUN go install golang.org/x/tools/gopls@latest \
+    && go install github.com/go-delve/delve/cmd/dlv@latest
+
+ENTRYPOINT ["/bin/sh", "-c", "/bin/bash"]
```
abrown-RV:
```suggestion
ENTRYPOINT ["/bin/sh", "-c", "/bin/bash"]

```

.devcontainer/go/toolchain.env
```diff
@@ -0,0 +1,13 @@
...
+IMAGE_VERSION="v0.0.1"
+IMAGE_VENDOR="Rossvideo"
+
+# Tool versions used across build system
+DOXYGEN_VERSION=1.9.8+ds-2build5
+NODEJS_VERSION=22
+UBUNTU_VERSION=24.04
+VALGRIND_VERSION=1:3.22.0-0ubuntu3
+CURL_VERSION=8.5.0-2ubuntu10.6
+GO_VERSION=go1.22.5.linux-amd64
```
abrown-RV:
```suggestion
GO_VERSION=go1.22.5.linux-amd64

```

.devcontainer/README.md
abrown-RV, line 1: Paths here are wrong, this is in the .devcontainer dir, should reference both dev containers now. Some description on how to pick which dev container? (even if vscode does just take care of it, can give people confidence)

.github/workflows/docker_builder.yml
```diff
@@ -5,18 +5,21 @@ name: Create and publish a Docker images
 on:
   push:
     # branches: ['develop', 'main'] don't build on main for now
-    branches: ['develop']
+    branches: ['develop', 'go_sdk']
```
abrown-RV:
```suggestion
    branches: ['develop']
```

.vscode/launch.json
abrown-RV, line 1: Suggest introducing groups to launch.json:
```json
"presentation": {
  "hidden": false,
  "group": "go",
  "order": 1
}
```
I think cpp, go, node/js/codegen (or something) for groups would go cool. Looks like it separates them by group in the debug menu. Did a quick test with a couple in my branch and it looks like this:

Image

Didn't see a way to filter our targets that don't work in the current dev container, so I think it's the best we can do to de-clutter that menu.

abrown-RV: https://code.visualstudio.com/docs/debugtest/debugging-configuration#_launchjson-attributes

docs/DevContainer.md
abrown-RV, line 1: paths are wrong here too

sdks/go/Makefile
abrown-RV, line 1: can we use cmake? do we want to use cmake? XD

abrown-RV (dismissed):
looks like its working, I've just left comments as comments so I can just approve I don't think any of them were earth shattering and it looks like the cpp will continue to function, so approving since I'm off tomorrow. Feel free to merge with as many comments addressed as you see fit.

sdks/go/Makefile
ctwarog-ross: technically make is seperate, but is there a better thing to use here?

docs/DevContainer.md
NR-RV: for @nelgd to do in another branch

.vscode/launch.json
NR-RV: for @nelgd to do in another branch

.devcontainer/README.md
NR-RV: for @nelgd to do in another branch

NR-RV (approved):

PR #1142, opened 2025-12-10
V1 Go implementation

ctwarog-ross:
Closes issue#1128

logging, catena device framework and catena device example are NOT complete
catena messaging protocol serdes added
Each endpoint is a slot mapped to a callback exposed in BL

basic device example sans catena device model implemented: 
basic_device_bl uses raw fqoid:value maps in BL and a list of ints for slots
ALL error handling, data manipulation and html body handling is done in BL


abrown-RV:
Need to add go's `__debug_bin*` files to git ignore. Simply `__debug_bin*` is sufficient. There is also apparently a launch.json setting to tell it where to put the files, so we could move them to `build/go` as well. But that would be per launch config thing which will clutter up our already very cluttered launch.json XD.... maybe there's a Delve setting for it as well?

Suggest as well we pepper our gitignore with comments breaking it up into sections for go/cpp/etc.

sdks/go/internal/json_utils.go
```diff
@@ -0,0 +1,55 @@
...
+func ReadValueFromRequest(r *http.Request) (any, error) {
+	//Decode body into catena value type
+	var v any
+
+	if err := json.NewDecoder(r.Body).Decode(&v); err != nil {
+		return nil, err
+	}
+	//Convert catena value type to go native type
+	switch val := v.(type) {
+	case map[string]interface{}:
```
abrown-RV: interface{} can be replaced by any (from the go extension)
```suggestion
	case map[string]any
```

abrown-RV (approved):
We've talked offline and in additional issues

PR #1148, opened 2025-12-11
add ssl to the go toolchain

abrown-RV:
I need ssl in the go dev container because I checked out with the ssh url instead of https.


ctwarog-ross (approved):
needs rebase then approved

PR #1152, opened 2025-12-15
Refactor and types

### LINKED ISSUE #1139: Catena error type
Create catena error structure and propagate throughout the project source code

### LINKED ISSUE #1146: Change callbacks to not use http data
potential structure for callbacks instead:
type GetValueHandler func(slot int, fqoid string) (any, error)
type SetValueHandler func(slot int, fqoid string, value any) error

callbacks return values that the server code then handles
move internal function calls to server ant not bl

ctwarog-ross:
Closes #1146
Closes #1139
Progress on #1151

http objects abstracted out of server endpoint callbacks implemented new error type catena.StatusResult for callback returns to signify operation result All non struct types are handled and provided in basic_device_bl.go

### DIFF
--- a/sdks/go/examples/basic_device_bl/basic_device_bl.go
+++ b/sdks/go/examples/basic_device_bl/basic_device_bl.go
@@ -1,100 +1,76 @@
 package main
 
 import (
-	"fmt"
-	"net/http"
-	"os"
-	"reflect"
-	"strconv"
-	"sync"
+    "fmt"
+    "os"
+    "reflect"
+    "strconv"
+    "sync"
 
-	"github.com/rossvideo/catena/sdks/go/internal"
-	"github.com/rossvideo/catena/sdks/go/pkg/logger"
-	"github.com/rossvideo/catena/sdks/go/pkg/rest"
+    "github.com/rossvideo/catena/sdks/go/pkg/logger"
+    "github.com/rossvideo/catena/sdks/go/pkg/rest"
+    "github.com/rossvideo/catena/sdks/go/pkg/catena"
 )
 
-//callback for connect endpoint that returns go channel to connection manager
-
-//Implementation for registering parameter handlers for every fqoid on a given slot
+// Implementation for registering parameter handlers for every fqoid on a given slot
 func registerBasicParamHandlers(srv *rest.Server, params *sync.Map, slot int) {
-    srv.RegisterSetParamHandler(slot, func(w http.ResponseWriter, r *http.Request, slot int, fqoid string) {
+    srv.RegisterSetParamHandler(slot, func(value any, slot int, fqoid string) catena.StatusResult {
         logger.Info("Slot %d: SetParam %s", slot, fqoid)
-        v, err := internal.ReadValueFromRequest(r)
-        if err != nil {
-            logger.Error("Slot %d: SetParam %s - invalid JSON: %v", slot, fqoid, err)
-            http.Error(w, "invalid JSON", http.StatusBadRequest)
-            return
-        }
         val, ok := params.Load(fqoid)
         if !ok {
             logger.Warning("Slot %d: SetParam %s - param not found", slot, fqoid)
-            http.Error(w, "param not found", http.StatusNotFound)
-            return
+            return catena.NotFound("param not found")
         }
-        if reflect.TypeOf(val) != reflect.TypeOf(v) {
+        if reflect.TypeOf(val) != reflect.TypeOf(value) {
             logger.Error("Slot %d: SetParam %s - type mismatch", slot, fqoid)
-            http.Error(w, "type mismatch", http.StatusBadRequest)
-            return
+            return catena.BadRequest("type mismatch")
         }
-        params.Store(fqoid, v)
-        w.WriteHeader(http.StatusNoContent)
+        params.Store(fqoid, value)
+        return catena.NoContent()
     })
 
-    srv.RegisterGetParamHandler(slot, func(w http.ResponseWriter, r *http.Request, slot int, fqoid string) {
+    srv.RegisterGetParamHandler(slot, func(slot int, fqoid string) catena.StatusResult {
         logger.Info("Slot %d: GetParam %s", slot, fqoid)
         v, ok := params.Load(fqoid)
         if !ok {
             logger.Warning("Slot %d: GetParam %s - param not found", slot, fqoid)
-            http.Error(w, "param not found", http.StatusNotFound)
-            return
+            return catena.NotFound("param not found")
         }
-        internal.WriteResponseJSON(w, v)
+        return catena.OK(v)
     })
 }
 
-//For a given slot implement param handlers for a specific param oid only
+// For a given slot implement param handlers for a specific param oid only
 func registerSpecificParamHandlers(srv *rest.Server, params *sync.Map, fqoid string, slot int) {
-    //Implementation for registering parameter handlers
-    srv.RegisterSetParamHandler(slot, func(w http.ResponseWriter, r *http.Request, slot int, fqoid_ string) {
+    srv.RegisterSetParamHandler(slot, func(value any, slot int, fqoid_ string) catena.StatusResult {
         if fqoid_ != fqoid {
-            srv.DefaultSetParamValueHandler(w, r, slot, fqoid_)
-            return
+            return catena.NotImplemented("no handler for fqoid " + fqoid_)
         }
         logger.Info("Slot %d: SetSpecificParam %s", slot, fqoid_)
-        v, err := internal.ReadValueFromRequest(r)
-        if err != nil {
-            logger.Error("Slot %d: SetSpecificParam %s - invalid JSON: %v", slot, fqoid_, err)
-            http.Error(w, "invalid JSON", http.StatusBadRequest)
-            return
-        }
         val, ok := params.Load(fqoid_)
         if !ok {
             logger.Warning("Slot %d: SetSpecificParam %s - param not found", slot, fqoid_)
-            http.Error(w, "param not found", http.StatusNotFound)
-            return
+            return catena.NotFound("param not found")
         }
-        if reflect.TypeOf(val) != reflect.TypeOf(v) {
+        if reflect.TypeOf(val) != reflect.TypeOf(value) {
             logger.Error("Slot %d: SetSpecificParam %s - type mismatch", slot, fqoid_)
-            http.Error(w, "type mismatch", http.StatusBadRequest)
-            return
+            return catena.BadRequest("type mismatch")
         }
-        params.Store(fqoid_, v)
-        w.WriteHeader(http.StatusNoContent)
+        params.Store(fqoid_, value)
+        return catena.NoContent()
     })
 
-    srv.RegisterGetParamHandler(slot, func(w http.ResponseWriter, r *http.Request, slot int, fqoid_ string) {
+    srv.RegisterGetParamHandler(slot, func(slot int, fqoid_ string) catena.StatusResult {
         if fqoid_ != fqoid {
-            srv.DefaultGetParamValueHandler(w, r, slot, fqoid_)
-            return
+            return catena.NotImplemented("no handler for fqoid " + fqoid_)
         }
         logger.Info("Slot %d: GetSpecificParam %s", slot, fqoid_)
         v, ok := params.Load(fqoid_)
         if !ok {
             logger.Warning("Slot %d: GetSpecificParam %s - param not found", slot, fqoid_)
-            http.Error(w, "param not found", http.StatusNotFound)
-            return
+            return catena.NotFound("param not found")
         }
-        internal.WriteResponseJSON(w, v)
+        return catena.OK(v)
     })
 }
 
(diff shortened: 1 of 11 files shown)


sdks/go/pkg/catena/status_result.go
```diff
@@ -0,0 +1,20 @@
+package catena
+
+import (
+	"net/http"
+)
+// StatusResult represents an HTTP outcome (success or error) that the server can render uniformly.
+type StatusResult struct {
+    Status  int    // e.g., 200, 204, 400, 404, 405, 501, 500
```
abrown-RV: This is still quite HTTP centric, fine for now, but when the time comes we'll have to decide how we map these constructors to the different http/grpc codes.

sdks/go/pkg/catena/status_result.go
```diff
@@ -0,0 +1,20 @@
...
+}
+
+// Convenience constructors
+func OK(payload any) StatusResult                 { return StatusResult{Status: http.StatusOK, Payload: payload} }
+func NoContent() StatusResult                     { return StatusResult{Status: http.StatusNoContent} }
+func BadRequest(msg string) StatusResult          { return StatusResult{Status: http.StatusBadRequest, Message: msg} }
+func NotFound(msg string) StatusResult            { return StatusResult{Status: http.StatusNotFound, Message: msg} }
+func MethodNotAllowed(msg string) StatusResult    { return StatusResult{Status: http.StatusMethodNotAllowed, Message: msg} }
+func NotImplemented(msg string) StatusResult      { return StatusResult{Status: http.StatusNotImplemented, Message: msg} }
+func InternalServerError(msg string) StatusResult { return StatusResult{Status: http.StatusInternalServerError, Message: msg} }
```
abrown-RV: not totally up on go, but can the status result have private fields so that no one can monkey with it? is there a catena.StatusResult() ctor that would create an empty one? both things we should prevent if we can.

Basically I'm wanting to make sure you can't circumvent the Convenience constructors.

sdks/go/pkg/catena/status_result.go
```diff
@@ -0,0 +1,20 @@
+package catena
+
+import (
+	"net/http"
+)
+// StatusResult represents an HTTP outcome (success or error) that the server can render uniformly.
+type StatusResult struct {
+    Status  int    // e.g., 200, 204, 400, 404, 405, 501, 500
+    Payload any    // optional body to JSON-encode on success (2xx). If nil and Status is 204, no body.
+    Message string // optional error message for non-2xx statuses
```
abrown-RV: Maybe want a Exception field as well? would let creators add exception details that could at least be logged, even if it wouldn't go back to the user. Although what's our pattern, do we just let callbacks throw exceptions?

sdks/go/pkg/rest/server.go
```diff
@@ -1,226 +1,248 @@
...
+    if s != "" {
+        return s
+    }
+    return fallback
+}
 
+// Handlers now return HTTPResult so the server can respond consistently.
+type DeviceHandler func(w http.ResponseWriter, r *http.Request) catena.StatusResult
+type GetParamValueHandler func(slot int, fqoid string) catena.StatusResult
+type SetParamValueHandler func(value any, slot int, fqoid string) catena.StatusResult
```
abrown-RV: cpp sdk just calls these GetValue and SetValue should we be consistant?

sdks/go/pkg/rest/server.go
```diff
@@ -1,226 +1,248 @@
...
-				s.DefaultGetParamValueHandler(w, r, slot, fqoid)
-				return
-			}
+                    res := handler(slot, fqoid)
+                    writeHTTPResult(w, res)
+                    logger.Info("GET /value/%s finished", fqoid)
+                    return
+                }
+                writeHTTPResult(w, s.DefaultGetParamValueHandler(w, r, slot, fqoid))
             case http.MethodPut, http.MethodPatch:
```
abrown-RV: why patch? openapi.yaml only specifies PUT and GET for /value

sdks/go/pkg/rest/server.go
```diff
@@ -1,226 +1,248 @@
...
+                    writeHTTPResult(w, res)
+                    logger.Info("%s /value/%s finished", r.Method, fqoid)
+                    return
+                }
+                writeHTTPResult(w, s.DefaultSetParamValueHandler(value, slot, fqoid))
             default:
-				logger.Warning("%s /value/%s: method not allowed", r.Method, fqoid)
-                http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+                logger.Warning("%s /value/%s: method not allowed", r.Method, fqoid)
+                writeHTTPResult(w, catena.MethodNotAllowed("method not allowed"))
```
abrown-RV: Method is pretty specific to HTTP, I know its convienent to use the StatusResult here, but I think MethodNotAllowed pollutes the StatusResult with a too specific condition. And we're not in bix logic code here, so it's ok to not use StatusResult

sdks/go/pkg/rest/server.go
```diff
@@ -1,226 +1,248 @@
...
+    h, ok := s.getAsset[slot]
+    return h, ok
 }
 
 func (s *Server) RegisterRoutes() {
-	logger.Debug("Registering routes for %d slots", len(s.slotList))
+    logger.Debug("Registering routes for %d slots", len(s.slotList))
     for _, slot := range s.slotList {
         // capture loop variable
         slot := slot
```
abrown-RV: vscode seems to think this is not needed?

sdks/go/pkg/rest/server.go
```diff
@@ -1,226 +1,248 @@
...
-    if _, ok := s.getParamValue[slot]; !ok {
-        s.getParamValue[slot] = handler
-		logger.Debug("Registered GET param handler for slot %d", slot)
-    }
+    s.getParamValue[slot] = handler
+    logger.Debug("Registered GET param handler for slot %d", slot)
 }
 
-func (s *Server) RegisterSetParamHandler(slot int, handler ParamValueHandler) {
+func (s *Server) RegisterSetParamHandler(slot int, handler SetParamValueHandler) {
```
abrown-RV: `RegisterSetParamValueHandler` or `RegisterSetValueHandler`

sdks/go/pkg/rest/server.go
```diff
@@ -1,226 +1,248 @@
...
-        s.getDevice = handler
-		logger.Debug("Registered device handler for slot %d", slot)
-    }
+    // getDevice is global; last registration wins.
+    s.getDevice = handler
+    logger.Debug("Registered device handler (global), requested by slot %d", slot)
 }
 
-func (s *Server) RegisterGetParamHandler(slot int, handler ParamValueHandler) {
+func (s *Server) RegisterGetParamHandler(slot int, handler GetParamValueHandler) {
```
abrown-RV: `RegisterGetParamValueHandler` or `RegisterGetValueHandler`

sdks/go/examples/basic_device_bl/basic_device_bl.go
```diff
@@ -1,100 +1,76 @@
...
+            return catena.NotFound("param not found")
         }
-        internal.WriteResponseJSON(w, v)
+        return catena.OK(v)
     })
 }
 
-//For a given slot implement param handlers for a specific param oid only
+// For a given slot implement param handlers for a specific param oid only
 func registerSpecificParamHandlers(srv *rest.Server, params *sync.Map, fqoid string, slot int) {
```
abrown-RV, line 44: Do we want this as an example? Isn't basically the same as the Basic Handler, and you would not be able to register 2 specific handlers for the same slot. As in you could not have
```go
registerSpecificParamHandlers(srv, params1, "alpha", 1)
registerSpecificParamHandlers(srv, params1, "counter", 1)
```
the specific handler for `alpha` would be overwritten by the handler for `counter` so it seems like a bad pattern at have as an example.

sdks/go/examples/basic_device_bl/basic_device_bl.go
```diff
@@ -1,100 +1,76 @@
...
+        if reflect.TypeOf(val) != reflect.TypeOf(value) {
             logger.Error("Slot %d: SetParam %s - type mismatch", slot, fqoid)
-            http.Error(w, "type mismatch", http.StatusBadRequest)
-            return
+            return catena.BadRequest("type mismatch")
         }
-        params.Store(fqoid, v)
-        w.WriteHeader(http.StatusNoContent)
+        params.Store(fqoid, value)
+        return catena.NoContent()
```
abrown-RV: I feel like we could be doing a little more for the return type... Here we are relying on the user to know that SetValue is supposed to return `NoContent` on success, but other endpoints return `OK` on success. I think we should just have one success type and the server handles converting that to the approriate response. Oh this is actually a grpc/rest argument as well, in grpc SetValue responds with an st2138.empty and http with 204 status.

abrown-RV (commented):
Just some comments, I know there's more changes in the pipe so I'll save approval for then

sdks/go/pkg/catena/status_result.go
ctwarog-ross: Thats the Message field, I felt that name was more appropriate. Or are you thinking another string?

sdks/go/pkg/rest/server.go
ctwarog-ross: good catch, chatgpt put patch in there. It should be removed

sdks/go/pkg/rest/server.go
ctwarog-ross: Im actually not sure, do you think we should?

sdks/go/examples/basic_device_bl/basic_device_bl.go
ctwarog-ross: it was just to show you can implement more specific handlers than a blanket case

sdks/go/pkg/rest/server.go
ctwarog-ross: slot is used several times in the function body

sdks/go/pkg/catena/status_result.go
ctwarog-ross: will be refactored in future work

sdks/go/examples/basic_device_bl/basic_device_bl.go
```diff
@@ -2,160 +2,163 @@ package main
...
+    params2.Store("struct_variant_array", []catena.StructVariantValue{
+        {
+            StructVariantType: "typeA",
+            Value:             "value A1",
+        },
+        {
+            StructVariantType: "typeB",
+            Value:             123,
+        },
+    })
```
NR-RV: intenting switched from tabs to spaces

NR-RV (changes requested):

sdks/go/pkg/rest/server.go
```diff
@@ -6,221 +6,239 @@ import (
...
-	http.Error(w, "no getParamValue handler registered for slot "+strconv.Itoa(slot)+": fqoid /"+fqoid, http.StatusNotImplemented)
+	return catena.NotImplemented("no getParamValue handler registered for slot " + strconv.Itoa(slot))
 }
 
-func (s *Server) DefaultSetParamValueHandler(w http.ResponseWriter, r *http.Request, slot int, fqoid string) {
+func (s *Server) DefaultSetParamValueHandler(value any, slot int, fqoid string) catena.StatusResult {
 	logger.Warning("PUT /value/%s: no handler registered for slot %d", fqoid, slot)
-	http.Error(w, "no setParamValue handler registered for slot "+strconv.Itoa(slot)+": fqoid /"+fqoid, http.StatusNotImplemented)
+	return catena.NotImplemented("no setParamValue handler registered for slot " + strconv.Itoa(slot))
 }
```
abrown-RV: For consistancy should rename these too

sdks/go/pkg/rest/server.go
```diff
@@ -6,221 +6,239 @@ import (
...
-    if m, ok := s.setParamValue[slot]; ok {
-        return m, ok
-    }
-    return nil, false
+func (s *Server) lookupSetParamValue(slot int) (SetValueHandler, bool) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	h, ok := s.setValue[slot]
+	return h, ok
 }
```
abrown-RV, line 151: for consistancy should rename these too

abrown-RV (changes requested):
You've got the in-progress logging stuff in here somehow

sdks/go/pkg/rest/server.go
ctwarog-ross: wdym in progress logging? I have what was in his original submission

NR-RV (approved):

sdks/go/internal/json_utils.go
```diff
@@ -13,43 +15,144 @@ func ReadValueFromRequest(r *http.Request) (any, error) {
...
+				return catena.StructVariantValue{
+					StructVariantType: svType,
+					Value:             svValue,
+				}, nil
 			}
-		} else if stringVal, ok := val["string_value"]; ok {
-			if s, ok := stringVal.(string); ok {
-				return s, nil
+		}
+	} else if structVariantArrVal, ok := val["struct_variant_array_value"]; ok {
```
abrown-RV, line 88: These should be recursive

sdks/go/internal/json_utils.go
```diff
@@ -13,43 +15,144 @@ func ReadValueFromRequest(r *http.Request) (any, error) {
...
+		res.Payload = map[string]map[string]any{"struct_variant_value": {
+			"struct_variant_type": val.StructVariantType,
+			"value":              val.Value,
+		}}
+	case []catena.StructVariantValue:
+		var structVariantArr []map[string]any
+		for _, item := range val {
+			structVariantArr = append(structVariantArr, map[string]any{
+				"struct_variant_type": item.StructVariantType,
+				"value":              item.Value,
```
abrown-RV, line 147: recursive here too, on all the struct types

abrown-RV (approved):

PR #1158, opened 2025-12-16
GO Logging Improvements

### LINKED ISSUE #1153: Additional Logging Features
Discussed in meeting adding features to go. 

- Need to add debug/release logs
- More types of logs like we have in C++

Using an already made library might be the best way to do this

nelgd:
Logging now uses GO's build in `log/slog` lib

To see different log levels use the following args when executing:

- No args = Error
- `-v` = Error and Warning
- `-vv` = Error, Warning, and Info
- `-vvv` = Debug

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,7 +10,7 @@
         "cwd": "${workspaceFolder}",
         "args": [],
         "env": {
-          "CATENA_DEBUG": "true" // Enable Debug Mode
+          "CATENA_LOG_LEVEL": "DEBUG", // Set log level to DEBUG
         }
       },
       {
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,9 @@
         "program": "${workspaceFolder}/sdks/go/examples/${input:pickGoService}",
         "cwd": "${workspaceFolder}",
         "args": [],
-        "env": {}
+        "env": {
+          "CATENA_DEBUG": "true" // Enable Debug Mode
+        }
       },
       {
         "type": "cppdbg",
--- a/sdks/go/README.md
+++ b/sdks/go/README.md
@@ -100,4 +100,17 @@ func (v Variant) Print() {
         fmt.Println("string:", v.StrVal)
     }
 }
+
+## Logging Config
+
+Config can be changed with env variables:
+
+| Variable | Description | Allowed values | Default |
+|---|---|---|---|
+| `CATENA_LOG_DIR` | Directory for log files | Path | `./logs` |
+| `CATENA_SILENT` | Suppress all output | `true` | — |
+| `CATENA_LOG_LEVEL` | Logging verbosity | `debug`, `info`, `warn`, `error` | `error` |
+| `CATENA_LOG_FILE` | Enable file logging | `true`, `false` | `true` |
+| `CATENA_LOG_CONSOLE` | Enable console logging | `true`, `false` | `true` |
+| `CATENA_LOG_JSON` | Output logs as JSON | `true` | — |
 ```
--- a/sdks/go/README.md
+++ b/sdks/go/README.md
@@ -100,6 +100,7 @@ func (v Variant) Print() {
         fmt.Println("string:", v.StrVal)
     }
 }
+```
 
 ## Logging Config
 
@@ -113,4 +114,15 @@ Config can be changed with env variables:
 | `CATENA_LOG_FILE` | Enable file logging | `true`, `false` | `true` |
 | `CATENA_LOG_CONSOLE` | Enable console logging | `true`, `false` | `true` |
 | `CATENA_LOG_JSON` | Output logs as JSON | `true` | — |
+
+### Custom Prefix
+
+By default, the logger uses `CATENA` as the environment variable prefix. If you want to use your own prefix, pass a custom prefix to `ParseFromEnv`:
+
+```go
+// Use the default CATENA prefix
+cfg := logger.ParseFromEnv("")  // reads CATENA_LOG_LEVEL, CATENA_SILENT, etc.
+
+// Use a custom prefix for your application
+cfg := logger.ParseFromEnv("MYAPP")  // reads MYAPP_LOG_LEVEL, MYAPP_SILENT, etc.
 ```
--- a/sdks/go/examples/basic_device_bl/basic_device_bl.go
+++ b/sdks/go/examples/basic_device_bl/basic_device_bl.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"reflect"
 	"strconv"
@@ -18,7 +19,7 @@ func registerBasicParamHandlers(srv *rest.Server, params *sync.Map, slot int) {
 		logger.Info("SetParam", "slot", slot, "fqoid", fqoid)
 		val, ok := params.Load(fqoid)
 		if !ok {
-			logger.Warning("SetParam param not found", "slot", slot, "fqoid", fqoid)
+			logger.Error("SetParam param not found", "slot", slot, "fqoid", fqoid)
 			return catena.NotFound("param not found")
 		}
 		if reflect.TypeOf(val) != reflect.TypeOf(value) {
@@ -33,7 +34,7 @@ func registerBasicParamHandlers(srv *rest.Server, params *sync.Map, slot int) {
 		logger.Info("GetParam", "slot", slot, "fqoid", fqoid)
 		v, ok := params.Load(fqoid)
 		if !ok {
-			logger.Warning("GetParam param not found", "slot", slot, "fqoid", fqoid)
+			logger.Error("GetParam param not found", "slot", slot, "fqoid", fqoid)
 			return catena.NotFound("param not found")
 		}
 		return catena.OK(v)
@@ -49,7 +50,7 @@ func registerSpecificParamHandlers(srv *rest.Server, params *sync.Map, fqoid str
 		logger.Info("SetSpecificParam", "slot", slot, "fqoid", fqoid_)
 		val, ok := params.Load(fqoid_)
 		if !ok {
-			logger.Warning("SetSpecificParam param not found", "slot", slot, "fqoid", fqoid_)
+			logger.Error("SetSpecificParam param not found", "slot", slot, "fqoid", fqoid_)
 			return catena.NotFound("param not found")
 		}
 		if reflect.TypeOf(val) != reflect.TypeOf(value) {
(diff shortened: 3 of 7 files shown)


sdks/go/pkg/logger/config.go
```diff
@@ -38,35 +38,93 @@
...
+
+	// DebugMode enables verbose logging (all levels including DEBUG)
+	// When true, ignores Verbosity and shows all levels
+	DebugMode bool
+
+	// Verbosity controls log level in release mode (when DebugMode is false)
+	// 0 (no -v): ERROR only
+	// 1 (-v): ERROR, WARNING
+	// 2 (-vv): ERROR, WARNING, INFO
+	Verbosity int
```
abrown-RV: More traditional to just have a Level setting, applications then just set the level they prefer, instead of abstracting that away with multiple config options

sdks/go/examples/basic_device_bl/basic_device_bl.go
```diff
@@ -99,15 +99,16 @@ func registerSpecificParamHandlers(srv *rest.Server, params *sync.Map, fqoid str
 }
 
 func main() {
-    // Initialize the logger
     logDir := envOr("CATENA_LOG_DIR", "./logs")
     silent := os.Getenv("CATENA_SILENT") == "true"
+    debugMode := os.Getenv("CATENA_DEBUG") == "true"
```
abrown-RV: suggest a `ParseFromEnv` function generate a config options from env vars. Make sure that it has a prefix param that can default to `"CATENA"` for maximum portability.

sdks/go/pkg/logger/logger.go
```diff
@@ -38,118 +38,216 @@
...
-    file       *os.File
-    writers    []io.Writer
+	config  Config
+	slogger *slog.Logger
+	file    *os.File
 }
 
-// Init initializes the global logger (matches C++ Logger::init)
+// Init initializes the global logger
 func Init(cfg Config) error {
```
abrown-RV, line 68: should probably throw an error or something if this is called multiple times to indicate to the user that they can't re-init logging

sdks/go/pkg/logger/logger.go
```diff
@@ -38,118 +38,216 @@
...
-    var initErr error
-    once.Do(func() {
-        globalLogger = &Logger{config: cfg}
-        initErr = globalLogger.setup()
-    })
-    return initErr
+	var initErr error
+	once.Do(func() {
+		globalLogger = &Logger{config: cfg}
+		initErr = globalLogger.setup()
```
abrown-RV: should nil globalLogger if setup returns an error so there's no partially setup logger floating around

abrown-RV: Actually could all of this be simplified to not even deal with the globalLogger?
just give the logger to slog.SetDefault, and then all those passthrough Info() Warn() etc functions can just call slog.Info() removing those extra if checks?

abrown-RV (changes requested):
all other logs (like in server.go) need to be updated with attributes (bugbot also pointed this out)

I'd like to see also what other logging frameworks call "named loggers" the idea being that the server for example will use it's own custom logger with some kind of attribute on it to identify logs as coming from the server specifically. With slog that appears to be the `.With()` func. So I think we'd want a `GetLogger` or `GetNamed` or something in the logger wrapper that returns just the default logger `.With("l", name)` copilot suggested attribute names like "component" or "logger", I kind of like logger, but maybe just "l" for short? so it doesn't clutter up the output too much? idk, something to debate. Let me know if this all makes sense.

nelgd:
Level settings have been implemented and the other changes have been made. 

Config can be changed with env variables:

| Variable | Description | Allowed values | Default |
|---|---|---|---|
| `CATENA_LOG_DIR` | Directory for log files | Path | `./logs` |
| `CATENA_SILENT` | Suppress all output | `true` | — |
| `CATENA_LOG_LEVEL` | Logging verbosity | `debug`, `info`, `warn`, `error` | `error` |
| `CATENA_LOG_FILE` | Enable file logging | `true`, `false` | `true` |
| `CATENA_LOG_CONSOLE` | Enable console logging | `true`, `false` | `true` |
| `CATENA_LOG_JSON` | Output logs as JSON | `true` | — |

ctwarog-ross:
bugbot run

ctwarog-ross:
bugbot run

sdks/go/pkg/rest/server.go
```diff
@@ -27,10 +28,18 @@ type Server struct {
...
 	mux 			http.ServeMux
     mu             	sync.RWMutex
+	log				*logger.Logger
+}
+
+// serverLog returns a named logger for the REST server.
+// This is a function rather than a package-level variable to ensure it uses
+// the configured logger handler after logger.Init() has been called.
+func serverLog() *slog.Logger {
+	return logger.GetNamed("rest-server")
```
abrown-RV: It would be better to call this once and save the logger as a class field instead of the overhead of creating a new logger every time we want to log

sdks/go/pkg/logger/logger.go
```diff
@@ -38,118 +38,192 @@
...
+
+	if initialized {
+		return ErrAlreadyInitialized
+	}
+
+	if initErr != nil {
+		return initErr
+	}
+
+	once.Do(func() {
```
abrown-RV: with the mutex, so you still need the once call? They accomplish the same thing right?

sdks/go/pkg/logger/logger.go
```diff
@@ -38,118 +38,192 @@
...
+	globalLogger *Logger
+	once         sync.Once
+	initialized  bool
+	initErr      error
+	initMu       sync.Mutex
 )
 
-// Logger handles logging to console and file
+// Logger handles structured logging using slog
 type Logger struct {
```
abrown-RV: can it be private? just holds 2 ptrs, don't think there's any reason for anyone to use outside of the package?

sdks/go/examples/basic_device_bl/basic_device_bl.go
```diff
@@ -99,18 +100,24 @@ func registerSpecificParamHandlers(srv *rest.Server, params *sync.Map, fqoid str
...
+    // This allows: ./basic_device_bl -v (warn level) or -vv (info level)
+    cliLevel := logger.ParseVerbosityFromOS()
+    if cliLevel < cfg.Level {
+        cfg.Level = cliLevel
+    }
+
+    // Check for debug mode via environment variable (sets DEBUG level)
+    if os.Getenv("CATENA_DEBUG") == "true" {
+        cfg.Level = slog.LevelDebug
+    }
```
abrown-RV: Just set `CATENA_LOG_LEVEL` in the launch.json

sdks/go/examples/basic_device_bl/basic_device_bl.go
```diff
@@ -99,18 +100,24 @@ func registerSpecificParamHandlers(srv *rest.Server, params *sync.Map, fqoid str
...
+    // This checks: CATENA_LOG_DIR, CATENA_SILENT, CATENA_LOG_LEVEL, etc.
+    cfg := logger.ParseFromEnv("CATENA")
+    cfg.AppName = "basic_device_bl"
+
+    // Override level from command-line verbosity flags (-v, -vv) if specified
+    // This allows: ./basic_device_bl -v (warn level) or -vv (info level)
+    cliLevel := logger.ParseVerbosityFromOS()
+    if cliLevel < cfg.Level {
+        cfg.Level = cliLevel
+    }
```
abrown-RV: Is there a nice clear way to combine this logic and handle it for implementors?

abrown-RV (changes requested):

.vscode/launch.json
```diff
@@ -9,7 +9,9 @@
         "program": "${workspaceFolder}/sdks/go/examples/${input:pickGoService}",
         "cwd": "${workspaceFolder}",
         "args": [],
-        "env": {}
+        "env": {
+          "CATENA_DEBUG": "true" // Enable Debug Mode
```
abrown-RV: No longer needed?

abrown-RV: actually I think it should be:
```
"CATENA_LOG_LEVEL": "DEBUG", // Set log level to DEBUG
```

sdks/go/pkg/logger/config.go
```diff
@@ -38,35 +38,164 @@
...
+		return slog.LevelError
+	}
+}
+
+// ParseVerbosity parses command-line arguments for -v and -vv flags
+// Returns the corresponding slog.Level:
+//   - 0 (no -v): ERROR only
+//   - 1 (-v): WARNING and ERROR
+//   - 2+ (-vv): INFO, WARNING, and ERROR
+func ParseVerbosity(args []string) slog.Level {
```
abrown-RV, line 170: I think this function could be simplified, also may as well do -vvv for debug

sdks/go/pkg/logger/config.go
```diff
@@ -38,35 +38,164 @@
...
+		prefix = "CATENA"
+	}
+
+	cfg := DefaultConfig()
+
+	if v := os.Getenv(prefix + "_LOG_DIR"); v != "" {
+		cfg.LogDir = v
+	}
+
+	if os.Getenv(prefix+"_SILENT") == "true" {
```
abrown-RV: as bugbot suggests, true values in env vars should be more permissive, people can be pretty inconsistant about what they put in the env var and expect it to be true.

abrown-RV:
You need to update the description and resolve merge conflicts as well

ctwarog-ross:
bugbot run

ctwarog-ross:
> Level settings have been implemented and the other changes have been made.
> 
> Config can be changed with env variables:
> 
> Variable	Description	Allowed values	Default
> `CATENA_LOG_DIR`	Directory for log files	Path	`./logs`
> `CATENA_SILENT`	Suppress all output	`true`	—
> `CATENA_LOG_LEVEL`	Logging verbosity	`debug`, `info`, `warn`, `error`	`error`
> `CATENA_LOG_FILE`	Enable file logging	`true`, `false`	`true`
> `CATENA_LOG_CONSOLE`	Enable console logging	`true`, `false`	`true`
> `CATENA_LOG_JSON`	Output logs as JSON	`true`	—

could you add this to the readme?

sdks/go/README.md
```diff
@@ -100,4 +100,17 @@ func (v Variant) Print() {
         fmt.Println("string:", v.StrVal)
     }
 }
+
+## Logging Config
+
+Config can be changed with env variables:
```
abrown-RV, line 107: expand on the customizable prefix if implementors don't want to use the default.

abrown-RV (approved):

PR #1161, opened 2025-12-18
Added documentation for validator.md

JChen2005:
Changes made to validator.md

remake the validation.md so that the title is "Validating Device Models"
a subsection that explains and links to st2138-a
make 3 sections:
-Why a Validator
-Required parts for a valid device model
-How to run the validator


docs/Validation.md
```diff
@@ -1,42 +1,279 @@
...
-ams/variantlist_example/param.AudioSlot.json
-applying: param schema to input file ./example_device_models/device_tree/params/variantlist_example/param.AudioSlot.json
-data was valid.
+- Type mismatches (e.g., using a string where an integer is required)
+- Missing required fields (e.g., mandatory product information)
+- Invalid enum values
+- Incorrect nested structures
+- Schema constraint violations
+
+![Device Model Error Highlighting](images/device.minimal.error.png)
```
abrown-RV, line 33: maybe have some example import paths? this implies to me that there exists a directory called include, and it somehow import the files in that dir.

docs/Validation.md
```diff
@@ -1,42 +1,279 @@
...
+According to the ST2138-a specification and enforced by the [validation logic](../tools/codegen/index.js#L80-L167), a valid device model must include:
+
+### 1. **Device Metadata**
+
+- **`slot`**: Device slot number
+- **`default_scope`**: Default access scope (e.g., `"st2138:mon"`)
+- **`detail_level`**: Default detail level for responses
+- **`access_scopes`**: Array of supported access scopes
+- **`multi_set_enabled`**: Boolean indicating multi-set support
+- **`subscriptions`**: Boolean indicating subscription support
```
abrown-RV, line 116: None of these are a must. Also missing fields like params, commands, languages, etc.

docs/Validation.md
```diff
@@ -1,42 +1,279 @@
...
+  "params": {
+    "product": {
+      "type": "STRUCT",
+      "read_only": true,
+      "access_scope": "st2138:mon",
+      "params": {
+        "name": { "type": "STRING" },
+        "vendor": { "type": "STRING" },
+        "version": { "type": "STRING" },
+        "serial_number": { "type": "STRING" }
```
abrown-RV: All 6 params still have to be defined, just the optional one(s) don't need values. Should probably include an explaination about that: the struct needs all the fields, but the values can be generated automatically for them if they're optional. (looks like we're missing the generation for catena_sdk, making a ticket)

docs/Validation.md
```diff
@@ -1,42 +1,279 @@
...
 ```
 
-`validate.js` uses the `ajv` module to perform the validation.
+### 3. **Valid Parameter Types**
+
+All parameters must use one of the supported types defined in the protobuf specification:
+
+- Scalar: `INT32`, `INT64`, `FLOAT32`, `FLOAT64`, `STRING`, `BOOLEAN`
+- Complex: `STRUCT`, `VARIANT`, `ARRAY`
+- Special: Commands and external objects
```
abrown-RV: not correct, we don't have 64 types, or bool
the complex list is longer as well

docs/Validation.md
```diff
@@ -1,42 +1,279 @@
...
+There are multiple ways to validate your device models:
+
+### Method 1: Command-Line Validation (Standalone)
+
+The standalone validator is located in [smpte/tools/validate.js](../smpte/tools/validate.js).
+
+**Usage:**
+
+```bash
+cd /home/jasonchen/Catena
```
abrown-RV:
```suggestion
cd ~/Catena
```

docs/Validation.md
```diff
@@ -1,42 +1,279 @@
...
+node smpte/tools/validate.js <path-to-your-file>
```
abrown-RV, line 216: should we include stuff about running smpte's install-tooling.sh and build-openapi.sh?

docs/Validation.md
```diff
@@ -1,42 +1,279 @@
...
+
+1. Extracts the schema name from the filename prefix (e.g., `device`, `param`)
+2. Loads the appropriate schema from [device.json](../smpte/interface/schemata/device.json)
+3. Parses the input file (supports both JSON and YAML)
+4. Uses [AJV (Another JSON Validator)](https://ajv.js.org/) to validate against the schema
+5. Reports errors with line numbers using source maps
+
+### Method 2: NPM Package (Published Validator)
+
+Install the official ST2138-a validator globally:
```
abrown-RV: It's not installable globally like that, there's no published package yet. It can be installed with a `path:` uri in the package.json, but even then you can to use npx

docs/Validation.md
```diff
@@ -1,42 +1,279 @@
...
+```
+
+This uses the published schema from the SMPTE GitHub repository.
+
+### Method 3: Integrated Code Generation
+
+When generating code, validation happens automatically. The [codegen tool](../tools/codegen/index.js) validates before generating:
+
+```bash
+cd /home/jasonchen/Catena/tools/codegen
```
abrown-RV:
```suggestion
cd ~/Catena/tools/codegen
```
best not hardcode with your username

abrown-RV (changes requested):
some suggestions and some inaccuracies

docs/Validation.md
```diff
@@ -1,42 +1,305 @@
...
+- Root device model: Look in the `params` sub-folder for files named `param.<name>.json` or `param.<name>.yaml`
+- Nested imports: Look in sub-folders matching the parent param name
+
+File system for `device.import_params.json` example:
+
+![Device Tree Folder Structure](images/import_params_tree.png)
+
+### 3. **IDE Integration**
+
+With proper configuration, validators provide real-time feedback in your IDE:
```
ctwarog-ross, line 80: how is the configuration setup

docs/Validation.md
```diff
@@ -1,42 +1,305 @@
...
+- Authorization scope misconfigurations
+- Constraint reference errors
+
+The validation logic is integrated into the code generation pipeline in [index.js](../tools/codegen/index.js#L173-L187), ensuring no invalid models make it to production.
+
+---
+
+## Required Parts for a Valid Device Model
+
+According to the ST2138-a specification and enforced by the [validation logic](../tools/codegen/index.js#L80-L167), a valid device model could include:
```
ctwarog-ross:
```suggestion
According to the ST2138-a specification and enforced by the [validation logic](../tools/codegen/index.js#L80-L167), a valid device model can only include:
```

ctwarog-ross (dismissed):
only a couple small nitpicks, could be merged as is

abrown-RV (approved):
approving to clear my approval flag, would be good if another looked over my changes after taking over for Jason

NR-RV (approved):

ctwarog-ross (approved):

PR #1167, opened 2026-01-06
REST Endpoints for GO

### LINKED ISSUE #1130: REST endpoints
Implement every endpoint catena supports in rest

nelgd:
## 📖 Description

Implement all REST endpoints for the GO SDK

## 🎯 Goal / Outcome

Have all endpoints for the GO SDK implemented and verify functionality with Postman

## ✅ Definition of Done (DoD)
- [x] connect
- [x] executeCommand
- [x] getAsset
- [x] setParam
- [x] getParam
- [x] getDevice
- [x] Testing with Postman
- [x] Fully Functional

### DIFF
--- a/sdks/go/pkg/rest/server.go
+++ b/sdks/go/pkg/rest/server.go
@@ -44,8 +44,9 @@ type SetValueHandler func(value any, slot int, fqoid string) catena.StatusResult
 // AssetHandler keeps w/r so implementations can stream content directly; they still return an HTTPResult for status.
 type AssetHandler func(w http.ResponseWriter, r *http.Request, slot int, fqoid string) catena.StatusResult
 
-// ConnectHandler handles connection requests (e.g., websocket upgrade or session init).
-type ConnectHandler func(w http.ResponseWriter, r *http.Request, slot int) catena.StatusResult
+// ConnectHandler handles connection requests (e.g., SSE stream for push updates).
+// This is a global endpoint that handles all slots.
+type ConnectHandler func(w http.ResponseWriter, r *http.Request) catena.StatusResult
 
 // ExecuteCommandHandler handles command execution requests.
 type ExecuteCommandHandler func(w http.ResponseWriter, r *http.Request, slot int, commandFqoid string, payload any) catena.StatusResult
@@ -59,7 +60,7 @@ type Server struct {
 	getValue       map[int]GetValueHandler
 	setValue       map[int]SetValueHandler
 	getAsset       map[int]AssetHandler
-	connect        map[int]ConnectHandler
+	connect        ConnectHandler
 	executeCommand map[int]ExecuteCommandHandler
 	notFound       NotFoundHandler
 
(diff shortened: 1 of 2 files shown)


nelgd:
bugbot run

sdks/go/pkg/rest/server.go
```diff
@@ -44,16 +44,25 @@ type SetValueHandler func(value any, slot int, fqoid string) catena.StatusResult
 // AssetHandler keeps w/r so implementations can stream content directly; they still return an HTTPResult for status.
 type AssetHandler func(w http.ResponseWriter, r *http.Request, slot int, fqoid string) catena.StatusResult
 
-// NotFoundHandler is called when an endpoint is requested that does not exist.
+// ConnectHandler handles connection requests (e.g., SSE stream for push updates).
+// This is a global endpoint that handles all slots.
+type ConnectHandler func(w http.ResponseWriter, r *http.Request) catena.StatusResult
```
abrown-RV, line 49: I think we should be helping implementors with streaming the results. I would normally suggest a yield return. Quick searching suggests go doesn't do yields, instead use an iterator function patters: https://boldlygo.tech/archive/2024-08-19-yield/. So the handler would get the `yeild` function and implementors would call `yeild(new value)` every time they want to push an update, then the server converts that into the sse stream.

sdks/go/pkg/rest/server.go
```diff
@@ -264,5 +312,51 @@ func (s *Server) RegisterRoutes() {
...
+				writeHTTPResult(w, res)
+				s.log.Info("Request finished", "method", "POST", "endpoint", "/command", "command", commandFqoid)
+				return
+			}
+			writeHTTPResult(w, s.DefaultExecuteCommandHandler(w, r, slot, commandFqoid, payload))
+		})
 	}
+
+	// Global Connect endpoint: GET /connect (outside slot loop)
+	s.mux.HandleFunc("/st2138-api/v1/connect", func(w http.ResponseWriter, r *http.Request) {
```
abrown-RV, line 347: Would probably be smart to start laying the foundations for all of the connection management stuff we still haven't solved in cpp

sdks/go/pkg/rest/server.go
abrown-RV: Actually wait, does Connect even need a handler? Maybe so the biz logic can decide if the client *should* be allowed to connect, but then the `PushUpdate` can just be a func on the server that the biz logic can call from anywhere.

nelgd: What would be the better approach? Keep it in the server or let the biz logic handle it?

abrown-RV: At least some of it will have to go into the server if we want it to be generic between rest and grpc. Grpc does use sse so that will need to be abstracted away.

abrown-RV: Can be resolved later.

abrown-RV (approved):

Issue #1101, opened 2025-11-07
API error messages


abrown-RV:
REST endpoints do not seem to return the message from a `catena::exception_with_status`. It only coverts the status into an HTTP error code and throws away the message. gRPC we actually send back the message.

Should we aim for consistency? Queue debate on verbosity of error messages vs security concerns!

mejohnnaylor:
return messages in `dev` suppress them in `prod`

PR #1170, opened 2026-01-08
Proto functionality in Go

### LINKED ISSUE #1156: Proto imports
Generate Catena value types from definitions in st2138 .protos

### LINKED ISSUE #1159: refactor StatusResult
Some suggestions for refactoring the `StatusResult`:

We should abstract away the http-ness of some of the ctor names (NoContent) for example. different server implementations are the responsible for mapping the StatusResult codes into their own status codes. Couple of suggestions:
- define a catena status result codes enum. It is a field on StatusResult. Servers to the mapping
- define multiple fields on the StatusResult for each server type the ctors set both http and grpc codes during init.

For the NoContent in SetValue specifically the biz logic should still just return an OK, and the handlers in the server need to convert the success into the 204.

> not totally up on go, but can the status result have private fields so that no one can monkey with it? is there a catena.StatusResult() ctor that would create an empty one? both things we should prevent if we can.
> Basically I'm wanting to make sure you can't circumvent the Convenience constructors.

> Maybe want a Exception field as well? would let creators add exception details that could at least be logged, even if it wouldn't go back to the user. Although what's our pattern, do we just let callbacks throw exceptions?
Thinking here is to add an *optional* Error param so that the biz logic can include the causing error (if there is one) for the error they are reporting. Then we can handle logging that for them passing in the error object lets us log the stack trace.

ctwarog-ross:
Closes Issue#1156

Generates Go objects from the smpte protos
Changes in devcontainer as well as Makefile
ToProto and FromProto methods implemented
toolchain container updated to use Go v1.23 for proto functionality


ctwarog-ross:
bugbot run

.devcontainer/go/postCreateCommand.sh
```diff
@@ -0,0 +1,24 @@
...
+go work edit -toolchain=go1.24.11
+go work sync
+
+sudo chown -R ${USER_NAME}:${USER_NAME} ${BUILD_TARGET}
+cd ${ROOT_DIR}/sdks/go
+go mod tidy
+go mod download
+make
+echo 'Docker setup complete, please restart the dev container!'
+
```
abrown-RV:
```suggestion

```

.devcontainer/go/postCreateCommand.sh
```diff
@@ -0,0 +1,24 @@
+set -e
+
+USER_NAME=$1
+BUILD_TARGET=$2
+ROOT_DIR=$3
+
+git submodule update --init --recursive
```
abrown-RV: maybe move this back out to the `postCreateCommand` the workflow should be doing this with the checkout action

sdks/go/pkg/rest/server.go
```diff
@@ -222,15 +222,15 @@ func (s *Server) RegisterRoutes() {
...
+				value, err := internal.ReadRequestJSON(r)
 				if err != nil {
 					s.log.Error("Invalid JSON in request", "slot", slot, "fqoid", fqoid, "error", err)
 					writeHTTPResult(w, catena.BadRequest("invalid JSON"))
 					return
 				}
 				s.log.Info("Request started", "method", r.Method, "endpoint", "/value", "fqoid", fqoid)
 				if handler, ok := s.lookupSetValue(slot); ok {
-					res := handler(value, slot, fqoid)
+					res := handler(catena.FromProto(value), slot, fqoid)
```
abrown-RV: I would suggest just passing the proto to the handler instead of converting to a go type. They already have to get to know protos since they have to return them, may as well pass them in as well. And you can drop that big switch statement in `FromProto`.

Could even simplify the handlers: for example: SetValue just gets a `SingleSetValuePayload` since that's already defined, it contains the slot, oid and value right in there. Now for REST there's a little more work since you'd have to actually *build* the `SingleSetValuePayload` out of the oid and slot from the REST path and the value out of the body, BUT on gRPC it will be simple since the request already arrives in that form.

sdks/go/Makefile
```diff
@@ -13,8 +22,41 @@ clean:
...
+		"$(PROTO_DIR)/service.proto"
+
+update-imports:
+	@echo "Updating imports to $(IMPORT_PATH)"
+	grep -RIl 'github.com/rossvideo/catena/sdks/go/protos' ../.. | xargs -r sed -i \
+		's#github.com/rossvideo/catena/sdks/go/protos#$(IMPORT_PATH)#g'
 
-.PHONY: all clean $(CMDS) run-basic-device
+clean-protos:
+	rm -rf "$(OUT_DIR)"
```
abrown-RV:
```suggestion
	rm -rf "$(OUT_DIR)"

```

abrown-RV (changes requested):
I had one big comment that I'm happy to again bump down the road with some discussion.

ctwarog-ross:
bugbot run

ctwarog-ross:
good bot

sdks/go/pkg/catena/value.go
abrown-RV, line 1: missing 3 value kinds: `UndefinedValue`, `Empty`, and `DataPayload` in both `toProto` and `fromProto`

sdks/go/pkg/catena/value.go
```diff
@@ -1,6 +1,143 @@
...
+// ToProto converts native Go types to protos.Value
+func ToProto(v any) *protos.Value {
+	switch val := v.(type) {
+	case int32:
+		return &protos.Value{Kind: &protos.Value_Int32Value{Int32Value: val}}
+	case int:
+		if val > int(math.MaxInt32) || val < int(math.MinInt32) {
+			return nil
+		}
+		return &protos.Value{Kind: &protos.Value_Int32Value{Int32Value: int32(val)}}
```
abrown-RV: is there a more generic way to handle these types that all map to int32? quick googling shows go has far more int and uint types that could be mapped into our int32.

Philosophical question: do we truncate, error, or nil?

sdks/go/pkg/catena/status_result.go
```diff
@@ -2,18 +2,22 @@ package catena
...
+	Status  int           // e.g., 200, 204, 400, 404, 405, 501, 500
+	Payload *protos.Value // optional body to JSON-encode on success (2xx). If nil and Status is 204, no body.
+	Message string        // optional error message for non-2xx statuses
 }
 
 // Convenience constructors
-func OK(payload any) StatusResult { return StatusResult{Status: http.StatusOK, Payload: payload} }
-func NoContent() StatusResult     { return StatusResult{Status: http.StatusNoContent} }
+func OK(value any) StatusResult {
+	return StatusResult{Status: http.StatusOK, Payload: ToProto(value)}
```
abrown-RV: what would happen here if ToProto fails (returns nil)? for a 201 says it expects it to be nil, but what about for a 200? I don't think the server would handle it nicely.

sdks/go/pkg/catena/value.go
abrown-RV: Same goes for float and arrays

sdks/go/internal/json_utils.go
```diff
@@ -2,157 +2,55 @@ package internal
...
+	"google.golang.org/protobuf/encoding/protojson"
+
+	"github.com/rossvideo/catena/build/go/protos"
 	"github.com/rossvideo/catena/sdks/go/pkg/catena"
 )
 
-func ReadValueFromRequest(r *http.Request) (any, error) {
-	//Decode body into catena value type
-	var v any
+func ReadRequestJSON(r *http.Request) (*protos.Value, error) {
```
abrown-RV, line 17: Should this check `Content-Type: application/json`? or is it enough that `Unmarshal` wants a json string?

sdks/go/internal/json_utils.go
```diff
@@ -2,157 +2,55 @@ package internal
...
-		for _, item := range val {
-			structVariantArr = append(structVariantArr, map[string]any{
-				"struct_variant_type": item.StructVariantType,
-				"value":              item.Value,
-			})
-		}
-		res.Payload = map[string][]map[string]any{"struct_variant_array_value": structVariantArr}
-	default:
-		http.Error(w, "failed to encode JSON response: ", http.StatusInternalServerError)
+	if res.Payload == nil {
```
abrown-RV: In a bunch of places this is called directly after a `toProto` call. `toProto` will sometimes return nil if it fails to convert in any way, which will then make the response empty, which is not proper Catena. Need some more error handling, but that might come with the refactor of `StatusResult`

sdks/go/pkg/catena/status_result.go
```diff
@@ -2,18 +2,22 @@ package catena
...
 )
 
 // StatusResult represents an HTTP outcome (success or error) that the server can render uniformly.
 type StatusResult struct {
-	Status  int    // e.g., 200, 204, 400, 404, 405, 501, 500
-	Payload any    // optional body to JSON-encode on success (2xx). If nil and Status is 204, no body.
-	Message string // optional error message for non-2xx statuses
+	Status  int           // e.g., 200, 204, 400, 404, 405, 501, 500
+	Payload *protos.Value // optional body to JSON-encode on success (2xx). If nil and Status is 204, no body.
+	Message string        // optional error message for non-2xx statuses
```
abrown-RV: Should make sure we have a prod/dev flag for weather this message is actually returned. Its a security thing.

sdks/go/Makefile
```diff
@@ -13,8 +22,42 @@ clean:
...
+
```
abrown-RV:
```suggestion

```

abrown-RV (changes requested):

sdks/go/pkg/rest/server.go
abrown-RV, line 1: All of the Handlers other than Get/Set Value still need to be de-RESTified

sdks/go/pkg/catena/value.go
ctwarog-ross: The proto types are all 32. This allows the user to input types that aren't 32 specified and they get converted. Seems like there are more cases to cover and covering all of them may be a poor design choice. I think it's sounding liek the best solution is to just support int32 and err the rest?

sdks/go/internal/json_utils.go
ctwarog-ross: yes, good thinking

sdks/go/internal/json_utils.go
ctwarog-ross: what should happen when it returns nil?

sdks/go/pkg/catena/status_result.go
ctwarog-ross: what do you mean?

sdks/go/internal/json_utils.go
abrown-RV: It should get converted into an error somehow I think. I would say that there should be no circumstance where the server doesn't respond with a valid st2138 message. And an empty 200 response is never valid sy2138. So, something needs to see that nil return and make sure it gets turned into I guess in this case an error? Falls into the same discussion we were having this morning with John about casting ints.

sdks/go/pkg/catena/status_result.go
abrown-RV: It is generally considered a security leak to respond to rest requests with anything more than the standard text that goes along with the codes. However, it's also super convenient in dev to see real error messages. So if we're going to have more descriptive error messages we should only return them in dev. we have a ticket for this on the cpp sdk as well: #1101

sdks/go/pkg/catena/value.go
```diff
@@ -1,6 +1,224 @@
...
+		}
+
+		dp := &protos.DataPayload{
+			Metadata:        val.Metadata,
+			Digest:          val.Digest,
+			PayloadEncoding: encoding,
+		}
+
+		// Set either URL or Payload based on which is provided
+		if val.URL != "" {
```
abrown-RV: error if both are set?

ctwarog-ross:
bugbot run

sdks/go/internal/json_utils.go
```diff
@@ -2,162 +2,65 @@ package internal
...
-	if err := json.NewDecoder(r.Body).Decode(&v); err != nil {
-		// Empty body is valid for commands that don't require a payload
-		if err == io.EOF {
-			return nil, nil
-		}
-		return nil, err
+	// Check Content-Type header
+	contentType := r.Header.Get("Content-Type")
+	if contentType != "" && contentType != "application/json" {
+		return nil, fmt.Errorf("unsupported content type: %s, expected application/json", contentType)
```
abrown-RV: this will silently pass if the client doesn't specify content-type at all. Should enforce its presence AND its value. But only for the endpoints that want it. Not all do, GETs for example shouldn't have a content-type, cause they have no body. reference the openapi.yaml spec for which should and shouldn't check.

abrown-RV: basically this check probably can't go here.

sdks/go/pkg/rest/server.go
```diff
@@ -1,362 +1,338 @@
...
+func (s *Server) RegisterGetAssetHandler(slot int, handler GetAssetHandler) {
+	s.getAssetHandlers[slot] = handler
 }
 
 func (s *Server) RegisterConnectHandler(handler ConnectHandler) {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	s.connect = handler
-	s.log.Debug("Registered connect handler")
+	s.connectHandler = handler
```
abrown-RV, line 155: don't need mutex anymore?

sdks/go/pkg/rest/server.go
```diff
@@ -1,362 +1,338 @@
...
+func (s *Server) handleValueEndpoint(w http.ResponseWriter, r *http.Request, slot int, pathParts []string) {
+	fqoid := strings.Join(pathParts, "/")
+
+	switch r.Method {
+	case http.MethodGet:
+		handler := s.lookupGetValue(slot)
+		val, res := handler(slot, fqoid)
+		writeHTTPResult(w, val, res)
+
+	case http.MethodPut, http.MethodPatch:
```
abrown-RV: openapi.yaml only specifies PUT

sdks/go/pkg/rest/server.go
```diff
@@ -1,362 +1,338 @@
...
+	s.mux.HandleFunc("/st2138-api/v1/", func(w http.ResponseWriter, r *http.Request) {
+		parts := strings.Split(strings.Trim(r.URL.Path, "/"), "/")
+		if len(parts) < 3 {
+			val, res := catena.ReplyBadRequest("invalid path format")
+			writeHTTPResult(w, val, res)
+			return
+		}
+
+		slotStr := parts[2]
+		slot, err := strconv.Atoi(slotStr)
```
abrown-RV, line 219: check slot range max int16

sdks/go/pkg/rest/server.go
```diff
@@ -1,362 +1,338 @@
...
-		})
-
-		// Param endpoint: fqoid is the path after /value/
-		s.mux.HandleFunc(prefix+"/value/", func(w http.ResponseWriter, r *http.Request) {
-			fqoid := strings.TrimPrefix(r.URL.Path, prefix+"/value/")
-			if fqoid == "" {
-				s.log.Warn("Missing param fqoid", "method", r.Method, "endpoint", "/value")
-				writeHTTPResult(w, catena.BadRequest("missing param fqoid"))
+	// Device endpoint: GET /st2138-api/v1/{slot}
+	s.mux.HandleFunc("/st2138-api/v1/", func(w http.ResponseWriter, r *http.Request) {
```
abrown-RV, line 210: I think as the endpoint coverage of the full spec grows, we're going to want to take advantage of the pattern matching provided by mux https://pkg.go.dev/net/http#hdr-Patterns-ServeMux. Basically I would expect it to turn into something along the likes of
```go
func (s *Server) RegisterRoutes() {
  s.mux.HandleFunc("GET /st2138-api/v1/{slot}/value/{fqoid...}", func(w http.ResponseWriter, r * http.Request) {
    params := s.mux.Vars(r)
    slot := params["slot"]
    ...
  }

  s.mux.HandleFunc("PUT /st2138-api/v1/{slot}/value/{fqoid...}", func(w http.ResponseWriter, r * http.Request) {
    params := s.mux.Vars(r)
    slot := params["slot"]
    ...
  }
}
```
tough for slot handling, so idk if this is actually better, but especially if everything goes to a class function (instead of inline) it is very nice and clean to just have a list of all the endpoints in one place like this:
```go
func (s *Server) RegisterRoutes() {
  s.mux.HandleFunc("GET /st2138-api/v1/{slot}/value/{fqoid...}", HandleGetValue)
  s.mux.HandleFunc("PUT /st2138-api/v1/{slot}/value/{fqoid...}", HandlePutValue)
  s.mux.HandleFunc("GET /st2138-api/v1/{slot}", HandleDeviceRequest)
  // method is optional so vv also works, just calls for all methods
  s.mux.HandlerFunc("/st2138-api/v1/{slot}/value/{fqoid...}", HandleAsset)
}

File: sdks/go/internal/json_utils.go [hunk 17]
```diff
@@ -2,162 +2,65 @@ package internal
...
+	b, err := (protojson.MarshalOptions{
+		UseProtoNames:   true,
+		EmitUnpopulated: false,
+	}).Marshal(value)
+	if err != nil {
+		// Marshaling failed (e.g., invalid UTF-8). Don't return a 2xx with an empty body.
+		// Prefer a JSON error body so clients that always parse JSON don't explode.
+		w.WriteHeader(http.StatusInternalServerError)
+		_ = json.NewEncoder(w).Encode(map[string]string{
+			"error": "failed to marshal response payload",
```
abrown-RV, line 68: need to only send back a real error message when in dev

sdks/go/pkg/catena/value.go
```diff
@@ -1,6 +1,224 @@
...
+			dp.Kind = &protos.DataPayload_Url{Url: val.URL}
+		} else if val.Payload != nil {
+			dp.Kind = &protos.DataPayload_Payload{Payload: val.Payload}
+		}
+
+		return &protos.Value{Kind: &protos.Value_DataPayload{DataPayload: dp}}, nil
+	default:
+		return nil, fmt.Errorf("unsupported type: %T", v)
+	}
```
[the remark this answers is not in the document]
abrown-RV: bad bot, not a problem but to be fair it doesn't know that CatenaValue is actually only for GetValue endpoint

sdks/go/pkg/rest/server.go
```diff
@@ -1,362 +1,338 @@
...
-// This is a global endpoint that handles all slots.
-type ConnectHandler func(w http.ResponseWriter, r *http.Request) catena.StatusResult
-
-// ExecuteCommandHandler handles command execution requests.
-type ExecuteCommandHandler func(w http.ResponseWriter, r *http.Request, slot int, commandFqoid string, payload any) catena.StatusResult
+	// If there's an error message, write it as JSON
+	if result.Error != "" {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(httpStatus)
+		fmt.Fprintf(w, `{"error": "%s"}`, result.Error)
```
abrown-RV: should only reply with the real error message in dev

sdks/go/pkg/rest/server.go
ctwarog-ross: Can you make an issue for that, this PR has already gotten out of hand scope wise

sdks/go/pkg/rest/server.go
ctwarog-ross: could you make an issue for it

sdks/go/pkg/rest/server.go
ctwarog-ross: issue created

sdks/go/pkg/catena/value.go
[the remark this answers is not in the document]
ctwarog-ross: very baaaad bot

sdks/go/pkg/rest/server.go
abrown-RV: https://github.com/rossvideo/Catena/issues/1193

abrown-RV (approved):
let'r rip

PR #1172, opened 2026-01-09
Update issue templates

NR-RV:
this is the issue template thing


KeonF-ross (approved):
Looks good as far as I can tell.

PR #1176, opened 2026-01-09
1138 param_value_string bug fix

### LINKED ISSUE #1138: utils param_value_string missing struct array
`catena::param_value_string` in `utils.cpp` does not handle `kStructArrayValues` should output `[struct array values]`
Also does not handle `kUndefinedValue` or `KIND_NOT_SET` but those are probably less important.

KeonF-ross:
## 📖 Description

Addresses issue #1138 by modifying `catena::param_value_string`  in `utils.cpp` to handle the following value types: `kStructArrayValues`, `kUndefinedValue`, `KIND_NOT_SET`.

---

## ✅ Definition of Done (DoD)
- [x] Bug is fixed and feature works as expected
- [x] Edge cases are handled

---

## 🛠️ Implementation Notes

Added tests to `utils_test.cpp`.

---

## 🔗 Related Issues / Links

[Issue #1138](https://github.com/rossvideo/Catena/issues/1138)

### DIFF
--- a/unittests/cpp/common/tests/utils_test.cpp
+++ b/unittests/cpp/common/tests/utils_test.cpp
@@ -136,10 +136,11 @@ TEST(UtilsTest, Fmt_NoArgs) {
     EXPECT_EQ(result, "Hello, World!");
 }
 
+// PARAM_VALUE_STRING TESTS
+
 TEST(UtilsTest, Value_Param_Empty) {
     st2138::Value value;
-    st2138::Empty* emptyValue = new st2138::Empty();
-    value.set_allocated_empty_value(emptyValue);
+    value.mutable_empty_value();
     std::string result = catena::param_value_string(value);
     EXPECT_EQ(result, "");
 }
@@ -160,14 +161,15 @@ TEST(UtilsTest, Value_Param_Float32) {
 
 TEST(UtilsTest, Value_Param_String) {
     st2138::Value value;
-    value.set_string_value("test string");
+    auto str = value.mutable_string_value();
+    str->assign("test string");
     std::string result = catena::param_value_string(value);
     EXPECT_EQ(result, "test string");
 }
 
 TEST(UtilsTest, Value_Param_Struct) {
     st2138::Value value;
-    value.set_allocated_struct_value(new st2138::StructValue());
+    value.mutable_struct_value();
     std::string result = catena::param_value_string(value);
     EXPECT_EQ(result, "[struct value]");
 }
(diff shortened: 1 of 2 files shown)


ctwarog-ross:
Nice first code submission Keon, complete with unit tests. We can no longer hit the default case, so unsure what to do about that. Will get back

abrown-RV:
> Nice first code submission Keon, complete with unit tests. We can no longer hit the default case, so unsure what to do about that. Will get back

We get Nathan's permission to put in an exclude block around it.

unittests/cpp/common/tests/utils_test.cpp
```diff
@@ -225,8 +225,22 @@ TEST(UtilsTest, Value_Param_StructVariantArrayValues) {
     EXPECT_EQ(result, "[struct variant array values]");
 }
 
-TEST(UtilsTest, Value_Param_DefaultCase) {
+TEST(UtilsTest, Value_Param_StructArrayValues) {
+    st2138::Value value;
+    value.set_allocated_struct_array_values(new st2138::StructList());
```
```suggestion
    value.mutable_struct_array_values();
```
is the more "protobuf way" to do this, but I see you're just pattern matching. Can you update all the other instances as well?

abrown-RV (changes requested):
very nice, we'll get that exclude direction for you

unittests/cpp/common/tests/utils_test.cpp
KeonF-ross: I've updated most of the tests to use the `.mutable_xxx` methods. However, I wasn't able to find any for `int32`, `float32`, or `UndefinedValue`, so I left them the same. Was there some other method that you'd like me to use for those tests?

unittests/cpp/common/tests/utils_test.cpp
```diff
@@ -160,14 +161,15 @@ TEST(UtilsTest, Value_Param_Float32) {
 
 TEST(UtilsTest, Value_Param_String) {
     st2138::Value value;
-    value.set_string_value("test string");
+    auto str = value.mutable_string_value();
+    str->assign("test string");
```
```suggestion
    value.set_string_value("test string");
```
Oh shoot, sorry for the confusion, its totally cool to do it this way with strings and other primitives.

unittests/cpp/common/tests/utils_test.cpp
abrown-RV: the set_xxx functions are fine for primitives

abrown-RV:
@KeonF-ross add //GCOVR_EXCL_START and //GCOVR_EXCL_END around the default and send Nathan a message with the filename and line numbers for the exclusion. There's other examples of this in the code you can match against.

abrown-RV (approved):

Issue #1179, opened 2026-01-13
Add timestamps to message data


ctwarog-ross:
## 📖 Description

References https://github.com/SMPTE/ST-2138-10/issues/2
Implement handling in SocketReader::read for Request-Start header IF it is present in a request, where it sets the value of a new member of SocketReader to a timestamp
Also create another member of SocketReader to store RequestReceived timestamp
---

## 🎯 Goal / Outcome

SocketReader now has timestamps for when a request was sent and received. Business logic will be reworked to expose this information with the goal of being able to throw out requests that took too long to send.
---

## ✅ Definition of Done (DoD)

- [x] Feature is implemented and works as expected
- [x] Edge cases are handled
- [x] Tests are added or updated (if applicable)
- [x] Documentation is updated (if needed)
- [x] Code is reviewed and approved 

---

## 🛠️ Implementation Notes (Optional)

---

## 🔗 Related Issues / Links

---
## 🚧 Risks / Open Questions (Optional)

---

PR #1185, opened 2026-01-15
Rest timestamps #1179

KeonF-ross:
## 📖 Description

Implements handling in SocketReader::read() for Request-Start header IF it is present in a request, and stores the data in a new member. Also calculates and stores the receival time.

---

## ✅ Definition of Done (DoD)

- [x] Edge cases are handled

---

## 🛠️ Implementation Notes

- Modified SocketReader.cpp
    - Added if statement to the header loop for Request-Start
    - Added helper function for validating the value of the Request-Start header
    - Added calculation for request receival time
- Modified SocketReader.h
    - Added requestStart_ and requestReceived_ member variables and relevant getter functions
- Modified SocketReader_Test.cpp
    - Modified existing tests and functions to support Request-Start header
    - Added test for invalid Request-Start values
   
---

## 🔗 Related Issues / Links

[Catena Issue #1179](https://github.com/rossvideo/Catena/issues/1179)
[ST-2138-10 Issue #2](https://github.com/SMPTE/ST-2138-10/issues/2)

---
## 📝 Additional Context

Due to its dynamic nature, the requestReceived_ member was tested manually using the debugger and the command `curl -H "Request-Start: $(date +%s.%3N)" http://localhost:6254/st2138-api/v1/0"`. This command automatically populates the request-start header with the current time in the correct format, which was then compared to requestReceived_ to check the accuracy.

The validation of the Request-Start value only checks if it is a positive base 10 number (digits are 0-9, no more than 1 decimal point), it does not check if the value actually makes sense. If Request-Start is invalid or missing, requestStart_ will default to 0 and requestReceived_ will be recorded as normal.

The correct format for the timestamps is `<number of seconds since start of epoch>.<number of milliseconds since start of current second>`


ctwarog-ross (commented):
Good submission and nice work with all the unit tests cases, impressive. One nitpick I have though is could make a const somewhere instead of using a hardcoded 0.0 for the default value? Name them DEFAULT_REQUEST_START and DEFAULT_REQUEST_RECEIVED

abrown-RV:
bugbot run

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -27,6 +27,29 @@ static inline bool iequals_header_name(std::string_view a, std::string_view b) {
...
+    const unsigned char* ps = reinterpret_cast<const unsigned char*>(s.data());
+    const std::size_t n = s.size();
+    bool hasPeriod = false;
+
+    for(std::size_t i = 0; i < n; ++i) {
+        unsigned char cs = ps[i];
+
+        // Checks if the char is not a number or is a second period(decimal point)
+        // and if the first char is a period, invalid if so
+        if ((cs != '.' || hasPeriod) && (!isdigit(cs)) || (cs == '.' && i == 0)) {
```
abrown-RV: too complicated, break into multiple ifs please. Will also improve coverage.

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -27,6 +27,29 @@ static inline bool iequals_header_name(std::string_view a, std::string_view b) {
...
+            return false;
+        }
+        if (cs == '.') {
+            hasPeriod = true; // Decimal point found
+        }
+    }
+    // No invalid characters found, value is valid
+    return true;
```
abrown-RV, line 56: doesn't make sure there's exactly one period, the spec says it is of the form `12345.12345`. Should check with John if we want to enforce the spec that closely.

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -368,4 +375,80 @@ TEST_F(RESTSocketReaderTests, SocketReader_HeaderWithoutColonIgnored) {
...
+    const std::string fqoid = "/test/oid";
+    const bool stream = false;
+    const std::unordered_map<std::string, std::string> fields = {{"test-field-1", "1"}, {"test-field-2", "2"}};
+    const std::string jwsToken = "test-jws-token";
+    const std::string origin = "*";
+    const auto detailLevel = st2138::Device_DetailLevel_NONE;
+    const std::string language = "en";
+    const std::string jsonBody = "{test_json_body}";
+    // Test with non-number/period in value
+    writeRequestWithHeaderNames(method, slot, endpoint, fqoid, stream, fields,
```
abrown-RV: just use `writeRequest`

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -140,6 +170,12 @@ void SocketReader::read(tcp::socket& socket) {
                 detailLevel_ = dlMap.at(dl);
             }
         }
+        // Getting time request was sent
+        else if (requestStart_ == DEFAULT_REQUEST_START && iequals_header_name(name, "request-start")){
+            if (valid_start_time(value)){
+                requestStart_ = stod(value);
```
abrown-RV: may as well wrap in a try/catch just in case, would hate to find a missing case in the `valid_start_time` could take down the whole server with an uncaught exception

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -368,4 +375,80 @@ TEST_F(RESTSocketReaderTests, SocketReader_HeaderWithoutColonIgnored) {
     EXPECT_EQ(socketReader.jsonBody(), jsonBody);
     EXPECT_EQ(socketReader.authorizationEnabled(), true);
     EXPECT_EQ(socketReader.stream(), stream);
+    EXPECT_EQ(socketReader.requestStart(), std::stod(requestStart));
 }
+
+/**
+ * TEST 17 - Invalid Request-Start value is set to default(0.0)
+ */
+TEST_F(RESTSocketReaderTests, SocketReader_InvalidRequestStart) {
```
abrown-RV: can you make a successful test one as well? throughout the updated tests it's always `Request-Start: 0.0` would like to see some random other values just for kicks. Also need a test for if the header is missing entirely that it still works.

abrown-RV (changes requested):
You also need to go through and update the dates at the top of any of the files that have copywrites/@date doc strings. Also add yourself as an @author.

abrown-RV:
suggest taking a look at `strtod` (from_bytes) instead of `stod` since it doesn't throw on parse failure. You can probably get away with just trying to do the conversion and using its success/fail to do most of the validation instead of your own logic.

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -140,6 +172,15 @@ void SocketReader::read(tcp::socket& socket) {
                 detailLevel_ = dlMap.at(dl);
             }
         }
+        // Getting time request was sent
+        else if (requestStart_ == DEFAULT_REQUEST_START && iequals_header_name(name, "request-start") && valid_start_time(value)){
+            try {
+                char* end = NULL;
+                requestStart_ = strtod(value.c_str(), &end);
```
abrown-RV: this actually has a guarantee it won't throw. Apologies for the confusion, I made the comment about try/catch, and then went and read up more about alternative conversion functions and commented about that. So my bad, the try/catch is redundant here.

KeonF-ross: Removed the try/catch. Also, bubgbot had mentioned that strtod can behave differently with decimal points based on locale so I changed it to use from_chars which should be locale-independent and shouldn't throw any exceptions. Did the same for gRPC timestamps as well.

sdks/cpp/connections/REST/include/SocketReader.h
```diff
@@ -193,6 +195,16 @@ class SocketReader : public ISocketReader {
...
+    /**
+     * @brief Returns the time the client started the request formatted as,
+     * <number of seconds since start of epoch>.<number of milliseconds since start of current second>
+     */
+    const double requestStart() const override { return requestStart_; }
+    /**
+     * @brief Returns the time the client's request was received formatted as,
+     * <number of seconds since start of epoch>.<number of milliseconds since start of current second>
+     */
+    const double requestReceived() const override { return requestReceived_; }
```
abrown-RV: Add a test that just checks that RequestReceived > 0. To get coverage on this line

abrown-RV (changes requested):

abrown-RV (changes requested):
Can you put the timestamp validation into a helper function and then unit test the crap out of it there with all the weird iterations of header values currently in the SocketReader_tests

sdks/cpp/common/src/utils.cpp
```diff
@@ -191,3 +191,15 @@ std::string catena::param_value_string(const st2138::Value& value) {
...
+        if (!std::isdigit(value[0])) {
+            return;
+        }
+        double temp;
+        const char* val = value.c_str();
+        auto [ptr, ec] = std::from_chars(val, val + value.length(), temp, std::chars_format::fixed);
+        if (ec == std::errc() && ptr == val + value.length()) {
+            dest = temp;
+        }
+}
```
abrown-RV:
```suggestion
}

```

sdks/cpp/common/src/utils.cpp
```diff
@@ -191,3 +191,15 @@ std::string catena::param_value_string(const st2138::Value& value) {
             // GCOVR_EXCL_STOP
     }
 }
+
+void catena::readTimestamp(std::string& value, double& dest) {
+        if (!std::isdigit(value[0])) {
```
abrown-RV:
```suggestion
        if (value.length() == 0 || !std::isdigit(value[0])) {
```

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -140,6 +147,10 @@ void SocketReader::read(tcp::socket& socket) {
                 detailLevel_ = dlMap.at(dl);
             }
         }
+        // Getting time request was sent
+        else if (requestStart_ == DEFAULT_REQUEST_START && iequals_header_name(name, "request-start") && std::isdigit(value[0])){
```
abrown-RV: suggest just removing the `isdigit` check here now that its in the `readTimestamp`. Either that or adding a length guard on the value, but this if is already complicated enough, so I'd just remove the check.

sdks/cpp/common/src/utils.cpp
```diff
@@ -191,3 +191,15 @@ std::string catena::param_value_string(const st2138::Value& value) {
             // GCOVR_EXCL_STOP
     }
 }
+
+void catena::readTimestamp(std::string& value, double& dest) {
```
abrown-RV: can you also return true/false if successful. Don't really need it now, but just seems like a nice little QoL thing for later.

sdks/cpp/common/src/utils.cpp
```diff
@@ -191,3 +191,15 @@ std::string catena::param_value_string(const st2138::Value& value) {
             // GCOVR_EXCL_STOP
     }
 }
+
```
abrown-RV, line 194: fix indenting

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -368,4 +378,6 @@ TEST_F(RESTSocketReaderTests, SocketReader_HeaderWithoutColonIgnored) {
     EXPECT_EQ(socketReader.jsonBody(), jsonBody);
     EXPECT_EQ(socketReader.authorizationEnabled(), true);
     EXPECT_EQ(socketReader.stream(), stream);
-}
+    EXPECT_EQ(socketReader.requestStart(), std::stod(requestStart));
+    EXPECT_TRUE(socketReader.requestReceived() > 0);
+}
```
abrown-RV:
```suggestion
}

```

unittests/cpp/common/tests/utils_test.cpp
```diff
@@ -244,3 +244,54 @@ TEST(UtilsTest, Value_Param_KIND_NOT_SET) {
...
+    catena::readTimestamp(value, result);
+    EXPECT_EQ(result, 0.0);
+}
+
+TEST(UtilsTest, Read_Timestamp_Invalid_Empty) {
+    double result = 0.0;
+    std::string value = "";
+    catena::readTimestamp(value, result);
+    EXPECT_EQ(result, 0.0);
+}
```
```suggestion
}

```

abrown-RV (changes requested):
we also want to just check again that we really want to format the timestamp as `<seconds>.<millis>` so just park this until we get an answer.

unittests/cpp/common/tests/utils_test.cpp
```diff
@@ -244,3 +244,61 @@ TEST(UtilsTest, Value_Param_KIND_NOT_SET) {
...
+
+// READTIMESTAMP TESTS
+
+TEST(UtilsTest, Read_Timestamp_Valid_Normal) {
+    long result = 0;
+    bool valid;
+    std::string value = "123123";
+    valid = catena::readTimestamp(value, result);
+    EXPECT_EQ(result, 123123);
+}
```
abrown-RV, line 276: test valid as well

sdks/cpp/common/include/utils.h
```diff
@@ -33,8 +33,10 @@
 /**
  * @brief Utility functions.
  * @file utils.h
- * @copyright Copyright © 2024 Ross Video Ltd
  * @author John R. Naylor (john.naylor@rossvideo.com)
+ * @author keon.foster@rossvideo.com
+ * @date 2026/01/20
+ * @copyright Copyright © 2026 Ross Video Ltd
```
abrown-RV, line 39: update at the top of file as well please

sdks/cpp/common/include/utils.h
```diff
@@ -109,4 +111,14 @@ std::string fmt(const char* f, ...);
  */
 std::string param_value_string(const st2138::Value& value);
 
+/**
+ * @brief Converts a string to a long and stores in dest
+ * 
+ * <number of milliseconds since start of epoch>
+ * 
+ * @param value The string to convert. Entire string must be a valid positive long.
+ * @param dest The destination to store the converted value. Unchanged if value is invalid.
```
abrown-RV, line 120:
```suggestion
 * @param dest The destination to store the converted value. Unchanged if value is invalid.
 * @return true if input string is valid.
```

unittests/cpp/REST/RESTTest.h
```diff
@@ -32,7 +32,8 @@
  * @brief A parent class for REST test fixtures.
  * @author benjamin.whitten@rossvideo.com
  * @author Nelson Daniels (nelson.daniels@rossvideo.com)
- * @date 2025-11-27
+ * @author keon.foster@rossvideo.com
+ * @date 2026/01/20
  * @copyright Copyright © 2025 Ross Video Ltd
```
abrown-RV: update copyright

unittests/cpp/common/tests/utils_test.cpp
```diff
@@ -244,3 +244,61 @@ TEST(UtilsTest, Value_Param_KIND_NOT_SET) {
     std::string result = catena::param_value_string(value);
     EXPECT_EQ(result, "[kind not set]");
 }
+
+// READTIMESTAMP TESTS
+
+TEST(UtilsTest, Read_Timestamp_Valid_Normal) {
```
abrown-RV, line 251: Can you throw in some tests for very non-numbers? like:
`asd1234`
`asdfasiuhe`
`1234asdf`
make sure the prefix/suffix handling is there.

abrown-RV (changes requested):
Make sure to go and update all those copyrights, I called some of them out, but there's more. Don't forget the license block at the top of files.

unittests/cpp/common/tests/utils_test.cpp
abrown-RV:
```suggestion
    EXPECT_EQ(result, 123123123);
    EXPECT_TRUE(valid);
```
sorry I wasn't clear

abrown-RV (changes requested):

abrown-RV (changes requested):
sorry I was't clear

abrown-RV:
In the shuffle I think we lost the test that makes sure if the RequestTime header is missing that everything is ok?

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -433,3 +442,22 @@ TEST_F(RESTSocketReaderTests, SocketReader_ValidContentType) {
...
+    const std::string endpoint = "/test-call";
+    const std::string fqoid = "/test/oid";
+    const bool stream = false;
+    const std::unordered_map<std::string, std::string> fields = {{"test-field-1", "1"}, {"test-field-2", "2"}};
+    const std::string jsonBody = "";
+    std::map<std::string, std::string> headers;
+    // Testing with valid
+    writeRequestWithHeaders(method, slot, endpoint, fqoid, stream, fields, jsonBody, headers);
+    EXPECT_EQ(socketReader.requestStart(), DEFAULT_REQUEST_START);
+}
```
abrown-RV:
```suggestion
}

```

abrown-RV (approved):
nice work!

Issue #1183, opened 2026-01-15
Guard against json content-type in C++ SDK


ctwarog-ross:
## 📖 Description
ensure that json contentType == "application/json" on a request, and throw out the request otherwise.

---

## 🎯 Goal / Outcome
Error output on json request where contentType != "application/json"

---

## ✅ Definition of Done (DoD)
- [x] Feature is implemented and works as expected
- [x] Edge cases are handled
- [x] Tests are added or updated (if applicable)
- [x] Documentation is updated (if needed)
- [x] Code is reviewed and approved

---

## 🛠️ Implementation Notes (Optional)

---

## 🔗 Related Issues / Links

---
## 🚧 Risks / Open Questions (Optional)

---

PR #1188, opened 2026-01-16
guard json Content-Type #1183

KeonF-ross:
## 📖 Description

Checks `Content-Type` header of a request and throws an exception if it isn't `application/json`.

---

## ✅ Definition of Done (DoD)

- [x] Edge cases are handled

---

## 🛠️ Implementation Notes

- Modified `SocketReader.cpp`
    - Added if-statement for the `Content-Type` header inside of the header loop of `SocketReader::read()`
    - Throws exception if value of `Content-Type` is not `application/json`
- Modified `SocketReader_test.cpp`
    - Added unit test for incorrect `Content-Type` values

---

## 🔗 Related Issues / Links

[Issue #1183](https://github.com/rossvideo/Catena/issues/1183)


KeonF-ross:
bugbot run

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -369,3 +369,73 @@ TEST_F(RESTSocketReaderTests, SocketReader_HeaderWithoutColonIgnored) {
...
+    const std::string fqoid = "/test/oid";
+    const bool stream = false;
+    const std::unordered_map<std::string, std::string> fields = {{"test-field-1", "1"}, {"test-field-2", "2"}};
+    const std::string jwsToken = "test-jws-token";
+    const std::string origin = "*";
+    const auto detailLevel = st2138::Device_DetailLevel_NONE;
+    const std::string language = "en";
+    const std::string jsonBody = "";
+    // Testing with xml
+    writeRequestWithHeaderNames(method, slot, endpoint, fqoid, stream, fields,
```
abrown-RV: can probably get away with just doing `writeRequestWithHeaders(method, slot, endpoint, fqoid, stream, fields, jwsToken, "", { "Content-Type: <test>"});` not so cumbersome.

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -369,3 +369,73 @@ TEST_F(RESTSocketReaderTests, SocketReader_HeaderWithoutColonIgnored) {
...
+                                jwsToken, origin, detailLevel, language, jsonBody,
+                                "Origin",
+                                "Authorization",
+                                "Detail-Level",
+                                "Language",
+                                "Content-Length",
+                                /*extraHeaderLines=*/{"Content-Type: application/json-patch+json"});
+    EXPECT_THROW(socketReader.read(serverSocket_), catena::exception_with_status);
+    // Testing with extra parameter
+    writeRequestWithHeaderNames(method, slot, endpoint, fqoid, stream, fields,
```
abrown-RV: these last 2 are not invalid, need to be split out into their own, non-invalid test

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -369,3 +369,73 @@ TEST_F(RESTSocketReaderTests, SocketReader_HeaderWithoutColonIgnored) {
...
+                                jwsToken, origin, detailLevel, language, jsonBody,
+                                "Origin",
+                                "Authorization",
+                                "Detail-Level",
+                                "Language",
+                                "Content-Length",
+                                /*extraHeaderLines=*/{"Content-Type: aPpliCatIon/jSon"});
+    EXPECT_NO_THROW(socketReader.read(serverSocket_));
+}
```
abrown-RV:
```suggestion
}

```

abrown-RV (commented):
just some minor unit testing nit picks

unittests/cpp/REST/tests/SocketReader_test.cpp
abrown-RV: I meant the different function `writeRequestWithHeaders` instead of `writeRequestWithHeaderNames`

abrown-RV (changes requested):
was thinking about this overnight, I'm sorry the ticket wasn't clear:
- only some endpoints are content-type: application/json. Its going to be all of the POSTs/etc. all of the endpoints with bodies basically, but should check against the openapi.yaml to make sure it's consistent.
- The code as written only checks the header if present, but does nothing if it's not there.

You're going to need a way for different endpoints to either require the header or not. Might be enough for just blanket statement, all GETs do not have it, everything else does, but again verify with the openapi.yaml spec.

KeonF-ross:
Checked the openapi files and can confirm that all `requestBody`s are `application/json`, however requiring a body is not consistent within all methods. My thought is to just set a flag when reading the headers if `Content-Type` is present. The code already checks if the body exists before reading it, so if we just check the flag after that check we can throw an exception if it's false. Also, as far as I can tell, the code doesn't check if a required body is there or not. Is that handled elsewhere in the code or should I try adding that to `SocketReader` too?

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -144,9 +175,19 @@ void SocketReader::read(tcp::socket& socket) {
...
+            if (!valid_content_type(value, "application/json")) {
+                throw catena::exception_with_status("Invalid Content-Type", catena::StatusCode::INVALID_ARGUMENT);
+            }
+            hasContentType = true;
+        }
     }
     // If body exists, we need to handle leftover data and append the rest.
     if (contentLength > 0) {
+        if (!hasContentType) {
+            throw catena::exception_with_status("Content-Type missing", catena::StatusCode::INVALID_ARGUMENT);
```
abrown-RV, line 190: need a test to hit this

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -92,12 +93,13 @@ class RESTSocketReaderTests : public testing::Test, public RESTTest {
                   std::string origin,
                   st2138::Device_DetailLevel detailLevel,
                   std::string language,
+                  std::string contentType,
```
abrown-RV: I think you can just make contentType always `application/json` then you don't have have all these changes to this file.

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -32,8 +32,9 @@
  * @brief This file is for testing the SocketWriter.cpp file.
  * @author benjamin.whitten@rossvideo.com
  * @author Nelson Daniels (nelson.daniels@rossvideo.com)
- * @date 2025/11/26
- * @copyright Copyright © 2025 Ross Video Ltd
+ * @author keon.foster@rossvideo.com
+ * @date 2026/01/26
```
abrown-RV, line 36: also update copyright at the top of file

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -144,9 +175,19 @@ void SocketReader::read(tcp::socket& socket) {
...
+        else if (iequals_header_name(name, "content-type")) {
+            if (!valid_content_type(value, "application/json")) {
+                throw catena::exception_with_status("Invalid Content-Type", catena::StatusCode::INVALID_ARGUMENT);
+            }
+            hasContentType = true;
+        }
     }
     // If body exists, we need to handle leftover data and append the rest.
     if (contentLength > 0) {
+        if (!hasContentType) {
```
abrown-RV, line 189:
```suggestion
        // All request bodies must be json according to the spec, so we can just check here
        if (!hasContentType) {
```

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -27,6 +27,35 @@ static inline bool iequals_header_name(std::string_view a, std::string_view b) {
...
+        if (cs >= 'A' && cs <= 'Z') {
+            cs = static_cast<unsigned char>(cs | 0x20); // to lowercase
+            if (cs == ccontentType) continue;
+        }
+
+        // If we get here, they are not equal
+        return false;
+    }
+    // Ensure MIME type is exactly contentType
+    if (ps[n] != '\0' && ps[n] != ';') {
```
abrown-RV: the comparison to `\0` is a little fragile, according to my reading `std:string_view` doesn't guarrentee that it will end in a null string (cause its a view into another string), so better to make this check a length check.

abrown-RV (changes requested):

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -369,3 +374,88 @@ TEST_F(RESTSocketReaderTests, SocketReader_HeaderWithoutColonIgnored) {
...
+    EXPECT_THROW(socketReader.read(serverSocket_), catena::exception_with_status);
+    // Testing with missing header
+    headers["Content-Length"] = std::to_string(10);
+    writeRequestWithHeaders(method, slot, endpoint, fqoid, stream, fields, jwsToken, headers);
+    EXPECT_THROW(socketReader.read(serverSocket_), catena::exception_with_status);
+}
+
+/**
+ * TEST 18b - Body present but no Content-Type header throws exception
+ */
```
abrown-RV: I think we want this test, yes?

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -369,3 +374,88 @@ TEST_F(RESTSocketReaderTests, SocketReader_HeaderWithoutColonIgnored) {
...
+    // Testing with valid
+    writeRequestWithHeaders(method, slot, endpoint, fqoid, stream, fields, jwsToken, headers, {"Content-Type: application/json"});
+    EXPECT_NO_THROW(socketReader.read(serverSocket_));
+    // Testing with extra parameter
+    writeRequestWithHeaders(method, slot, endpoint, fqoid, stream, fields, jwsToken, headers, {"Content-Type: application/json; charset=utf-8"});
+    EXPECT_NO_THROW(socketReader.read(serverSocket_));
+    // Testing case-insensitive
+    writeRequestWithHeaders(method, slot, endpoint, fqoid, stream, fields, jwsToken, headers, {"Content-Type: aPpliCatIon/jSon"});
+    EXPECT_NO_THROW(socketReader.read(serverSocket_));
+}
```
abrown-RV, line 435: Should have a test for no content-type *and* no body which works.

abrown-RV (changes requested):
Friendly reminder to get the copyrights at the top of the files as well

unittests/cpp/REST/tests/SocketReader_test.cpp
KeonF-ross: I added it into the regular 18, just forgot to delete 18b

abrown-RV (approved):
Nice work. Especially once again the simplicity of the enforcement logic you came up with.

PR #1195, opened 2026-01-21
[GO] Detailed Error Messages

### LINKED ISSUE #1193: [GO SDK] detailed errors messages
## 📖 Description

Come up with a way for denoting dev vs prod in the service and only return detailed error messages in responses when in dev

---

## 🎯 Goal / Outcome

Only return detailed errors messages in responses from endpoints in dev

---

## ✅ Definition of Done (DoD)

---

Bigger question is how to denote the environment dev/prod. A lot of project just use an env var which makes sense here as well. We're already using env vars for logging config. Env vars are super convenient when in containers, that's the primary way of inject config, so that would be my preference. Another option would be cmd line args which is more in line with the cpp sdk, but I'd kinda like to change that about the cpp sdk... XD 

---

## 🔗 Related Issues / Links

#1101

---

---

nelgd:
### Error Message Behavior

Adds environment based error message handling to the Go SDK. 

`CATENA_ENV=prod` -> HTTP error responses return generic status text

`CATENA_ENV=dev`-> Detailed error messages are returned to aid debugging

### Unified Config

`catena.Config` struct centralizes all SDK configs including logging

### InitOptions Optional Arguments

`InitOptions` now uses the optional struct pattern, allowing it to be called with no arguments:

- `catena.InitOptions()` — uses all defaults (CATENA prefix)
- `catena.InitOptions(catena.Options{AppName: "myapp"})` — with app name
- `catena.InitOptions(catena.Options{Prefix: "MYAPP"})` — custom prefix
- `catena.InitOptions(catena.Options{Prefix: "MYAPP", AppName: "myapp"})` — both

---

## ✅ Definition of Done (DoD)

- [x] `config.go` reads `CATENA_ENV` at startup
- [x] new config file unifies logging and environment setup 
- [x] `server.go` uses the env var to determine what to return when `writeHTTPResult()` is called
- [x] `InitOptions` accepts optional `Options` struct (no empty string required)
- [x] All example files updated to use new `InitOptions` syntax

---

## 🛠️ Implementation Notes
- Added env var `CATENA_ENV` in `launch.json`
- Added env var `CATENA_PORT` -> default set to 6254
- Added `config.go` to read `CATENA_ENV` at startup
- Modified `server.go` to return detailed errors only in `dev` mode
- Added `Options` struct for optional `InitOptions` parameters
- Updated `getDevice_REST`, `getSetValue_REST`, `getAsset_REST`, `executeCommand_REST` examples

---

## 🔗 Related Issues / Links
[Detailed Error Messages](https://github.com/rossvideo/Catena/issues/1193)

Closes issue #1193

### DIFF
--- a/sdks/go/examples/executeCommand_REST/executeCommand_REST.go
+++ b/sdks/go/examples/executeCommand_REST/executeCommand_REST.go
@@ -44,7 +44,6 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
-	"strconv"
 	"sync"
 	"syscall"
 	"time"
@@ -118,15 +117,15 @@ func (c *CounterState) Increment() {
 }
 
 func main() {
-	// Parse config from environment variables with CATENA prefix
-	cfg := logger.ParseConfigWithVerbosity("CATENA")
-	cfg.AppName = "executeCommand_REST"
+	// Single unified config initialization with prefix
+	cfg := catena.ParseConfigWithVerbosity("CATENA")
+	cfg.Logger.AppName = "executeCommand_REST"
 
-	if err := logger.Init(cfg); err != nil {
-		fmt.Fprintf(os.Stderr, "failed to initialize logger: %v\n", err)
+	if err := catena.Init(cfg); err != nil {
+		fmt.Fprintf(os.Stderr, "failed to initialize catena: %v\n", err)
 		os.Exit(1)
 	}
-	defer logger.Close()
+	defer catena.Close()
 
 	// Handle signals for graceful shutdown
 	sigChan := make(chan os.Signal, 1)
@@ -137,13 +136,8 @@ func main() {
 		close(shutdownChan) // closes the channel to signal the main goroutine to shut down
 	}()
 
-	// Get port from environment variables or use default 6254
-	portStr := envOr("CATENA_PORT", "6254")
-	port, err := strconv.Atoi(portStr) // converts the string to an integer
-	if err != nil {
-		logger.Error("invalid CATENA_PORT", "error", err)
-		os.Exit(1) // exits the program with a non-zero status code
-	}
+	// Port comes from the unified config
+	port := cfg.Port
 
 	// Initialize counter
 	counter := &CounterState{}
@@ -298,11 +292,4 @@ func main() {
 	// Wait for shutdown signal
 	<-shutdownChan
 	logger.Info("Server shutdown complete")
-}
-
-func envOr(key, def string) string {
-	if v := os.Getenv(key); v != "" {
-		return v
-	}
-	return def
 }
--- a/sdks/go/examples/executeCommand_REST/executeCommand_REST.go
+++ b/sdks/go/examples/executeCommand_REST/executeCommand_REST.go
@@ -118,7 +118,7 @@ func (c *CounterState) Increment() {
 
 func main() {
 	// Initialize SDK with prefix and app name
-	cfg, err := catena.InitSDK(catena.InitOptions{AppName: "executeCommand_REST"})
+	cfg, err := catena.InitOptions(catena.Options{AppName: "executeCommand_REST"})
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "failed to initialize SDK: %v\n", err)
 		os.Exit(1)
--- a/sdks/go/examples/executeCommand_REST/executeCommand_REST.go
+++ b/sdks/go/examples/executeCommand_REST/executeCommand_REST.go
@@ -118,7 +118,7 @@ func (c *CounterState) Increment() {
 
 func main() {
 	// Initialize SDK with prefix and app name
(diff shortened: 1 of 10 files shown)


abrown-RV (changes requested):
I'd like this to work well with the logger config. It should have the same config prefix functionality as the logger config, and ideally they only have to specific the config prefix once. So maybe instead of calling logger.ConfigFromEnv, you get the logger config out of this class or something. Whatever makes sense. There's probably going to be more config options, like port and TLS stuff for example. So we should gently plan ahead for expanding our config options. It wouldn't surprise me if there's a best practices framework for doing this. Most languages have either built-in support or highly used modules.

nelgd:
> I'd like this to work well with the logger config. It should have the same config prefix functionality as the logger config, and ideally they only have to specific the config prefix once. So maybe instead of calling logger.ConfigFromEnv, you get the logger config out of this class or something. Whatever makes sense. There's probably going to be more config options, like port and TLS stuff for example. So we should gently plan ahead for expanding our config options. It wouldn't surprise me if there's a best practices framework for doing this. Most languages have either built-in support or highly used modules.

I centralized all the config stuff in `catena/config.go`. Logging config is still parsed in its own config file but is stored as a struct field in the centralized config.

Added in config option for the port with default set to 6254.

sdks/go/pkg/catena/config.go
```diff
@@ -0,0 +1,170 @@
...
+	if v := os.Getenv(prefix + "_ENV"); v != "" {
+		cfg.Env = parseEnvString(v)
+	}
+
+	// Parse port
+	if v := os.Getenv(prefix + "_PORT"); v != "" {
+		if port, err := strconv.Atoi(v); err == nil && port > 0 {
+			cfg.Port = port
+		}
+	}
```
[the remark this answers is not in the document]
nelgd: Should this be fixed?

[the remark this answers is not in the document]
abrown-RV: ya I think so, in the Init func just add it to the err.

sdks/go/examples/executeCommand_REST/executeCommand_REST.go
```diff
@@ -118,15 +117,15 @@ func (c *CounterState) Increment() {
...
-	// Parse config from environment variables with CATENA prefix
-	cfg := logger.ParseConfigWithVerbosity("CATENA")
-	cfg.AppName = "executeCommand_REST"
+	// Single unified config initialization with prefix
+	cfg := catena.ParseConfigWithVerbosity("CATENA")
+	cfg.Logger.AppName = "executeCommand_REST"
 
-	if err := logger.Init(cfg); err != nil {
-		fmt.Fprintf(os.Stderr, "failed to initialize logger: %v\n", err)
+	if err := catena.Init(cfg); err != nil {
```
abrown-RV: possible do you think to combine `Parse___` with `Init`? seems kinda cumbersome to make them always call 2 funcs back to back.

Also I don't really like `catena.Init` feels to un specific for me. What are you initing? .... catena... yes, but not actually the catena *server* which is more what comes to mind when you say init catena. At least to me, maybe we can talk about this one more offiline

sdks/go/pkg/rest/server.go
```diff
@@ -66,15 +66,21 @@ type Server struct {
...
 		w.Header().Set("Content-Type", "application/json")
 		w.WriteHeader(httpStatus)
-		json.NewEncoder(w).Encode(map[string]string{"error": result.Error})
+
+		// Only return detailed error messages in dev mode
+		if catena.IsDev() {
+			json.NewEncoder(w).Encode(map[string]string{"error": result.Error})
+		} else {
+			json.NewEncoder(w).Encode(map[string]string{"error": http.StatusText(httpStatus)})
+		}
```
abrown-RV, line 82: consider a helper `func writeHTTPError` I know there are more places where this is needed, maybe not in this PR, but coming.

sdks/go/pkg/catena/config.go
```diff
@@ -0,0 +1,170 @@
...
+
+	return cfg
+}
+
+// ParseConfigWithVerbosity parses config and applies CLI verbosity flags (-v, -vv, -vvv).
+func ParseConfigWithVerbosity(prefix string) Config {
```
abrown-RV: ParseConfigWithVerbosity should call ParseConfig so we don't have duplicated code

abrown-RV (changes requested):

sdks/go/examples/executeCommand_REST/executeCommand_REST.go
nelgd: I added in a new function that handles everything. `InitSDK()` does it all and the naming should be a bit more obvious of what it does.

abrown-RV (dismissed):
nicely done, if @ctwarog-ross is happy, I'm happy

ctwarog-ross (approved):

PR #1197, opened 2026-01-22
Catena Go types

### LINKED ISSUE #1182: De-RESTify endpoint callbacks
## 📖 Description
Decouple http objects from Go endpoints pattern-matching GetValue and SetValue

---

## 🎯 Goal / Outcome
Endpoint callbacks will be more loosely coupled with REST, adhering to coding standards and making GRPC implementation more straightforward

---

## ✅ Definition of Done (DoD)
- [ ] Edge cases are handled

---

---

## 🔗 Related Issues / Links

---

---

ctwarog-ross:
## 📖 Description
Closes issue 1151

Adds new type CatenaAsset, which is returned by asset endpoint callbacks.

---

## ✅ Definition of Done (DoD)
- [x] Edge cases are handled
- [ ] Code is reviewed and approved -->

---

## 🛠️ Implementation Notes
Implemented new type CatenaAsset:
Contains ContentType, ContentDisposition, Payload and CustomHeaders Payload is of type DataPayload, and it stores the actual asset data

Added new type CatenaDevice:
CatenaAsset and CatenaDevice now use json marshaling to validate proto
Marshals a value to json, then unmarshals to proto
json value is cached if successful otherwise set to nil
cached json value is used to populate response

Also removed http objects from handlers
Left http handler in fallback
fallback handler is http specific and it would be good to have scope of the objects Shown in ExecuteCommand example, it allows for webpages for example

---

## 🔗 Related Issues / Links

---
## 📝 Additional Context

---

## 🚧 Risks / Open Questions

---
## ✨ Updates

### DIFF
--- a/sdks/go/examples/getAsset_REST/getAsset_REST.go
+++ b/sdks/go/examples/getAsset_REST/getAsset_REST.go
@@ -39,6 +39,7 @@
 package main
 
 import (
+	"crypto/sha256"
 	"embed"
 	"fmt"
 	"io/fs"
@@ -47,7 +48,6 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -59,13 +59,6 @@ import (
 
 //go:embed static/*
 var staticFS embed.FS
-
-// Asset represents a binary asset with metadata
-type Asset struct {
-	ContentType string
-	Data        []byte
-	Filename    string
-}
 
 // Global state for graceful shutdown
 var (
@@ -124,33 +117,21 @@ func main() {
 		val, ok := assets.Load(fqoid)
 		if !ok {
 			logger.Warning("Asset not found", "slot", slot, "fqoid", fqoid)
-			return catena.ReplyAssetNotFound("asset not found: " + fqoid)
+			return catena.ReplyError[catena.CatenaAsset](catena.NOT_FOUND,"asset not found: " + fqoid)
 		}
 
-		asset := val.(Asset)
-
-		// Create DataPayload in business logic
-		payload := catena.DataPayload{
-			Data:        asset.Data,
-			ContentType: asset.ContentType,
-			Cachable:    true,
-		}
-
-		// Add filename to content-disposition if available
-		if asset.Filename != "" {
-			payload.ContentDisposition = fmt.Sprintf("attachment; filename=%q", asset.Filename)
-		}
+		payload := val.(catena.DataPayload)
 
 		// Convert DataPayload to CatenaAsset when returning
-		catenaAsset := payload.ToCatenaAsset()
+		catenaAsset := payload.ToCatenaAsset(true)
 
-		logger.Info("Asset download complete", "slot", slot, "fqoid", fqoid, "size", len(asset.Data))
-		return catena.ReplyAsset(catenaAsset)
+		logger.Info("Asset download complete", "slot", slot, "fqoid", fqoid, "size", len(payload.Payload))
+		return catena.Reply(catenaAsset)
 	})
 
 	// Not found handler
 	srv.RegisterFallbackHandler(func(w http.ResponseWriter, r *http.Request) (catena.CatenaValue, catena.StatusResult) {
-		return catena.ReplyNotFound("endpoint not found")
+		return catena.ReplyError[catena.CatenaValue](catena.NOT_FOUND,"endpoint not found")
 	})
 
 	// ==========================================================================
@@ -217,13 +198,23 @@ func loadAssetsFromEmbedded(embedFS embed.FS, root string, assets *sync.Map) {
 		// Normalize path separators for URL use
 		assetID := strings.ReplaceAll(relPath, string(filepath.Separator), "/")
 
-		asset := Asset{
-			ContentType: contentType,
-			Data:        data,
-			Filename:    filepath.Base(path),
+		// Build metadata
+		metadata := map[string]string{
+			"content-type":        contentType,
+			"content-disposition": fmt.Sprintf("attachment; filename=%q", filepath.Base(path)),
 		}
 
-		assets.Store(assetID, asset)
+		// Calculate SHA-256 digest
+		hash := sha256.Sum256(data)
+
+		// Store as DataPayload directly
+		payload := catena.DataPayload{
+			Payload:  data,
+			Metadata: metadata,
+			Digest:   hash[:],
+		}
+
+		assets.Store(assetID, payload)
 		logger.Info("Loaded asset", "id", assetID, "size", len(data), "type", contentType)
 
 		return nil
--- a/sdks/go/examples/getAsset_REST/getAsset_REST.go
+++ b/sdks/go/examples/getAsset_REST/getAsset_REST.go
@@ -117,13 +117,17 @@ func main() {
 		val, ok := assets.Load(fqoid)
 		if !ok {
 			logger.Warning("Asset not found", "slot", slot, "fqoid", fqoid)
-			return catena.ReplyError[catena.CatenaAsset](catena.NOT_FOUND,"asset not found: " + fqoid)
+			return catena.ReplyError[catena.CatenaAsset](catena.NOT_FOUND, "asset not found: "+fqoid)
 		}
 
 		payload := val.(catena.DataPayload)
 
 		// Convert DataPayload to CatenaAsset when returning
-		catenaAsset := payload.ToCatenaAsset(true)
+		catenaAsset, err := catena.ToCatenaAsset(payload, true)
+		if err != nil {
+			logger.Error("Failed to convert payload to asset", "slot", slot, "fqoid", fqoid, "error", err)
+			return catena.ReplyError[catena.CatenaAsset](catena.INTERNAL, "failed to convert asset: "+err.Error())
+		}
 
 		logger.Info("Asset download complete", "slot", slot, "fqoid", fqoid, "size", len(payload.Payload))
 		return catena.Reply(catenaAsset)
@@ -131,7 +135,7 @@ func main() {
 
 	// Not found handler
 	srv.RegisterFallbackHandler(func(w http.ResponseWriter, r *http.Request) (catena.CatenaValue, catena.StatusResult) {
-		return catena.ReplyError[catena.CatenaValue](catena.NOT_FOUND,"endpoint not found")
+		return catena.ReplyError[catena.CatenaValue](catena.NOT_FOUND, "endpoint not found")
 	})
 
 	// ==========================================================================
@@ -200,8 +204,8 @@ func loadAssetsFromEmbedded(embedFS embed.FS, root string, assets *sync.Map) {
 
 		// Build metadata
 		metadata := map[string]string{
-			"content-type":        contentType,
-			"content-disposition": fmt.Sprintf("attachment; filename=%q", filepath.Base(path)),
+			"content-type": contentType,
+			"file-name":    filepath.Base(path),
 		}
 
 		// Calculate SHA-256 digest
--- a/sdks/go/examples/getAsset_REST/getAsset_REST.go
+++ b/sdks/go/examples/getAsset_REST/getAsset_REST.go
@@ -98,16 +98,17 @@ func main() {
 	// Asset Storage
 	// ==========================================================================
 	assets := &sync.Map{}
-	assetsList := &sync.Map{}
 
 	// Load assets from embedded static directory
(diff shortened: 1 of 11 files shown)


abrown-RV (changes requested):
AssestRequest should return a json body with the proto
```
message ExternalObjectPayload {
  bool cachable = 1; // If false, do not cache the external object payload
  DataPayload payload = 2;
}
```
as the body

ctwarog-ross:
@cursor review

ctwarog-ross:
bugbot run

sdks/go/executeCommand_REST
abrown-RV, line 1: accidental binary commit?

sdks/go/pkg/catena/value.go
```diff
@@ -171,6 +138,9 @@ func FromProto(pv *protos.Value) any {
 		return nil
 	}
 	switch pv.GetKind().(type) {
+	case *protos.Value_DataPayload:
+		// Return the raw DataPayload proto
+		return pv.GetDataPayload()
```
abrown-RV: comment says, it returns as native go types, this isn't true anymore, shouldn't it take the DataPayload struct?

sdks/go/pkg/catena/value.go
```diff
@@ -68,6 +60,9 @@ func ToCatenaValue(v any) (CatenaValue, error) {
 // ToProto converts native Go types to protos.Value
 func ToProto(v any) (*protos.Value, error) {
 	switch val := v.(type) {
+	case *protos.DataPayload:
+		// Pass through DataPayload directly
+		return &protos.Value{Kind: &protos.Value_DataPayload{DataPayload: val}}, nil
```
abrown-RV: same point as in FromProto

sdks/go/pkg/catena/status_code.go
```diff
@@ -295,3 +295,113 @@ func ReplyWithCode(code StatusCode, msg string) (CatenaValue, StatusResult) {
...
+
+// Asset reply helper functions that return (CatenaAsset, StatusResult)
+
+// ReplyAsset returns a successful asset response with the given asset data.
+func ReplyAsset(asset CatenaAsset) (CatenaAsset, StatusResult) {
+	return asset, StatusResult{Code: OK}
+}
+
+// ReplyAssetNotFound returns a 404 Not Found error for asset requests.
+func ReplyAssetNotFound(msg string) (CatenaAsset, StatusResult) {
```
abrown-RV: This is so much repetition, surely something can be done with Generics?

sdks/go/pkg/catena/device_v2.go
abrown-RV: need to get our naming consistent, John's been saying "layer" or "level" and you're saying "version"

sdks/go/examples/getDevice_REST/getDevice_REST.go
```diff
@@ -82,52 +81,376 @@ func main() {
...
 			"subscriptions":     true,
 			"access_scopes":     []string{"st2138:mon", "st2138:op", "st2138:cfg", "st2138:adm"},
 			"default_scope":     "st2138:op",
-			"capabilities": map[string]any{
-				"max_channels":   8,
-				"supports_4k":    true,
-				"supports_hdr":   true,
-				"input_formats":  []string{"SDI", "HDMI", "IP"},
-				"output_formats": []string{"SDI", "HDMI"},
+			"params": map[string]any{
```
abrown-RV, line 89: our cpp device request always returns the current value of each param. NOW since its protobuf, we can't enforce the presence of fields. It's just up to the BL to produce whatever data they want. So technically I believe this does conform to the spec, unless it's written down somewhere that certain parts are actually required. Maybe worth a larger discussion?

sdks/go/examples/getDevice_REST/getDevice_REST.go
```diff
@@ -137,36 +460,29 @@ func main() {
...
-			}
-			slotStr := parts[2]
-			slot, err := strconv.Atoi(slotStr)
-			if err != nil {
-				logger.Error("GetDevice invalid slot", "slot", slotStr, "error", err)
-				return catena.ReplyBadRequest("invalid slot")
-			}
-
+		slot := slot
+		srv.RegisterGetDeviceHandler(slot, func() (catena.CatenaDevice, catena.StatusResult) {
```
abrown-RV, line 466: There are a few more parameters to pass in:
`DetailLevel`

and the `DeviceRequestPayload` message also has
```
string language = 2;
repeated string subscribed_oids = 4;
```
but those aren't in the openapi.yaml. I'll make an issue on st2138-a

sdks/go/examples/getDevice_REST/getDevice_REST.go
```diff
@@ -82,52 +81,376 @@ func main() {
...
+					"widget":    "DROPDOWN",
+					"read_only": false,
+				},
+				"lock_enabled": map[string]any{
+					"name": map[string]any{
+						"display_strings": map[string]string{
+							"en": "Lock Enabled",
+						},
+					},
+					"type":      int32(4), // INT32 (treating boolean as int32)
```
abrown-RV: can we get an enum for these types?

sdks/go/pkg/catena/value.go
ctwarog-ross: not used, ill remove the case

sdks/go/examples/getDevice_REST/getDevice_REST.go
ctwarog-ross: I think this is up to BL to decide

sdks/go/pkg/catena/value.go
```diff
@@ -132,34 +124,6 @@ func ToProto(v any) (*protos.Value, error) {
 		return &protos.Value{Kind: &protos.Value_StructVariantArrayValues{StructVariantArrayValues: &protos.StructVariantList{
 			StructVariants: structVariants,
 		}}}, nil
```
abrown-RV, line 151: Why remove? Value can be a DataPayload. Can you use the conversion functions in asset.go?

sdks/go/pkg/catena/asset.go
```diff
@@ -0,0 +1,120 @@
...
+type DataPayload struct {
+	Metadata        map[string]string // Information about the payload (e.g. content-type, filename)
+	Digest          []byte            // SHA-256 digest of the payload
+	PayloadEncoding int32             // 0=UNCOMPRESSED, 1=GZIP, 2=DEFLATE
+	Payload         []byte            // Embedded binary data (use this OR Url, not both)
+	Url             string            // URL to external resource (use this OR Payload, not both)
+}
+
+// ToCatenaAsset converts DataPayload to CatenaAsset
+func (dp DataPayload) ToCatenaAsset(cachable bool) CatenaAsset {
```
abrown-RV: in other cases, isn't the `ToCatena_____` a static function? the inconsistency could be confusing.

sdks/go/examples/getAsset_REST/getAsset_REST.go
```diff
@@ -223,14 +198,23 @@ func loadAssetsFromEmbedded(embedFS embed.FS, root string, assets *sync.Map, ass
...
+
+		// Calculate SHA-256 digest
+		hash := sha256.Sum256(data)
+
+		// Store as DataPayload directly
+		payload := catena.DataPayload{
+			Payload:  data,
+			Metadata: metadata,
+			Digest:   hash[:],
 		}
```
abrown-RV, line 219: would be cool for some helpers like `FromBytes` or `FromPath` to automatically create DataPayloads. Good topic for a new issue.

sdks/go/pkg/catena/asset.go
```diff
@@ -0,0 +1,120 @@
...
+		// The error will be logged by the caller
+		return CatenaAsset{asset: nil, jsonData: nil}
+	}
+
+	return catenaAsset
+}
+
+// ToCatenaAsset converts a Go map/struct to CatenaAsset
+// This allows developers to work with native Go types and convert to the protobuf format
+func ToCatenaAsset(m map[string]any) (CatenaAsset, error) {
```
abrown-RV: what is the purpose of this func can't it just go at the end of the other ToCatenaAsset?

sdks/go/pkg/catena/device.go
```diff
@@ -31,85 +31,122 @@
...
-	String string `json:"string,omitempty"`
-	// Extend with numeric, boolean, etc.
+// CatenaDevice wraps protos.Device for device model handling
+type CatenaDevice struct {
+	device   *protos.Device
+	jsonData []byte // Cached JSON representation
+}
+
+// NewCatenaDevice creates a CatenaDevice from a protos.Device
+func NewCatenaDevice(newDevice *protos.Device) CatenaDevice {
```
abrown-RV: not used?

sdks/go/pkg/catena/asset.go
```diff
@@ -0,0 +1,140 @@
...
+			"payload_encoding": dp.PayloadEncoding,
+		},
+	}
+
+	// Convert to CatenaAsset
+	catenaAsset, err := ToCatenaAsset(assetMap)
+	if err != nil {
+		// In case of error, return an empty CatenaAsset
+		// The error will be logged by the caller
+		return CatenaAsset{asset: nil, jsonData: nil}
```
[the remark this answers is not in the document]
ctwarog-ross: bad bot, if the payload is nil we set the response code to 500 at server.go:141

sdks/go/pkg/catena/device.go
```diff
@@ -31,85 +31,122 @@
...
 }
 
-// Simple in-memory implementation for an example device.
-type simpleDevice struct {
-	mu     sync.RWMutex
-	status string
-	// imagine a map of assetID -> path
-	assetRoot string
+// GetSlot returns the device slot
+func (cd CatenaDevice) GetSlot() uint32 {
```
abrown-RV: Do we need these? I think they don't fit well with the system. If the device isn't init'd then it should error somehow not do a default. User wouldn't know otherwise. I just don't think they're needed.

sdks/go/pkg/catena/status_code.go
abrown-RV, line 1: Love this <3

sdks/go/pkg/catena/device.go
ctwarog-ross: correct, removed

sdks/go/pkg/catena/value.go
```diff
@@ -68,6 +60,9 @@ func ToCatenaValue(v any) (CatenaValue, error) {
 // ToProto converts native Go types to protos.Value
 func ToProto(v any) (*protos.Value, error) {
 	switch val := v.(type) {
+	case *protos.DataPayload:
```
abrown-RV, line 85: This will expose the BL to protos. Should ask asset.go to process it.

sdks/go/pkg/catena/asset.go
```diff
@@ -0,0 +1,64 @@
...
+// ToCatenaAsset converts DataPayload to CatenaAsset by building the proto directly
+func ToCatenaAsset(dp DataPayload, cachable bool) (CatenaAsset, error) {
+	// Build the proto DataPayload directly
+	protoPayload := &protos.DataPayload{
+		Metadata:        dp.Metadata,
+		Digest:          dp.Digest,
+		PayloadEncoding: protos.DataPayload_PayloadEncoding(dp.PayloadEncoding),
+	}
+
+	// Handle oneof: either url or payload (not both)
```
abrown-RV, line 35: return error if both

sdks/go/examples/getDevice_REST/getDevice_REST.go
```diff
@@ -82,52 +80,379 @@ func main() {
...
-				"input_formats":  []string{"SDI", "HDMI", "IP"},
-				"output_formats": []string{"SDI", "HDMI"},
+			"params": map[string]any{
+				"brightness": map[string]any{
+					"name": map[string]any{
+						"display_strings": map[string]string{
+							"en": "Brightness",
+						},
+					},
+					"type": int32(4), // INT32
```
abrown-RV: should make an enum for value types

sdks/go/examples/getAsset_REST/getAsset_REST.go
abrown-RV: https://github.com/rossvideo/Catena/issues/1213

sdks/go/pkg/catena/value.go
ctwarog-ross: you asked me to put this back in... I removed it to let asset.go process it but now I put it back on request.

abrown-RV (approved):

PR #1199, opened 2026-01-22
Grpc timestamps

KeonF-ross:
## 📖 Description

Implementation of `requestStart` and `requestReceived` timestamps in gRPC. 

---

## ✅ Definition of Done (DoD)
- [x] Edge cases are handled

---

## 🛠️ Implementation Notes

- Modified `CallData.h`
    - Added `processTimestamps_()` method which handles both timestamps. Is called in child classes proceed function at the very start of processing.
    - Added `requestStart_` and `requestReceived_` variables to store timestamps, as well as getter functions
- Modified gRPC method unit tests
    - Added test case for valid and invalid request-start header values
    - Added helper function to make RPC call and return the read request-start value
 
---

## 🔗 Related Issues / Links

[REST Timestamps Issue](https://github.com/rossvideo/Catena/issues/1179)
[REST Timestamps PR](https://github.com/rossvideo/Catena/pull/1185)
[st2138-a Timestamps Issue](https://github.com/SMPTE/ST-2138-10/issues/2)

---
## 📝 Additional Context

Approached this issue the same as the REST timestamps, modifying `CallData` similar to `SockerReader`. If there are any changes that need to be made to the approach just let me know.

To implement the unit tests, an async RPC call is sent out and the currently processed `CallData` object is continuously polled until it updates the timestamps. If the polling doesn't detect an update after some time for whatever reason, `-1.0` is returned and the test will fail. Shouldn't happen often, but is worth noting. Any ideas for other implementations are welcome.

`requestReceived_` was tested manually using postman.


sdks/cpp/connections/gRPC/src/controllers/AddLanguage.cpp
```diff
@@ -68,6 +68,7 @@ void AddLanguage::proceed(bool ok) {
          * kFinish and notifying the responder once finished.
          */
         case CallStatus::kProcess:
+            processTimesatmps_();
```
ctwarog-ross: KEON NOOO the typo everywhere :(

ctwarog-ross (dismissed):

abrown-RV (changes requested):
looks like this hasn't been touched since the decision to go to longs, I won't look too hard at it yet.

unittests/cpp/common/tests/utils_test.cpp
```diff
@@ -355,3 +355,4 @@ TEST(UtilsTest, Read_Timestamp_Invalid_Empty) {
     EXPECT_EQ(result, 0);
     EXPECT_FALSE(valid);
 }
+
```
```suggestion
```

unittests/cpp/gRPC/tests/AddLanguage_test.cpp
```diff
@@ -96,6 +97,51 @@ class gRPCAddLanguageTests : public GRPCTest {
         EXPECT_TRUE(asyncCall_) << "Async handler was not created during runtime";
     }
 
+    /**
+     * Makes an async RPC and returns the requestStart value read by the handler.
+     */
+    long getRPCRequestStart(std::string expectedRequestStart) {
```
abrown-RV: Can you not just add the metadata to `clientContext_` and then just call `testRPC`? then `EXPECT testCall_->getRequestStart/Received` after `testRPC` finishes?

abrown-RV (changes requested):

unittests/cpp/gRPC/tests/AddLanguage_test.cpp
KeonF-ross: Not quite because `testCall_` would be reset by then, but you did give me the idea to try and retrieve the timestamps just before `testCall_` is reset which did simplify the testing quite a bit.

abrown-RV (changes requested):
just gotta fix some newlines at the ends of files.

abrown-RV (approved):
nice work!

Issue #1191, opened 2026-01-19
swap out stoi in SocketReader


abrown-RV:
## 📖 Description

`stoi` will throw an exception if it's fed an invalid string. So someone could probably take out a rest server right now by sending `Content-Length: Kill the server`. Should also complain about invalid `Content-Length` headers.

---

## 🎯 Goal / Outcome
more robust `content-length` header parsing.

---

## ✅ Definition of Done (DoD)

---

## 🛠️ Implementation Notes (Optional)
take a look at `std::from_bytes` or `std::strtol` for non-throwing conversions.

---

## 🔗 Related Issues / Links

---
## 🚧 Risks / Open Questions (Optional)

---

PR #1206, opened 2026-01-28
Swap Stoi in socket reader #1191

KeonF-ross:
## 📖 Description

Modify the recording of `contentLength` in `SocketReader` to not use `stoi` and improve the handling of invalid `Content-Length` headers. Also found and fixed bug with a too large, but valid Content-Length value.

---

## ✅ Definition of Done (DoD)
- [x] `stoi` swapped for better method
- [x] Invalid header values are handled
- [x] Tests are added or updated

---

## 🛠️ Implementation Notes

- Modified `SocketReader.cpp`
    - `contentLength` is now set using `from_chars` and an exception is thrown if `from_chars` fails due to an invalid header value
    - Handling of `contentLength` larger than `jsonBody_.size()` changed to fix bug
- Modified `SocketReader_test.cpp`
    - Added test for invalid `Content-Length` values

---

## 🔗 Related Issues / Links

[Issue #1191](https://github.com/rossvideo/Catena/issues/1191)

---
## 📝 Additional Context

While testing the invalid `Content-Length` values I discovered another bug. If the `Content-Length` was greater than the JSON body size, the program would completely freeze. There was already code in place to handle this scenario, it would assume that the full body hadn't been received yet and would call `boost::asio::read()` to read the rest when it comes. The problem with this is that `boost:asio::read()` blocks the program until it reads the full amount that is expected. For example, if `Content-Length` is 100 bytes but the received JSON is 50 bytes, the program will block until the remaining 50 bytes arrive. If the remaining 50 bytes never arrive, `boost::asio::read()` doesn't have a built-in timeout, causing it to block the program indefinitely. I solved it by temporarily disabling blocking and using a while loop with a custom timeout to read from the socket.

---

## 🚧 Risks / Open Questions

All tests passed, but my solution to the too big value bug was a quick draft. I'm unsure if there's any potential issues it could cause, particularly concerning setting the socket to non-blocking as well as the duration of the timeout. Any feedback on those parts is more than welcome.


sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -153,16 +153,40 @@ void SocketReader::read(tcp::socket& socket) {
...
+            } else {
+                throw catena::exception_with_status("Invalid Content-Length", catena::StatusCode::INVALID_ARGUMENT);
+            }
         }
     }
     // If body exists, we need to handle leftover data and append the rest.
     if (contentLength > 0) {
         jsonBody_ = std::string((std::istreambuf_iterator<char>(header_stream)), std::istreambuf_iterator<char>());
         if (jsonBody_.size() < contentLength) {
-            std::size_t remainingLength = contentLength - jsonBody_.size();
```
ctwarog-ross, line 205: Change it to reject request of content-length doesn't match jsonBody.size()

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -153,16 +153,40 @@ void SocketReader::read(tcp::socket& socket) {
         }
         // Getting body content-Length
         else if (contentLength == 0 && iequals_header_name(name, "content-length")) {
-            contentLength = stoi(value);
```
ctwarog-ross, line 187: change this to try catch stoi and handle the error by throwing a catena error

ctwarog-ross (changes requested):

KeonF-ross:
bugbot run

KeonF-ross:
bugbot run

KeonF-ross:
The implementation of read_with_timeout() is based on this article: https://cppalliance.org/asio/2023/01/02/Asio201Timeouts.html

abrown-RV (changes requested):
I think I'd like a face to face review for this one. Confused on some of the details.

sdks/cpp/common/include/utils.h
```diff
@@ -33,9 +33,10 @@
 /**
  * @brief Utility functions.
  * @file utils.h
+ * @copyright Copyright © 2024 Ross Video Ltd
```
NR-RV: its 2026

unittests/cpp/REST/tests/SocketReader_test.cpp
```diff
@@ -457,7 +486,240 @@ TEST_F(RESTSocketReaderTests, SocketReader_MissingRequestStart) {
...
+    socketReader.read(serverSocket_);
     EXPECT_EQ(socketReader.requestStart(), DEFAULT_REQUEST_START);
+    work.reset();
+    runner.join();
+}
+
+/*
+ * TEST 21 - Invalid Content-Length throws exception
+ */
+TEST_F(RESTSocketReaderTests, SocketReader_InvalidContentLength) {
```
NR-RV, line 503: also test a content-length that is 0  and >2^32
and also add some checks to make sure nothing is being changed, like
if the throw happens make sure that the request doesn't set any values by useing an expect_call(...).times(0)

NR-RV (changes requested):

NR-RV (dismissed):

ctwarog-ross (dismissed):
Nice work keon you applied my request on teams perfectly

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -159,7 +200,7 @@ void SocketReader::read(tcp::socket& socket) {
         }
 
         // Getting jwsToken.
-        if (service_->authorizationEnabled() && jwsToken_.empty() && iequals_header_name(name, "authorization")) {
+        if (iequals_header_name(name, "authorization") && service_->authorizationEnabled() && jwsToken_.empty()) {
```
abrown-RV: why flip the logic, this is slower when authz is disabled

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -81,7 +113,16 @@ void SocketReader::read(tcp::socket& socket) {
...
     // Reading from the socket.
     boost::asio::streambuf buffer;
-    boost::asio::read_until(socket, buffer, "\r\n\r\n");
+    auto reader = read_until_with_timeout(socket, buffer, "\r\n\r\n", timeout);
+    auto result = boost::asio::co_spawn(socket.get_executor(), std::move(reader), boost::asio::use_future);
+
+    try {
+        result.get();
+    } catch(const catena::exception_with_status& e) {
+        throw catena::exception_with_status("Timed out", catena::StatusCode::DEADLINE_EXCEEDED);
```
abrown-RV: just rethrow? there a potential for gobbling other exceptions in the future

sdks/cpp/connections/REST/src/SocketReader.cpp
```diff
@@ -195,18 +247,32 @@ void SocketReader::read(tcp::socket& socket) {
...
             jsonBody_.resize(contentLength);
-            boost::asio::read(socket, boost::asio::buffer(&jsonBody_[contentLength - remainingLength], remainingLength));
+
+            auto reader = read_with_timeout(socket, jsonBody_, start, leftover, timeout);
+            auto result = boost::asio::co_spawn(socket.get_executor(), std::move(reader), boost::asio::use_future);
+
+            try {
+                result.get();
+            } catch(const catena::exception_with_status& e) {
+                throw catena::exception_with_status("Timed out", catena::StatusCode::DEADLINE_EXCEEDED);
```
abrown-RV: rethrow

sdks/cpp/common/include/utils.h
```diff
@@ -35,7 +35,7 @@
  * @file utils.h
  * @author John R. Naylor (john.naylor@rossvideo.com)
  * @author keon.foster@rossvideo.com
- * @date 2026/01/20
+ * @date 2026/02/03
```
abrown-RV:
```suggestion
 * @date 2026/01/20
```
no changes in this file

abrown-RV (changes requested):
I still think the `DEFAULT_TIMEOUT` should be a cmake arg

abrown-RV (changes requested):
very nice, that looks right, just a really small thing, I think the cmake var and the #define should be the same name

abrown-RV (approved):

PR #1209, opened 2026-01-29
[GO] Connect Endpoint

### LINKED ISSUE #1207: [GO] Connect Endpoint
## 📖 Description
Implement the connect endpoint for the GO SDK [REST]

---

## 🎯 Goal / Outcome

---

## ✅ Definition of Done (DoD)

---

---

## 🔗 Related Issues / Links

---

---

nelgd:
## 📖 Description

Implements the Connect endpoint for the Go SDK REST server. Persistent SSE connections push real-time updates to clients when parameter values change via SetValue. Connection management is extracted into a dedicated `ConnectionQueue` class owned by the server. Each connection runs in its own goroutine, listening for setValue updates via channels. The server gracefully shuts down all connections on destroy.

---

## ✅ Definition of Done (DoD)

- [x] Connect endpoint implemented with SSE streaming
- [x] `ConnectionQueue` class owns all connection lifecycle management
- [x] Each connection runs in its own goroutine with dedicated update and done channels
- [x] Configurable maximum connection limit (default: 100)
- [x] `BroadcastUpdate(slot, oid, value)` sends value changes to all subscribers using native Go types
- [x] SSE payloads use protojson-serialized `PushUpdates` proto messages for C++ parity
- [x] Initial `slotsAdded` message sent on connection using proto format
- [x] Graceful shutdown: server iterates all connections, signals done, and waits for deregistration
- [x] New connections rejected during shutdown
- [x] Unit tests for `ConnectionQueue` and integration tests for server connect/shutdown/notification

---

### New File: `connection_queue.go`
- `Connection` struct: holds per-connection `updates` channel (`chan *protos.PushUpdates`) and `done` channel for server-initiated shutdown
- `ConnectionQueue` struct: thread-safe map of connections, `WaitGroup` for tracking active goroutines, `shuttingDown` flag to reject new registrations
- `RegisterConnection()` / `DeregisterConnection()`: lifecycle management with idempotent deregister
- `NotifySetValue()`: fans out a `*protos.PushUpdates` to every connection's channel (non-blocking with buffer of 100)
- `Shutdown()`: closes all `done` channels then waits for all goroutines to deregister
- `SetMaxConnections()` / `ConnectionCount()`: configuration and observability

### Changes to `server.go`
- Server owns a `*ConnectionQueue` instead of inline connection maps
- `BroadcastUpdate(slot int, oid string, value any)`: public API accepts native Go types, converts to proto internally via `catena.ToProto`, business logic never touches protos
- `Shutdown()`: delegates to `ConnectionQueue.Shutdown()`
- `sendSSEEvent()`: serializes `*protos.PushUpdates` with `protojson` instead of `json.Marshal`, producing the correct nested format (e.g. `{"slot":0,"value":{"oid":"counter","value":{"int32Value":19}}}`)
- Initial connect event constructs a proper `protos.PushUpdates` with `SlotsAdded` kind
- `SetMaxConnections()`: delegates to `ConnectionQueue`

### New File: `connection_queue_test.go`
- 14 unit tests: register/deregister, idempotent deregister, max connections, unlimited connections, set max, notify delivery, channel-full drop, no-connections safety, shutdown, shutdown-rejects-new, empty shutdown, concurrent register/deregister, connection count

### Changes to `server_test.go`
- 4 new integration tests: `TestServer_Shutdown`, `TestServer_Shutdown_MultipleConnections`, `TestServer_SetValue_NotifiesConnections`, `TestServer_SetValue_FailureDoesNotNotify`
- Existing connect/broadcast tests updated for new proto-based SSE format

### Changes to `getSetValue_REST.go` (example)
- SetValue handlers call `srv.BroadcastUpdate(slot, fqoid, value)` directly with plain Go types

### Connection Flow
1. Client connects to `/st2138-api/v1/connect`
2. `ConnectionQueue.RegisterConnection()` creates a `Connection` with `updates` and `done` channels
3. Server sends initial `slotsAdded` event with all populated device slots
4. Goroutine selects on three channels: `ctx.Done()` (client disconnect), `conn.done` (server shutdown), `conn.updates` (setValue notifications)
5. When business logic calls `srv.BroadcastUpdate(slot, oid, value)`, the value is converted to a proto `PushUpdates` message and fanned out to every connection's `updates` channel
6. On exit, deferred `DeregisterConnection()` removes the connection from the queue and decrements the `WaitGroup`
7. `Server.Shutdown()` closes all `done` channels and blocks until every connection has deregistered

---

## 🔗 Related Issues / Links

Closes issue #1207


nelgd:
still working on this. Need to add more unit tests and review everything

sdks/go/pkg/rest/server.go
```diff
@@ -318,20 +329,175 @@ func (s *Server) lookupGetAsset(slot int) GetAssetHandler {
...
+	// Set SSE headers
+	w.Header().Set("Content-Type", "text/event-stream")
+	w.Header().Set("Cache-Control", "no-cache")
+	w.Header().Set("Connection", "keep-alive")
+	origin := r.Header.Get("Origin")
+	if origin != "" {
+		w.Header().Set("Access-Control-Allow-Origin", origin)
+		w.Header().Set("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, OPTIONS")
+		w.Header().Set("Access-Control-Allow-Headers", "Content-Type, Authorization, Accept, Origin, X-Requested-With, Language, Detail-Level")
+		w.Header().Set("Access-Control-Allow-Credentials", "true")
```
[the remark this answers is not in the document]
abrown-RV: ya we aren't doing CORS anywhere else yet, is there a compelling reason to do it here as a special case?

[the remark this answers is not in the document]
abrown-RV: actually our c++ sdk does this for everything in the socket reader.... so maybe go should to? but for everything

sdks/go/pkg/rest/server.go
```diff
@@ -318,20 +329,175 @@ func (s *Server) lookupGetAsset(slot int) GetAssetHandler {
...
+	}
+	flusher.Flush()
+	return nil
+}
+
+// getPopulatedSlots returns the list of registered device slots
+func (s *Server) getPopulatedSlots() []int {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.slots
```
abrown-RV: can infer this from the registered handlers?

sdks/go/pkg/rest/server.go
```diff
@@ -318,20 +329,175 @@ func (s *Server) lookupGetAsset(slot int) GetAssetHandler {
...
+	}
+}
+
+// BroadcastUpdate sends an update to all connected SSE clients
+func (s *Server) BroadcastUpdate(update PushUpdate) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	for connID, conn := range s.connections {
+		// Only send to connections subscribed to this slot (or all if slot is 0)
```
abrown-RV: subscriptions don't depend on slot, subscriptions are unsupported for now, maybe think about implementing detail-level, but maybe that's a later issue or v2

sdks/go/pkg/rest/server.go
```diff
@@ -446,6 +603,15 @@ func (s *Server) handleValueEndpoint(w http.ResponseWriter, r *http.Request, slo
 		res := handler(nativeValue, slot, fqoid)
 		writeHTTPStatusResult(w, res)
 
+		// Broadcast value change to all SSE connections if successful
+		if res.Code == catena.OK || res.Code == catena.NO_CONTENT {
```
abrown-RV: not sure about this one.... maybe have it as a flag? cpp, BL has to manually emit the signal

abrown-RV (commented):
getting a little confused/complex with the slots, have some ideas maybe we can get together next week.

nelgd:
> getting a little confused/complex with the slots, have some ideas maybe we can get together next week.

Yeah good idea. I tried to get something together based on my understanding and looking at the cpp stuff but happy to go over it more next week.

sdks/go/pkg/rest/server.go
```diff
@@ -318,20 +327,193 @@ func (s *Server) lookupGetAsset(slot int) GetAssetHandler {
...
+func (s *Server) registerConnection() (int, chan PushUpdate) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// Check connection limit
+	if s.maxConnections > 0 && len(s.connections) >= s.maxConnections {
+		return -1, nil
+	}
+
+	s.nextConnID++
```
ctwarog-ross: This works until you deregister something. I think we need a better way of keeping track of this, a connectionqueue class like in cpp is my suggestion.

sdks/go/pkg/rest/server.go
```diff
@@ -72,9 +72,28 @@ type Server struct {
...
 	fallbackHandler        FallbackHandler
+
+	// SSE connection management
+	connections    map[int]*sseConnection
+	nextConnID     int
+	maxConnections int
+}
+
+// PushUpdate represents a streaming update sent to connected clients via SSE
+type PushUpdate struct {
```
ctwarog-ross: Should use the PushUpdates definition in the protos, device.pb.go

sdks/go/pkg/rest/server.go
```diff
@@ -318,20 +327,194 @@ func (s *Server) lookupGetAsset(slot int) GetAssetHandler {
...
+	initialJSON, err := json.Marshal(initialPayload)
+	if err != nil {
+		logger.Error("failed to marshal initial connect payload", "error", err)
+		return
+	}
+	if _, err := fmt.Fprintf(w, "data: %s\n\n", initialJSON); err != nil {
+		logger.Error("failed to send initial SSE event", "error", err)
+		return
+	}
+	flusher.Flush()
```
ctwarog-ross: Aren't we supposed to send a 200 reply with the number of populated slots in the body to acknowledge successful subscription?

sdks/go/pkg/rest/server.go
```diff
@@ -318,20 +327,194 @@ func (s *Server) lookupGetAsset(slot int) GetAssetHandler {
...
+
+	logger.Info("SSE Connect started", "connID", connID)
+
+	// Stream updates until client disconnects
+	ctx := r.Context()
```
ctwarog-ross, line 463: This section blocks the current thread until an error is returned by the updates go channel. This part needs to be removed, connection management to be moved to connectionqueue.

ctwarog-ross (commented):
Move the connection management to a new class connectionqueue, owned by server. Have each connection run in its own go channel. When the server is destroyed, go through each connection and unregister them before destroying. As for what the go channel for each subscription listens for, that needs to be discussed. For now start with like in c++, waiting for a setvalue, then sending that update to all subscribers.

sdks/go/pkg/rest/server_test.go
```diff
@@ -1057,3 +1099,391 @@ func TestWriteHTTPStatusResult_WithError(t *testing.T) {
 		t.Error("expected error in response")
 	}
 }
+
+// failWriter is an http.ResponseWriter that fails on Write (used to cover error paths).
+type failWriter struct {
```
ctwarog-ross, line 1104: I was going to say move this to the top until I saw it was line 1104. Made an issue for you to consolidate the test helpers into a different helper test file https://github.com/rossvideo/Catena/issues/1248

ctwarog-ross (dismissed):

sdks/go/pkg/rest/connection_queue.go
```diff
@@ -0,0 +1,162 @@
...
+		delete(cq.connections, connID)
+		cq.wg.Done()
+		logger.Info("SSE connection unregistered", "connID", connID, "remaining", len(cq.connections))
+	}
+}
+
+// NotifySetValue sends a PushUpdates message to all connected clients.
+// Called when a value changes (by client or server) to propagate
+// the update to all SSE subscribers.
+func (cq *ConnectionQueue) NotifySetValue(update *protos.PushUpdates) {
```
abrown-RV: rename to `NotifyUpdate`, its not just a setvalue

abrown-RV (changes requested):
Looking good, nicely done! Just some organization comments and there's a bunch of unresolved comments to deal with.

abrown-RV:
vscode extension deleted my comment:
> Everything `connection_queue.go` is private to the rest package so should be lowercased/unexported/private.

sdks/go/pkg/rest/server.go
abrown-RV, line 1: There's a bunch of err branches that aren't covered in the new code that we'd like to see hit.

sdks/go/pkg/rest/server.go
```diff
@@ -318,20 +307,177 @@ func (s *Server) lookupGetAsset(slot int) GetAssetHandler {
...
+				Value: protoValue,
+			},
+		},
+	}
+	s.connectionQueue.notifyUpdate(update)
+}
+
+// Shutdown gracefully shuts down the server by closing all SSE connections
+// and waiting for their goroutines to finish.
+func (s *Server) Shutdown() {
```
abrown-RV, line 345: This is a new function, so all the examples should be updated to call it I think?

abrown-RV (changes requested):

sdks/go/pkg/rest/server.go
nelgd: Added more unit tests. Coverage for the rest pkg is now at 97.2%. There are 3 blocks I wasn't able to hit

abrown-RV (approved):

PR #1249, opened 2026-02-20
Config system overhaul #1244

### LINKED ISSUE #1244: [C++] Overhaul Config system
Currently runtime configuration is handled with cmd like args (eg --port, --static_root, etc.). I'd kind of like to talk about converting this idea to env vars like we've done in the go sdk. This is more compatible with containers: you just set the env vars on the container and you're done.

Like Go, the config system should allow for a prefix that is customized by the BL when it reads the config from the environment. `{PREFIX}_PORT`, `{PREFIX}_STATIC_ROOT`

KeonF-ross:
## 📖 Description

An overhaul of the C++ runtime configuration system. Now uses the `boost::program_options` library to support both command line arguments as well as environment variables, with the command line taking priority. All of the old command line arguments are supported with the same default values. 

Environment variables require a specific, case-sensitive prefix with a trailing underscore, such as `EXAMPLES_`. They can be easily set in our environment in two ways, either directly on the container by updating `devcontainer.json` as below,
image
or on a specific debugger configuration.
image

The old way of setting the runtime configuration required the program to call `absl::ParseCommandLine()`. That has been replaced with a new function, `config::initConfigVariables()`, in `catena::common`. The function takes in the argc, argv, service type ("REST" or "GRPC") and environment variable prefix and sets all of the config variables, which can be accessed as `config::{var_name}`.

---

## ✅ Definition of Done (DoD)
- [x] Environment variables can be used for runtime configuration
- [x] Environment variables support prefixes
- [x] Command line arguments can be used for runtime configuration
- [x] Command line arguments take priority over environment variables
- [x] Tests are added and updated

---

## 🛠️ Implementation Notes

- Added `Config.h` and `Config.cpp`
    - Implements the `boost::program_options` library to read environment variables and command line arguments
    - Stores `config::` variables
- Added `Config_test.cpp`
    - Tests `config::initConfigVariables()` to ensure correct behaviour
- Removed `Shared_Flags.h` and `Shared_Flags.cpp`
- Modified `proto_interface.cmake` and `gRPC.cmake` to link `Boost::program_options`
- Modified `Logger.cpp` and `ServiceCredentials.cpp` to use new system
- Modified all examples to use new system
- Modified all unit tests to use new system
- Modified `ServiceImpl.h` in REST and GRPC
    - `ServiceConfig()` now has new constructor which sets multiple values automatically using the `config::` vars 
- Updated docstrings and copyright dates
- Updated `DevProcess.md` to use new system

---

## 🔗 Related Issues / Links

[Issue #1244](https://github.com/rossvideo/Catena/issues/1244)

---
## 📝 Additional Context

Added unit tests but I'm going to do additional manual testing using the examples just to make sure everything is working properly. Also going to add a help message, will edit the updates section when done.

---
## ✨ Updates
 `config::initConfigVariables()` now returns a `std::pair<boost::program_options::options_description, bool>`. The bool indicates whether `--help` was present or not and the `options_description` can be used to print the help message as shown below.
image

The output looks like this:
image

---


KeonF-ross:
bugbot run

sdks/cpp/common/src/Config.cpp
```diff
@@ -0,0 +1,70 @@
...
+        ("mutual_authc", po::value<bool>()->default_value(false)->implicit_value(true), "Use this to require client to authenticate")
+        ("authz", po::value<bool>()->default_value(false)->implicit_value(true), "Use OAuth token authorization")
+        ("silent", po::value<bool>()->default_value(false)->implicit_value(true), "Use this to suppress all log output.")
+        ;
+    char* home = getenv("HOME");
+    if (home != nullptr) {
+        configArgs.add_options()("certs", po::value<std::string>()->default_value("${HOME}/test_certs"), "path/to/certs/files");
+        configArgs.add_options()("static_root", po::value<std::string>()->default_value(home), "Specify the directory to search for external objects");
+    } else {
+        throw std::runtime_error("\"HOME\" environment variable not found. Required to initialize static_root and certs default values.");
```
abrown-RV: needs windows handling

sdks/cpp/connections/REST/examples/asset_request_REST/asset_request_REST.cpp
```diff
@@ -191,8 +184,16 @@ void RunRESTServer() {
 }
 
 int main(int argc, char* argv[]) {
-    absl::SetProgramUsageMessage("Runs the Catena Service");
-    absl::ParseCommandLine(argc, argv);
+    try {
+        auto help = config::initConfigVariables(argc, argv, "REST");
```
abrown-RV: Can we streamline this a little? maybe the print just alwawys happens inside initConfig... and then it just returns a bool for return 0/continue.

sdks/cpp/common/src/Config.cpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+    if (serviceType == "REST") {
+        configArgs.add_options()("port", po::value<uint16_t>()->default_value(443), "Catena REST service port");
+    } else if (serviceType == "GRPC") {
+        configArgs.add_options()("port", po::value<uint16_t>()->default_value(6254), "Catena gRPC service port");
+    }
+
+    //Read values from command line and environment variables
+    po::variables_map vars;
+    po::store(po::parse_command_line(argc, argv, configArgs), vars);
```
abrown-RV: which has priority? cmd line or env vars?

sdks/cpp/common/src/Config.cpp
```diff
@@ -0,0 +1,70 @@
...
+    po::notify(vars);
+    po::store(po::parse_environment(configArgs, prefix), vars);
+    po::notify(vars);
+
+    // help message
+    bool help = false;
+    if (vars.count("help")) {
+        help = true;
```
abrown-RV:
```suggestion
    bool help = vars.count("help") > 0;
```

unittests/cpp/common/tests/ParamWithValue_test/ParamWithFloat_test.cpp
abrown-RV, line 1: no actual changes in here

unittests/cpp/common/tests/ParamWithValue_test/ParamWithFloatArray_test.cpp
abrown-RV, line 1: no changes in here

unittests/cpp/common/tests/Config_test.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+ using namespace catena::common;
+
+ // Test fixture for Config tests
+ class ConfigTest : public testing::Test {
+    protected:
+        // Set up and tear down Google Logging
+        static void SetUpTestSuite() {
+            config::log_dir = UNITTEST_LOG_DIR;
+            Logger::init("ConnectionQueueTest");
```
abrown-RV:
```suggestion
            Logger::init("ConfigTest");
```

unittests/cpp/common/tests/Config_test.cpp
```diff
@@ -0,0 +1,413 @@
...
+            argv.clear();
+            argv.reserve(args.size() + 1);
+            for (auto& s : args) {
+                argv.push_back(s.data());
+            }
+            argv.push_back(nullptr); //Finish with null
+        }
+
+        // Set environment variables from given strings with format "NAME=VALUE"
+        void setEnvVars(std::vector<std::string> vars) {
```
abrown-RV, line 122: record which vars are set here, and then instead of having to call unsetEnvVar teardown iterates through the recorded vars and clears them. Could even store what their values were in set, and restore them in teardown

unittests/cpp/common/tests/Config_test.cpp
```diff
@@ -0,0 +1,413 @@
...
+    EXPECT_EQ(config::max_connections, 1);
+    EXPECT_EQ(config::port, 1);
+    EXPECT_EQ(config::authz, true);
+    EXPECT_EQ(config::mutual_authc, true);
+    EXPECT_EQ(config::private_ca, true);
+    EXPECT_EQ(config::silent, true);
+}
+
+/**
+ * TEST 3 - All values set through command-line
```
abrown-RV:
```suggestion
 * TEST 3 - All values set through env vars
```

unittests/cpp/common/tests/Config_test.cpp
```diff
@@ -0,0 +1,413 @@
...
+    // Save "HOME" value and delete environment variable
+    home = getenv("HOME");
+    unsetenv("HOME");
+
+    // Do call resulting in throw
+    EXPECT_THROW(config::initConfigVariables(argc, argv, "GRPC", "CONFIGTEST_"), std::exception);
+
+    // Reset "HOME"
+    setenv("HOME", home, 0);
+}
```
abrown-RV, line 478:
```suggestion
}

```

sdks/cpp/common/src/Config.cpp
```diff
@@ -0,0 +1,70 @@
...
+        configArgs.add_options()("certs", po::value<std::string>()->default_value("${HOME}/test_certs"), "path/to/certs/files");
+        configArgs.add_options()("static_root", po::value<std::string>()->default_value(home), "Specify the directory to search for external objects");
+    } else {
+        throw std::runtime_error("\"HOME\" environment variable not found. Required to initialize static_root and certs default values.");
+    }
+    if (serviceType == "REST") {
+        configArgs.add_options()("port", po::value<uint16_t>()->default_value(443), "Catena REST service port");
+    } else if (serviceType == "GRPC") {
+        configArgs.add_options()("port", po::value<uint16_t>()->default_value(6254), "Catena gRPC service port");
+    }
```
abrown-RV: handle an else here

sdks/cpp/common/src/Config.cpp
```diff
@@ -0,0 +1,70 @@
...
+    char* home = getenv("HOME");
+    if (home != nullptr) {
+        configArgs.add_options()("certs", po::value<std::string>()->default_value("${HOME}/test_certs"), "path/to/certs/files");
+        configArgs.add_options()("static_root", po::value<std::string>()->default_value(home), "Specify the directory to search for external objects");
+    } else {
+        throw std::runtime_error("\"HOME\" environment variable not found. Required to initialize static_root and certs default values.");
+    }
+    if (serviceType == "REST") {
+        configArgs.add_options()("port", po::value<uint16_t>()->default_value(443), "Catena REST service port");
+    } else if (serviceType == "GRPC") {
```
abrown-RV: elsewhere we spell it as "gRPC" should be consistent. Or better yet use an enum. Or even better just not have different default ports between the 2. That one we can bring up in the check-in

abrown-RV (changes requested):
Pretty cool Keon, good work

sdks/cpp/common/src/Config.cpp
KeonF-ross: Cmd line has priority. It goes by which is stored first, so say we added config files later on and put its store before env vars, it would be 2nd in priority after cmd line and above env vars.

abrown-RV: 👍

sdks/cpp/common/src/Config.cpp
ctwarog-ross: How do we verify windows functionality?

sdks/cpp/common/src/Config.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <boost/program_options.hpp>
+
+namespace po = boost::program_options;
+using namespace catena::common;
+
+std::pair<boost::program_options::options_description, bool> config::initConfigVariables(int argc, char* argv[], std::string serviceType, std::string prefix) {
+    //Declare options
+    po::options_description configArgs("Allowed options");
+    configArgs.add_options()
+        ("help,h", "Get help message")
```
ctwarog-ross: @abrown-RV do we want these hardcoded strings to be consts defined as CATENA_SECURE_COMMS for example?

sdks/cpp/common/src/Config.cpp
```diff
@@ -0,0 +1,70 @@
+#include <Config.h>
+#include <Device.h>
+#include <boost/program_options.hpp>
+
+namespace po = boost::program_options;
+using namespace catena::common;
+
+std::pair<boost::program_options::options_description, bool> config::initConfigVariables(int argc, char* argv[], std::string serviceType, std::string prefix) {
```
ctwarog-ross: @abrown-RV Do we want to throw errors here or would we rather handle them internally? Looking at how this is called in main() for the examples, it isn't very intuitive.

sdks/cpp/common/src/Config.cpp
abrown-RV: that... is a very good point... mainly just wanted it noted I guess, can save actioning with the rest of the windows changes

sdks/cpp/common/src/Config.cpp
abrown-RV: The main thing here is that there are multiple results from the function:
- Success (continue with execution)
- Help (stop, but return 0)
- Error (stop, but return 1)

So right now exceptions mean error, print and return 1
return help == true means print help and return 0
and help == false, just continue.

You're right that splitting the results between a return and an exception is a little inconsistent. Maybe something like it catches exceptions and prints them OR prints the help all handled within the function and then returns a tri-state enum for the 3 states. Then the BL only needs to check the response:
```cpp
auto results = config::initConfigVariables(...);
if (results == HELP) {
    return 0;
} else if (results == ERROR) {
    return 1;
}
// continue with BL
```

orrrr better yet could use a pair `<bool exit, int code>`
```cpp
const auto [exit, code] = config::initConfigVariables(...);
if (exit) {
    return code;
}
// continue with BL
```
ORRRRR
go the opposite and init is void with a special exception type
```cpp
try {
    config::initConfigVariables(...);
} catch (const ConfigException e) {
    return e.code;
}
// continue with BL
```

sdks/cpp/common/src/Config.cpp
abrown-RV: Sure, I'm pretty anti-hardcoded strings. Maybe not the most egregious cause I think they'll only ever appear here, but those sound like famous last words XD

ctwarog-ross (dismissed):

sdks/cpp/common/src/Config.cpp
```diff
@@ -0,0 +1,74 @@
...
+        //Read values from command line and environment variables
+        po::variables_map vars;
+        po::store(po::parse_command_line(argc, argv, configArgs), vars);
+        po::notify(vars);
+        po::store(po::parse_environment(configArgs, prefix), vars);
+        po::notify(vars);
+
+        // help message
+        if (vars.count(CATENA_HELP_NAME) > 0) {
+            configArgs.print(std::cout);
```
abrown-RV, line 41: make a test to hit this branch

sdks/cpp/common/src/Config.cpp
```diff
@@ -0,0 +1,74 @@
...
+            ;
+        // TODO: Add windows functionality
+        char* home = getenv("HOME");
+        if (home != nullptr) {
+            configArgs.add_options()
+            (CATENA_CERTS_NAME.c_str(), po::value<std::string>()->default_value("${HOME}/test_certs"), "path/to/certs/files")
+            (CATENA_STATIC_ROOT_NAME.c_str(), po::value<std::string>()->default_value(home), "Specify the directory to search for external objects")
+            ;
+        } else {
+            throw std::runtime_error("\"HOME\" environment variable not found. Required to initialize static_root and certs default values.");
```
abrown-RV: not entirely convinced this should be an exception, maybe only if the the value isn't given.... but at this point you can't tell if the value is set or not. Not that this is likely to happen, HOME should always be set in most linux environments. Just seems a shame for it to be theoretically possible for there to be an environment where they've set both --certs and --static_root, but then it fails to start cause HOME isn't set, but its not even needed. Don't have a good answer, these are just my thoughts.

unittests/cpp/common/tests/Config_test.cpp
```diff
@@ -0,0 +1,429 @@
...
+    EXPECT_EQ(config::default_total_array_size, 1);
+    EXPECT_EQ(config::max_connections, 1);
+    EXPECT_EQ(config::port, 1);
+    EXPECT_EQ(config::authz, true);
+    EXPECT_EQ(config::mutual_authc, true);
+    EXPECT_EQ(config::private_ca, true);
+    EXPECT_EQ(config::silent, true);
+
+    // Delete environment variables
+    // unsetEnvVars(args);
```
abrown-RV:
```suggestion
```

abrown-RV (changes requested):

sdks/cpp/common/src/Config.cpp
ctwarog-ross: If HOME isnt set, don't we have bigger problems? wouldnt ~ operator not work, and the program wouldn't compile?

abrown-RV: good point, we definitely might have other problems. Not compilation specifically cause this runtime not compile time, and it could have been compiled somewhere with a HOME so those parts of compilation are fine, but then copied to an environment for deployment that doesn't have HOME which would be a problem. But things like the default log dir use ~, but again if the user really wanted an environment without HOME they could also set the log dir to something that exists.

Service accounts often don't have HOME set, so its not impossible for catena to be installed as a daemon with it's own service account which I believe it good practice. Not how we're recommending deploying, we suggest docker, but who knows what other might require.

KeonF-ross: My only idea for that would be to handle the default value manually instead of attaching it to the option. Could try giving it no default and then if `vars.count() == 0` we check for HOME and directly set the `config::` var to the default. Either that or just get rid of the check all together.

KeonF-ross: I just pushed a draft of what I was talking about, with a few modifications so that the help message works properly.

abrown-RV: I think this is exposing that our defaults aren't great. Maybe they shouldn't depend on the home at all. I think it is just the config's responsibility to gather the values from the input environment. The actual validation for the value should fall on the code that uses it. If we really want config to validate the value, shouldn't just check ${HOME} but should scan the value, whatever it ends up being for env vars and resolve validate them right then and there. I guess now that I think of it, if we do go that route, the default can just go back to ${HOME} and then after storing all the value, can check for $s in value and attempt to resolve and fail out. But again, I'm not sure that's the config's job... but maybe it is

I think that to make this the closest to existing behaviour just make start_root's default be `getenv("HOME") ? getenv("HOME") : "/"` so that we don't accidently get a null, but it doesn't fail out. And then we can make a ticket to think deeper about these questions

ctwarog-ross (approved):

abrown-RV (approved):

abrown-RV:
we need to plan out Catena-deployments before merging. just wait for the morning

PR #1266, opened 2026-03-04
Read with timeout bug fix

KeonF-ross:
## 📖 Description

Encountered some sort of memory bug while working on the windows build. Seemed to be a use after free occurring when the REST read operation is timed out. Made changes to how objects are passed between the `SocketReader` and the `read_with_timeout` coroutine to avoid lifetime issues. Also modified error handling so that the coroutine itself no longer throws but returns an error code. Beforehand, the throws from the coroutine could be caught by the debugger which was annoying, and the change to returning error codes works nicely with the memory fix. Changes were also made and tested on the windows build.

I also found a separate bug as well while fixing it. The `catena::StatusCode::DEADLINE_EXCEEDED` was missing a mapping to a HTTP status code. This would cause the `SocketWriter` to fail and the socket to be hung up instead of returning an error message.

---

## ✅ Definition of Done (DoD)
- [x] Feature is implemented and works as expected
- [x] Edge cases are handled
- [x] Tests are added or updated (if applicable)
- [ ] Code is reviewed and approved

---

## 🛠️ Implementation Notes

- Modified `ServiceImpl.cpp`
   - The `tcp::socket` is now created as a shared pointer which is moved onto the worker thread.
   - This is done to support changes to `read_with_timeout`.
- Modified `SocketReader.cpp`
   -  Added `ReadResult` struct which contains output of `read_with_timeout`.
   -  `read_with_timeout` now returns error code instead of throwing.
   -  `read_with_timeout` creates and returns the buffer.
- Modified `SocketReader_test.cpp`
   - Updated to use shared pointer to `tcp::socket`.
- Modified `SocketWriter.h`
   - `catena::StatusCode::DEADLINE_EXCEEDED` now maps to `408`, `Request Timeout`.
---

## 🔗 Related Issues / Links

---
## 📝 Additional Context

If you have any ideas for improving/simplifying my fix feel free to share.

---
## 🚧 Risks / Open Questions

---
## ✨ Updates
<!-- (Optional) List summarizing changes made when editing the pull request>

PR #1271, opened 2026-03-10
[C++] Windows build fix up

KeonF-ross:
## 📖 Description

An extensive fix-up of the windows-build. All unit tests now pass and the examples run and perform as expected. Also includes documentation for setting up the windows build.

---

## ✅ Definition of Done (DoD)

- Added `docs/Windows.md`
    - Document containing all the necessary installations and build instructions for Catena on Windows
- Modified `CMakeLists.txt` for `common`, `gRPC`, and `REST` unit tests
    - Linking issues were causing the tests to not be found, Win32 now only links GTest::gmock_main instead of all of the     GTest files
- Modified unit tests and examples
    - Connection endpoint changed from `0.0.0.0` to `127.0.0.1` to avoid tripping the Windows firewall
- Modified `SocketReader`, `utils`, `CallData.h` and all relevant unit tests
    - Timestamps now use `long long` which is consistent across platforms
- Modified `SocketReader`
    - Changed how the timeout works to avoid memory bug caught by windows.
    - Changes have also been made to the Linux build, and currently have their own dedicated PR.
- Modified `Config.cpp` and `Config_Test.cpp`
    - Now use windows safe methods of retrieving `HOME` (now `USERPROFILE`)
- Modified `ExternalObjectRequest.cpp`
    - Now throws if file is a directory
- Modified `AssetRequest_Test.cpp`
    - Now uses different gzip digest and payload depending on platform
- Modified both `external_object_request`s `CMakeLists.txt`
    - Added support for static vcpkg triplet
- Modified `c_cpp_properties.json`
    - Now has Win32 configuration allowing Intellisense to work properly
- Modified `launch.json`
   - Added debug configurations for Unit tests, REST, and gRPC using clang
- Various files
   - Made changes to address build warnings

---

## 🔗 Related Issues / Links

[GTEST linking issue](https://github.com/google/googletest/issues/2157)
[SocketReader Timeout fix](https://github.com/rossvideo/Catena/pull/1266)

---
## 📝 Additional Context

It is important to note that `one_of_everything` examples currently do not work when Catena is built in debug. The issue seems to be caused by static initialization of the protobuf dll. Experimented with it a bit but couldn't figure out why it only happened to `one_of_everything`, only fix I've found so far is installing everything as static instead. UPDATE: Should be fixed now

Also, in Device_Test, you now need to call `getNext()` twice to retrieve an unhandled exception. Not sure as to why or if that behaviour is an issue, but it is something that I could look deeper into.

Other than that, there are still a lot of warnings while building which I'll look into fixing later.

---
## ✨ Updates

- The one_of_everything examples should now be running when built using debug and the dynamic triplet. The problem was caused by a post build command in the CMAKE: `COMMAND ${CMAKE_COMMAND} -E copy_if_different ${_vcpkg_debug_bin_dlls} ${_vcpkg_bin_dlls} $<TARGET_FILE_DIR:${TARGET}>`.


docs/Windows.md
```diff
@@ -0,0 +1,73 @@
...
+.\vcpkg install boost-system:x64-windows boost-date-time:x64-windows boost-url:x64-windows boost-asio:x64-windows boost-program-options:x64-windows curl:x64-windows zlib:x64-windows asio:x64-windows gtest:x64-windows abseil:x64-windows grpc:x64-windows protobuf:x64-windows protobuf[zlib]:x64-windows pkgconf:x64-windows --vcpkg-root .
+# If building in debug and running one_of_everything
+.\vcpkg install boost-system:x64-windows-static boost-date-time:x64-windows-static boost-url:x64-windows-static boost-asio:x64-windows-static boost-program-options:x64-windows-static curl:x64-windows-static zlib:x64-windows-static asio:x64-windows-static gtest:x64-windows-static abseil:x64-windows-static grpc:x64-windows-static protobuf:x64-windows-static protobuf[zlib]:x64-windows-static pkgconf:x64-windows-static --vcpkg-root .
+```
+
+### jwt-cpp installation
+
+Build both Debug and Release installs so Catena can use the matching one for each build type. From a Visual Studio developer prompt:
+
+```powershell
```
abrown-RV, line 36: this didn't work in Powershell. AI says its cause the .bat file executes in a sub cmd shell, so the vars that it sets don't persist back in the root PS shell. So I did it in cmd and then it was fine.

KeonF-ross: Ah, I had tried a bunch of different things while working on the windows build, namely the one_of_everything examples, and it looks like I forgot to update the doc so I'll go and fix that up right now.

KeonF-ross: Is there anything else that isn't working so far?

sdks/cpp/common/include/ISubscriptionManager.h
```diff
@@ -56,7 +56,7 @@ class ISubscriptionManager {
 
     /**
      * @brief Add an OID subscription
-     * @param oid The OID to subscribe to (can be either a unique OID like "/param" or a wildcard like "/param/*")
+     * @param oid The OID to subscribe to (can be either a unique OID like "/param" or a wildcard like "/param/ *")
```
abrown-RV: I think John would compile with this warning suppressed, instead of changing the example, cause now the example in the comment is incorrect and wouldn't work. Check with him.

sdks/cpp/common/include/Path.h
```diff
@@ -315,4 +315,4 @@ class Path {
 }  // namespace common
 }  // namespace catena
 
-catena::common::Path operator"" _path(const char* lit, std::size_t sz);
\ No newline at end of file
+catena::common::Path operator""_path(const char* lit, std::size_t sz);
```
abrown-RV:
```suggestion
catena::common::Path operator""_path(const char* lit, std::size_t sz);

```

sdks/cpp/common/src/utils.cpp
```diff
@@ -192,11 +192,11 @@ std::string catena::param_value_string(const st2138::Value& value) {
     }
 }
 
-bool catena::readTimestamp(std::string& value, long& dest) {
+bool catena::readTimestamp(std::string& value, long long& dest) {
     if (value.length() == 0 || !std::isdigit(value[0])) {
         return false;
     }
-    long temp;
+    long long temp;
```
abrown-RV: maybe use `uint64_t` so we're definitely 64bit

abrown-RV (changes requested):
binding the server to `127.0.0.1` binds it to the loopback interface. Meaning that it is not accessible from the outside world. You can connect to it with localhost, but another machine cannot connect. You said windows doesn't accept 0.0.0.0 I think? the grpc docs say to use `[::]:<port>` which is the ipv6 ANY ip... I tested going back to 0.0.0.0 and it works... so can probably just go back to that. Or can you expand more on the 0.0.0.0 not working under windows?

abrown-RV:
Can we get this set up in vscode as well? With a fully compiling system based on the instructions, I can compile from the VS developers console. Would be great to build it right from vscode.

KeonF-ross:
> binding the server to `127.0.0.1` binds it to the loopback interface. Meaning that it is not accessible from the outside world. You can connect to it with localhost, but another machine cannot connect. You said windows doesn't accept 0.0.0.0 I think? the grpc docs say to use `[::]:<port>` which is the ipv6 ANY ip... I tested going back to 0.0.0.0 and it works... so can probably just go back to that. Or can you expand more on the 0.0.0.0 not working under windows?

You're right, the examples didn't actually need that change so I reverted that. The unit tests however do need that change whenever a client attempts to connect so I just kept those to 127 on Windows.

KeonF-ross:
bugbot run

PR #1281, opened 2026-03-12
[GO] Fixed proto fields omitting default values

### LINKED ISSUE #1268: [GO] Include Protobuf Default Values
## 📖 Description

Protobuf omits fields set to their default values (e.g., `"slot": 0`) from the JSON payload when using REST. In gRPC this is a non-issue because proto-to-proto deserialization handles default values implicitly. However, REST JSON payloads do not have the same behavior — default-valued fields are silently dropped, which can lead to missing data on the receiving end.

---

## 🎯 Goal / Outcome
REST JSON payloads should include fields even when they are set to their protobuf default values (e.g., `0` for uint32, `""` for string, `false` for bool), ensuring parity with gRPC behavior.

---

## ✅ Definition of Done (DoD)
- [ ] REST JSON serialization emits protobuf default-valued fields
- [ ] Existing REST endpoints are verified to include default values in responses/requests
- [ ] Edge cases (nested messages, repeated fields, oneof with defaults) are handled
- [ ] Tests are added or updated to cover default-value inclusion in REST JSON

---

- The root cause is protobuf's default JSON marshaling behavior (`proto.Marshal` / `protojson.Marshal`), which follows the proto3 spec of omitting default values.
- `protojson.MarshalOptions{EmitDefaultValues: true}` (or the older `EmitUnpopulated`) can be used to force inclusion of default-valued fields.
- Need to audit all REST JSON serialization paths to ensure the option is applied consistently.
- Consider whether this should be a global setting or configurable per-endpoint.

---

## 🔗 Related Issues / Links

- Discovered during review of the one-of-everything REST example (`sdks/go/examples/oneOfEverything_REST/one_of_everything_REST.go`), where `"slot": 0` was missing from the device payload.

nelgd:
## 📖 Description
Overhauls JSON marshaling in the Go SDK by centralizing proto-to-JSON conversion into dedicated `rest` package helpers (`MarshalProtoJSON`, `MarshalDeviceJSON`, `MarshalAssetJSON`), adding SMPTE-compliant device JSON cleanup via fastjson, refactoring `injectJSONField` to use Go generics, and migrating `ToJSON` methods out of catena domain types. Also updates `ToCatenaDevice` to return idiomatic Go `error` instead of `StatusResult`, and adds `ConstraintType` constants.

---

## 🎯 Goal / Outcome
- All proto-to-JSON marshaling flows through shared, tested helpers in the `rest` package — eliminating scattered inline `protojson.MarshalOptions` usage.
- Device JSON output complies with the SMPTE schema: proto3 default values like `slot:0` are preserved, while empty/null fields, zero-valued schema-forbidden fields (`precision`, `max_length`, `total_length`), and `"response":false` in params are stripped.
- `injectJSONField` uses generics to enforce type safety at compile time and remove runtime error paths.
- `CatenaDevice` and `CatenaAsset` no longer carry JSON marshaling responsibilities — that's the REST layer's job.

---

## ✅ Definition of Done (DoD)
- [x] `MarshalProtoJSON`, `MarshalDeviceJSON`, `MarshalAssetJSON` helpers created in `json_utils.go`
- [x] fastjson-based device JSON cleanup (`cleanDeviceJSON`, `deleteZeroFields`, `deleteResponseFromParams`, `deleteEmptyValues`) implemented
- [x] `injectJSONField` refactored to generic `injectJSONField[T jsonPrimitive]` — no error return
- [x] `ToJSON()` removed from `CatenaDevice` and `CatenaAsset`
- [x] `ToCatenaDevice` returns `(CatenaDevice, error)` instead of `(CatenaDevice, StatusResult)`
- [x] `ConstraintType` constants added in `value.go`
- [x] All call sites updated (examples, REST server, gRPC tests)
- [x] Comprehensive tests added/migrated to `json_utils_test.go` and `server_test.go`
- [x] `fastjson` dependency added to `go.mod`

---

## 🛠️ Implementation Notes
- **fastjson** is used for in-place JSON AST manipulation (cleanup + field injection) to avoid repeated marshal/unmarshal round-trips.
- `MarshalDeviceJSON` uses `EmitUnpopulated: true` so proto3 default values (e.g. `slot:0`) survive marshaling, then a cleanup pass strips the unwanted empties. This is a deliberate two-pass approach because protojson doesn't offer per-field control.
- `injectJSONField` uses a `jsonPrimitive` type constraint (`~int | ~int8 | ... | ~string | ~bool`) with `fmt.Sprint` as a generic numeric formatter, replacing 6 individual `strconv` switch arms.
- `injectSSEFieldFunc` is instantiated as `injectJSONField[uint32]` since SSE events only inject slot numbers.
- `ToCatenaDevice` was changed from `StatusResult` to `error` return to follow Go conventions — the `jsonData` cache was also dropped since marshaling is now the REST layer's concern.

---

## 🔗 Related Issues / Links

Closes issue #1268

---


abrown-RV (changes requested):
The problem with `EmitDefaultValues: true` in the device request is that there are a bunch of default values that we don't want included. The smpte validator relies on this. So right now if you take what the device request produces it does not validate.

Things like the `param` field on parameters, if it is not a struct, the validator says there shouldn't be a `params` field on it.

In the codegen I got around this by using
```javascript
const deviceDesc = deviceDefinition.toObject(protoObj, {
    arrays: true,
    enums: String,
});
```
the js protobuf library let you make a distinction between including empty objects vs empty arrays. I'm not sure this is helpful, but maybe a datapoint.

End of the day, you need to make sure that the response from getDevice successfully validates. So I ran it in postman, took the response, put it in a file `device.test.json` and then run that through the validator: `node smpte/tools/validate.js device.test.json`

abrown-RV:
still doesn't validate, because of the `EmitDefaultValues: true` on `DeviceRequest` it will populate `params: {}` for non-struct params which breaks the schema rules

nelgd:
> still doesn't validate, because of the `EmitDefaultValues: true` on `DeviceRequest` it will populate `params: {}` for non-struct params which breaks the schema rules

Yeah, got sidetracked with other prs and never finished this up. Will get this done once I have my other prs merged

nelgd:
image

Finally got it to validate

sdks/go/pkg/catena/device.go
```diff
@@ -63,60 +64,111 @@ const (
...
-		return nil, StatusResult{Code: INTERNAL, Error: fmt.Sprintf("failed to marshal device to JSON: %v", err)}
+		return nil, err
+	}
+	return postProcessDeviceJSON(b, cd.device.GetSlot())
+}
+
+// postProcessDeviceJSON ensures all SMPTE-schema-required fields are present
+// in the Device JSON even when proto3 omits them at their default values.
+// Preserves the original field ordering from protojson.
+func postProcessDeviceJSON(data []byte, slot uint32) ([]byte, error) {
```
abrown-RV: hit all the coverage in this new func.

sdks/go/pkg/catena/device.go
```diff
@@ -63,60 +64,111 @@ const (
...
 	if cd.device == nil {
-		return nil, StatusResult{Code: OK}
-	}
-
-	if cd.jsonData != nil {
-		return cd.jsonData, StatusResult{Code: OK}
+		return nil, nil
 	}
 
 	b, err := (protojson.MarshalOptions{
```
abrown-RV: see if there's a way to marshal into maps instead of needing to waste cycles marshalling to byte and then unmarshalling to a map. And then marshalling the map to json. Would save some steps. Especially as the json gets bigger.

abrown-RV (changes requested):
There's some other stuff I need to figure out with defaults might be a bit before we can come together and nail it all down.

ctwarog-ross:
Andrew and I just chatted, we agreed you should use EmitUnpopulated and then edit the bytearray directly (no need for map conversion) to remove the params member if it is empty (and any others that we actually dont want if empty).

nelgd:
This works now but going to wait until pr #1302 is merged since that changes a lot of proto stuff

sdks/go/pkg/catena/device.go
```diff
@@ -63,60 +65,227 @@ const (
...
+	data = stripResponseFromParams(data)
+	return stripEmptyValues(data), nil
+}
 
-	if cd.jsonData != nil {
-		return cd.jsonData, StatusResult{Code: OK}
+// stripResponseFromParams removes "response":false from params but preserves
+// it in commands. Proto field ordering guarantees params (field 6) appears
+// before commands (field 8) in protojson output.
+func stripResponseFromParams(data []byte) []byte {
```
abrown-RV: I think all this json reprocessing should go into the rest package. It only matters for there. Even ToJSON can go over, and then in json utils make a shared JSON Marshaller (if applicable) All these json processing functions could be useful for other parts of rest as well, so that would be good to lay the foundation for sharing them.

abrown-RV (changes requested):
really nice to see all those unit tests!

sdks/go/pkg/rest/server.go
```diff
@@ -206,15 +199,14 @@ func (s *Server) Shutdown() {
...
-
 // sendSSEEvent writes a single SSE event to the response writer,
-// serializing the proto PushUpdates message with protojson.
+// serializing the proto PushUpdates message via MarshalProtoJSON.
 func (s *Server) sendSSEEvent(w http.ResponseWriter, flusher http.Flusher, update *protos.PushUpdates) error {
-	data, err := sseMarshaler.Marshal(update)
+	data, err := MarshalProtoJSON(update)
+	if err != nil {
+		return err
+	}
```
abrown-RV, line 214: hit these err branches please

sdks/go/pkg/rest/json_utils.go
```diff
@@ -44,12 +45,96 @@ import (
...
+
+// injectJSONField sets key:value in a JSON object using fastjson.
+func injectJSONField(data []byte, key string, value any) ([]byte, error) {
+	var p fastjson.Parser
+	v, err := p.ParseBytes(data)
+	if err != nil {
+		return nil, err
+	}
+	var a fastjson.Arena
+	switch tv := value.(type) {
```
abrown-RV: hit these other cases. Or come up with more general ways of converting so you don't need so many cases

would be awesome if instead of `value any` and returning an error you could make it so it only accepts the types it can handle (maybe with generics) and remove the need for an err return and save err checking on the calling side.

I know you can do something like `func injectJSONField[T string | int | float | bool] (data []byte, key string, value T) []byte {}` but that's not quite the full solution, so see if you can finish that off.

abrown-RV (changes requested):
really good work with fastjson!

sdks/go/pkg/rest/json_utils.go
nelgd: Ended up going with this 

```go
type jsonPrimitive interface {
    ~int | ~int8 | ~int16 | ~int32 | ~int64 |
        ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 |
        ~float32 | ~float64 | ~string | ~bool
}

func injectJSONField[T jsonPrimitive](data []byte, key string, value T) []byte
```

sdks/go/pkg/rest/json_utils.go
```diff
@@ -45,11 +46,89 @@ import (
...
+		~float32 | ~float64 | ~string | ~bool
+}
+
+// injectJSONField sets key:value in a JSON object using fastjson.
+// The type constraint ensures only JSON-compatible primitives are accepted.
+func injectJSONField[T jsonPrimitive](data []byte, key string, value T) []byte {
+	var p fastjson.Parser
+	v, err := p.ParseBytes(data)
+	if err != nil {
+		return data
```
abrown-RV, line 119: can you hit the other branches in the function with coverage, right now its only hitting the default case.

run ./scripts/run_coverage, I want to see green for this whole function

sdks/go/pkg/rest/server.go
```diff
@@ -206,18 +199,18 @@ func (s *Server) Shutdown() {
 	s.baseServer.ShutdownConnections()
 }
 
-// sseMarshaler is the protojson marshaler used for SSE events.
-var sseMarshaler = protojson.MarshalOptions{
-	EmitUnpopulated: false,
-}
+// For unit tests to override the default functions
+var marshalSSEFunc = MarshalProtoJSON
+var injectSSEFieldFunc = injectJSONField[uint32]
```
abrown-RV: unit tests don't use injectSSEFieldFunc, don't need it?

abrown-RV (changes requested):
I'd love it if you could bring up coverage especially in json_utils there's a bunch of un-hit err branches in the new code.

abrown-RV (changes requested):
some small things, mainly coverage I think

abrown-RV (approved):

Issue #1255, opened 2026-02-25
Logs grow to big


NR-RV:
## 📖 Description
if the logs are spammed they will grow tell the place running the device runs out of space

---

## 🎯 Goal / Outcome
new config
- `CATENA_LOG_LEVEL` string version of the log levels info, warning, error, etc. default info
- `CATENA_LOG_CONSOLE` bool to turn on console logging (stdout/stderr) default on
- `CATENA_LOG_FILE` bool to turn on file logging default on
- `CATENA_LOG_DIR` if file logging is on, path to dir to put log files default from the #define
- `CATENA_LOG_SIZE` number of bytes rotate files default 10mb
- `CATENA_LOG_COUNT` number of rotated files to keep before deleting. default 5
- `CATENA_LOG_MAX_SIZE` convenience option that will derive `size` and `count` by dividing `max` by 5 or 10 ignored if size or count are set. If this is set as well as count or size, should print a warning

---

## ✅ Definition of Done (DoD)

- [ ] Code is reviewed and approved

---

## 🛠️ Implementation Notes (Optional)

the args names are not set in stone so if you can thing of better names then run them by me :D
---

## 🔗 Related Issues / Links

---
## 🚧 Risks / Open Questions (Optional)

---

abrown-RV:
slight suggested change:
instead of logs on/off with a flag, have 2 flags one to turn of console logging, one to turn off file logging.

should also probably introduce a flag for the log level as well

abrown-RV:
Looked into this a little bit, after discussion with the other seniors. The answer to this is to implement log rotation, which is a solved problem... just not by absl. SO we're just going to yet again switch logging libraries, hopefully for the last time. Boost supports a "Backend" that support rotation https://www.boost.org/doc/libs/latest/libs/log/doc/html/log/detailed/sink_backends.html#log.detailed.sink_backends.text_file

So we can use this.

Also need to look at how boost does it debug logging compared to absl.

PR #1286, opened 2026-03-20
Logging overhaul #1255

KeonF-ross:
## 📖 Description

The logging system had been moved from the abseil library to the Boost.Log library. Now supports new features such as file rotation and log directory file count limits. The usage in the code will remain the same, with the LOG and VLOG macros still being present. There are new configuration options to support the new features: `log_level`, `log_console`, `log_file`, `log_size`, `log_count`, `log_max_size`, and `log_verbosity`.

---

## ✅ Definition of Done (DoD)
- [x] Edge cases are handled

---

## 🛠️ Implementation Notes

- Modified `Config`
    - Added new configuration options as well as some warnings
- Modified `Logger`
    - `Logger` now initializes a boost backend instead of abseil
    - `LOG()` and `VLOG()` have been adjusted to use boost
    - Format of logs remains the same
- Modified `Logger_test.cpp`
    - Now has new tests for the log filters
- Modified `cpp/gRPC.cmake` and `cpp/proto_interface.cmake`
    - Added `Boost::log`
    - Removed `absl`, also removed from unit tests `CMakeLists.txt`
- Modified all unit tests
    - Added necessary log configurations

---

## 🔗 Related Issues / Links

[Issue #1255](https://github.com/rossvideo/Catena/issues/1255)
[Boost.Log](https://www.boost.org/doc/libs/1_56_0/libs/log/doc/html/index.html)

---
## 📝 Additional Context

The `silent` option in the old system seems to just set the minimum severity to `error`. With the new `log_level` option it seems kind of redundant to have silent be the same as `log_level=error` so I changed it to silent all messages. If you want it to remain the same and be a slightly more convenient `log_level=error` just let me know and I can make that change easily.

I added a `log_verbosity` option which allows the verbosity level to be set to 0, 1, or 2. By the looks of it, the old abseil system decided the verbosity level based on if the project was built in `Release` or `Debug`, representing 0 and 2 respectively. The new option now allows for level 1 to be selected. The `Release` or `Debug` functionality is still present, with it now being used to determine the default value for `log_verbosity`.

Other than that, I'm pretty sure boost has some easy ways to add headers and footers, so if there's any information that we want added to the start or end of every log file then that it something that I could look into adding.

Should also mention that the library requires linking Boost::log, so the deployment will probably have to be updated similiar to how it was for program_options. The lib should be `libboost-log1.83.0`.

---
## ✨ Updates

 I didn't like the naming of the consts in the config system (CATENA_XXX, CATENA_XXX_NAME) so I changed them to be easier to understand. The const strings now follow the pattern XXX_KEY and the default values now follow the pattern XXX_DEFAULT.


sdks/cpp/common/include/Logger.h
```diff
@@ -36,44 +36,90 @@
...
 
-#include "absl/log/log.h"
-#include "absl/log/log_sink.h"
-#include "absl/log/log_sink_registry.h"
-#include "absl/log/log_entry.h"
-#include "absl/log/check.h"
-#include "absl/log/initialize.h"
-#include "absl/log/globals.h"
-#include "absl/flags/usage.h"
 #include <mutex>
```
abrown-RV, line 43: got some unused includes here, just double check.

sdks/cpp/common/include/Logger.h
```diff
@@ -36,44 +36,90 @@
...
+#include <boost/log/attributes/mutable_constant.hpp>
+#include <boost/log/expressions/keyword.hpp>
+#include <boost/log/sinks/sync_frontend.hpp>
+#include <boost/log/sinks/text_file_backend.hpp>
+#include <boost/log/sinks/text_ostream_backend.hpp>
+#include <boost/log/trivial.hpp>
+#include <boost/log/utility/manipulators/add_value.hpp>
+
+// Helper to get basename of __FILE__
+inline const char* log_basename(const char* path) {
```
abrown-RV: these helpers... they should be namespaced or prefixed to avoid collisions. Also I'd like their implementations to go down into Logger.cpp. Logger.h is included everywhere, so less chance to trigger a full project rebuild the better.

sdks/cpp/common/include/Logger.h
```diff
@@ -36,44 +36,90 @@
...
+  BOOST_LOG_TRIVIAL(severity) \
+    << boost::log::add_value("File", log_basename(__FILE__)) \
+    << boost::log::add_value("Line", __LINE__) \
+    << boost::log::add_value("KernelThreadID", kernel_thread_id()) \
+    << boost::log::add_value("RecordID", get_record_id())
+
+#define LOG(severity) \
+  LOG_IMPL(CATENA_SEV_(severity))
+
+#define VLOG(verbosity) \
```
abrown-RV: VLOG is a google thing. Going away from absl I think its a good time to align with the rest of the world and just use DEBUG and TRACE. I'd say VLOG(0) becomes DEBUG and VLOG(1,2) become TRACE. but use your judgement when changing the call sites (or let the agent use its judgement :P)

This also removes all the verbosity stuff, the struct, the config option. Simpler.

abrown-RV (changes requested):
Very nice Keon!

I'd like some tests for the rotation logic. We can trust the boost works, but could have some tests that makes sure the config logic works as expected with the different combinations of `log_count` `log_size` and `log_max_size`

sdks/cpp/connections/REST/examples/status_update_REST/status_update_REST.cpp
```diff
@@ -64,6 +64,8 @@
 #include <functional>
 #include <Logger.h>
 
+#include <boost/log/trivial.hpp>
```
abrown-RV: accidental include?

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -385,17 +385,16 @@ void NmosNode::browse_cb(
 {
     NmosNode* node = static_cast<NmosNode*>(userdata);
     switch (event) {
-        case AVAHI_BROWSER_NEW:
-            VLOG(1) << "Discovered service: " << name << " of type " << type << " in domain " << domain;
+        case AVAHI_BROWSER_NEW: {
```
abrown-RV: what's with the curly braces?

sdks/cpp/common/include/Config.h
```diff
@@ -49,40 +49,56 @@ namespace config{
...
+const bool DASHBOARD_TLS_ENABLED_DEFAULT = false;
+const bool LOG_CONSOLE_DEFAULT = true;
+const bool LOG_FILE_DEFAULT = true;
+const double LOG_SIZE_DEFAULT = 10.0;
+const int LOG_COUNT_DEFAULT = 5;
+const double LOG_MAX_SIZE_DEFAULT = 50.0;
+#ifdef NDEBUG
+const std::string LOG_LEVEL_DEFAULT = "info";
+#else
+const std::string LOG_LEVEL_DEFAULT = "trace";
```
abrown-RV, line 108: should be `debug`

unittests/cpp/REST/tests/AssetRequest_test.cpp
```diff
@@ -74,6 +74,10 @@ class RESTAssetRequestTests : public RESTEndpointTest {
     // Set up and tear down Google Logging
     static void SetUpTestSuite() {
         config::log_dir = UNITTEST_LOG_DIR;
+        config::log_file = true;
```
abrown-RV: maybe add a helper?
Is there a reason behind the stange values for size and count?

sdks/cpp/common/src/Logger.cpp
```diff
@@ -33,44 +33,159 @@
...
+
+// Helper for log_count=1, ensures only active file is in directory
+void clear_directory(const std::filesystem::path& path) {
+    for (const auto& entry : std::filesystem::directory_iterator(path)) {
+        std::filesystem::remove_all(entry.path());
+    }
+}
+
+// Helper filter for severity level
+bool catena_severity_filter(boost::log::attribute_value_set const& attrs) {
```
abrown-RV: should cache this so it doesn't do multiple string comparisons every time there's a log

sdks/cpp/common/src/Logger.cpp
```diff
@@ -33,44 +33,159 @@
...
+}
+
+// Helper for LOG() macro for log file line numbers
+unsigned int LogHelper::get_record_id() {
+    return ++current_record_id;
+}
+
+// Helper function dictating the format of a log record
+void catena_formatter(record_view const& rec, formatting_ostream &strm) {
+    strm << extract<unsigned int>("RecordID", rec)
```
abrown-RV: I think don't include the record id, its already got a changing value in the timestamp

sdks/cpp/common/include/Logger.h
```diff
@@ -36,43 +36,53 @@
...
+#define CATENA_SEV_ERROR  error
+#define CATENA_SEV_FATAL  fatal
+    
+#define LOG_IMPL(severity) \
+  BOOST_LOG_TRIVIAL(severity) \
+    << boost::log::add_value("File", LogHelper::log_basename(__FILE__)) \
+    << boost::log::add_value("Line", __LINE__) \
+    << boost::log::add_value("KernelThreadID", LogHelper::kernel_thread_id()) \
+    << boost::log::add_value("RecordID", LogHelper::get_record_id())
+
```
abrown-RV, line 77: check out the `BOOST_LOG_ATTRIBUTE_KEYWORD` for defining the record attrs instead of string keys.

sdks/cpp/common/src/Logger.cpp
```diff
@@ -33,44 +33,159 @@
...
+                file_sink->locked_backend()->set_file_collector(sinks::file::make_collector(
+                    keywords::target = config::log_dir,
+                    keywords::max_files = config::log_count - 1, // The backend doesn't count the active file, so adjust down by 1
+                    keywords::enable_final_rotation = false
+                ));
+                file_sink->locked_backend()->scan_for_files(sinks::file::scan_all);
+            } else {
+                file_sink->locked_backend()->set_target_file_name_pattern(fileName);
+                file_sink->locked_backend()->set_file_collector(nullptr);
+                clear_directory(config::log_dir);
```
abrown-RV: dangerous if there's other app logs in the same dir

sdks/cpp/common/src/Logger.cpp
```diff
@@ -33,44 +33,159 @@
...
+            // Set filtering, formatting, and file collecting
+            if (config::log_count > 1) {
+                file_sink->locked_backend()->set_file_collector(sinks::file::make_collector(
+                    keywords::target = config::log_dir,
+                    keywords::max_files = config::log_count - 1, // The backend doesn't count the active file, so adjust down by 1
+                    keywords::enable_final_rotation = false
+                ));
+                file_sink->locked_backend()->scan_for_files(sinks::file::scan_all);
+            } else {
+                file_sink->locked_backend()->set_target_file_name_pattern(fileName);
```
abrown-RV: isn't this already set with `keyworks::file_name`?

sdks/cpp/common/src/Logger.cpp
```diff
@@ -33,44 +33,159 @@
...
+        add_common_attributes();  // Timestamps
+        if (config::log_file) {
+            // Create file sink and set parameters
+            instance().file_sink_ = boost::make_shared<file_sink_t>(
+                keywords::file_name = fileName,
+                keywords::rotation_size = config::log_size * MB
+            );
+            auto& file_sink = instance().file_sink_;
+            // Set filtering, formatting, and file collecting
+            if (config::log_count > 1) {
```
abrown-RV, line 186: I don't think this if/else is needed, processing can be the same?

sdks/cpp/common/src/Config.cpp
```diff
@@ -64,30 +70,64 @@ std::pair<bool, int> config::initConfigVariables(int argc, char* argv[], std::st
...
+                #else
+                std::cout << "Defaulting to TRACE instead." << std::endl;
+                config::log_level = "TRACE"; 
+                #endif
+            }
+        }
+        if (vars.count(LOG_MAX_SIZE_KEY)) {
+            config::log_max_size = vars[LOG_MAX_SIZE_KEY].as<int>();
+            if (config::log_max_size < 10) {
+                std::cout << "WARNING: log_max_size is set below the minimum of 10. Ignoring log_max_size." << std::endl;
```
abrown-RV, line 144: can we hit these warnings in the unit tests?

unittests/cpp/common/tests/Logger_test.cpp
```diff
@@ -44,49 +44,293 @@
...
+    EXPECT_EQ(CountFiles(), 3);
+    LOG(INFO) << "ROTATE 3";
+    std::this_thread::sleep_for(std::chrono::milliseconds(750));
+    EXPECT_EQ(CountFiles(), 3); // At max, oldest file is deleted upon rotation
+
+    // Check contents of final file
+    auto file = NewestFile(config::log_dir);
+    std::string body = ReadFile(*file);
+    EXPECT_NE(body.find("ROTATE 3"), std::string::npos);
+}
```
abrown-RV, line 318: confused... after running this test (isolated) `ctest -R common.LoggerTest.Rotate$` there's *2* files left in the log_dir. Why not 3?

unittests/cpp/common/tests/Logger_test.cpp
```diff
@@ -44,49 +44,293 @@
...
+    Logger::init("LoggerTest");
+    if (CountFiles() >= 1) {
+        std::filesystem::remove_all(config::log_dir); // Clear any old files
+        std::filesystem::create_directory(config::log_dir);
+    }
+
+    // First file
+    LOG(INFO) << std::string(50, '_'); // Padding to fill up the file
+    std::this_thread::sleep_for(std::chrono::milliseconds(750));
+    EXPECT_EQ(CountFiles(), 1);
```
abrown-RV, line 285: can you do EXPECTs on the file size as well?

abrown-RV (changes requested):
Very nice Keon, looking good!

abrown-RV:
Another request, can we have a unit test that proves it works between runs? If you do a teardown and then another init that should simulate multiple runs of the application. We want to see it notice the files from past runs and just keep using them without make new ones..

KeonF-ross:
I made some changes to how the system works because I didn't like some of the code and to better address some of your comments. 

Previously, I was using `log_count` as the max files in the directory (1 active + x archived) for ALL apps, which is why the single file case would delete everything. This is also the reason that `log_count=1` has to be handled separately, as the file collector just doesn't seem to be able to handle 1 active + 0 archived, it always does at least 1 archived. I had fixed that by having a special case where no file collector is set and the target name is the active name, causing the file to rotate over onto itself.

Now, I'm using `log_count` as the max files (1 active + x archived) for that specific app. This means that when it cleans archived files on startup, only files matching the app name will be deleted. As the active file is still in the directory, `log_count=1` still needs to be handled differently, but I did change it to use its own custom file collector instead. I also changed the naming for the active file to no longer have the timestamp, which makes it easier to find the log as well as do the unit tests. 

You had asked about why there were only 2 files left in the log_dir instead of 3 after running the tests and that was because of a bug which kept `final_rotation` enabled. Since `log_count=3` is 1 active + 2 archived, a rotation on teardown would archive the active file leaving you with 0 active + 2 archived. Instead of full disabling this, I opted to just make it a feature and added a config option to enable it.

I also changed all of the unit tests, except logger, to use a helper for initializing the logging. The helper uses `log_count=3` with `log_final_rotation=true`, making 3 files during execution (current and last 2 runs) and 2 upon completion (current and the run before it). This replaces the previous `log_count=128`, which was a bit over 2 full runs of `ctest`.

You also asked about adding a unit test for using the same file from past runs. That actually wasn't a feature, so I went ahead and implemented it! Now, there is a new config option called `log_append` which toggles Append mode. Append mode will write to any old active files when turned on, and the system will just create a new one when it's turned off. The changes to the active file name pattern and `log_count=1` using a file collector were also necessary to make this mode work.

abrown-RV:
bugbot run

sdks/cpp/common/src/Logger.cpp
```diff
@@ -33,44 +33,221 @@
...
+    return p ? p + 1 : path;
+}
+
+// Helper for LOG() macro to get kernel id of thread
+int LogHelper::kernel_thread_id() {
+    return static_cast<int>(gettid());
+}
+
+// Helper to initialize logging for unit tests
+void LogHelper::set_up_test_logs(std::string directory, std::string appName) {
```
abrown-RV: this should go in the unit test utils. Ideally test helpers don't go in the main sdk

sdks/cpp/common/src/NmosNode.cpp
```diff
@@ -385,17 +385,15 @@ void NmosNode::browse_cb(
...
+        case AVAHI_BROWSER_NEW: 
+            LOG(DEBUG) << "Discovered service: " << name << " of type " << type << " in domain " << domain;
             // Found a new service, resolve it
             if (!(avahi_service_resolver_new(
                 node->getClient(), interface, protocol, name, type, domain,
                 AVAHI_PROTO_UNSPEC, (AvahiLookupFlags)0, resolve_cb, node)))
-            {
-                LOG(ERROR) << "Failed to resolve service '" << name << "': " << avahi_strerror(avahi_client_errno(node->getClient()));
-            }
+            LOG(ERROR) << "Failed to resolve service '" << name << "': " << avahi_strerror(avahi_client_errno(node->getClient()));
```
abrown-RV: put back the { } I don't like bare ifs.

And the cases above and below have spaces after, may as well clean up while you're in here.

sdks/cpp/common/src/Logger.cpp
```diff
@@ -33,44 +33,221 @@
...
+        << " " << rec[LogHelper::File]
+        << ":" << rec[LogHelper::Line]
+        <<"]  " << rec[expr::message];
+}
+
+// Helper filter for severity level
+bool catena_severity_filter(boost::log::attribute_value_set const& attrs) {
+    static trivial::severity_level filter_level = trivial::trace;
+    static std::string current_config = "";
+    if (current_config.compare(config::log_level) != 0) {
```
abrown-RV: apologies I wasn't clear, when I said cache, I should have said pre-compute. You've allowed here for dynamic log_level changes throughout the runtime of the app, but this isn't really needed or expected in a logging framework. Its fine for the level to get parsed into the boost::trivial enum in `init` and then just have the very simple filter based on that enum.

sdks/cpp/common/include/Config.h
```diff
@@ -133,6 +153,22 @@ inline bool mutual_authc = false;
...
+
+inline double log_size = 0;
+
+inline double log_max_size = 0;
+
+inline int log_count = 0;
+
+inline bool log_final_rotation = false;
+
+inline bool log_append = false;
```
abrown-RV: append should be true by default, this is more normal behaviour across frameworks. But still very nice to be able to turn off.

sdks/cpp/common/src/Logger.cpp
```diff
@@ -33,44 +33,221 @@
...
+        boost::shared_ptr<core> core = core::get();
+        add_common_attributes();  // Timestamps
+        if (config::log_file) {
+            // Create file sink and set parameters
+            instance().file_sink_ = boost::make_shared<file_sink_t>(
+                keywords::file_name = activeName,
+                keywords::target_file_name = targetName,
+                keywords::rotation_size = config::log_size * MB,
+                keywords::enable_final_rotation = config::log_final_rotation,
+                keywords::open_mode = std::ios_base::out | (config::log_append? std::ios_base::app : std::ios_base::trunc)
```
abrown-RV, line 182: append doesn't seem to work, in my testing it always seems to rotate at startup in multi file mode. It does work in single file mode however. Don't forget to unit test for this!

sdks/cpp/common/src/Config.cpp
```diff
@@ -64,30 +81,71 @@ std::pair<bool, int> config::initConfigVariables(int argc, char* argv[], std::st
...
+            }
+            else if (!vars[LOG_MAX_SIZE_KEY].defaulted()) {
```
abrown-RV, line 146: should have protected for negative int values. I think when I gave it log_count <=0 it treated it as log_count = 1? At least a warning of like count < 1, treating as 1, same goes for the other int configs

abrown-RV (changes requested):

KeonF-ross:
bugbot run

KeonF-ross:
bugbot run

KeonF-ross:
bugbot run

sdks/cpp/common/include/Config.h
```diff
@@ -49,40 +49,60 @@ namespace config{
...
+const double LOG_SIZE_DEFAULT = 10.0;
+const int LOG_COUNT_DEFAULT = 5;
+const double LOG_MAX_SIZE_DEFAULT = 50.0;
+const bool LOG_FINAL_ROTATION_DEFAULT = false;
+const bool LOG_APPEND_DEFAULT = false;
+#ifdef NDEBUG
+const std::string LOG_LEVEL_DEFAULT = "debug";
+#else
+const std::string LOG_LEVEL_DEFAULT = "trace";
+#endif
```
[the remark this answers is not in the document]
KeonF-ross: I agree with Cursor here, which I why I initially put it to info instead of debug, do you want me to switch it back or keep it set to debug?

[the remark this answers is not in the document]
abrown-RV: Agreed it should be info

KeonF-ross:
bugbot run

KeonF-ross:
bugbot run

sdks/cpp/common/src/Config.cpp
```diff
@@ -64,30 +88,88 @@ std::pair<bool, int> config::initConfigVariables(int argc, char* argv[], std::st
...
+                std::cout << "WARNING: log_count is set below the minimum of 1. Will be treated as log_count=1." << std::endl;
+                config::log_count = 1;
+            }
+        }
+        if (vars.count(LOG_LEVEL_KEY)){ 
+            config::log_level = vars[LOG_LEVEL_KEY].as<std::string>();
+            std::transform(config::log_level.begin(), config::log_level.end(), config::log_level.begin(), toupper);
+            if (!log_levels.contains(config::log_level.c_str())) {
+                std::cout << "WARNING: log_level {" << config::log_level << "} is invalid. ";
+                #ifdef NDEBUG
```
abrown-RV: just use the `const std::string LOG_LEVEL_DEFAULT`.

sdks/cpp/common/src/Config.cpp
```diff
@@ -64,30 +88,88 @@ std::pair<bool, int> config::initConfigVariables(int argc, char* argv[], std::st
...
+        if (vars.count(LOG_COUNT_KEY)) {
+            config::log_count = vars[LOG_COUNT_KEY].as<int>();
+            if (config::log_count < 1) {
+                std::cout << "WARNING: log_count is set below the minimum of 1. Will be treated as log_count=1." << std::endl;
+                config::log_count = 1;
+            }
+        }
+        if (vars.count(LOG_LEVEL_KEY)){ 
+            config::log_level = vars[LOG_LEVEL_KEY].as<std::string>();
+            std::transform(config::log_level.begin(), config::log_level.end(), config::log_level.begin(), toupper);
```
abrown-RV: can we use `boost::log::trivial::from_string` ?

abrown-RV:
bugbot run

abrown-RV (approved):

PR #1302, opened 2026-04-22
[Go] Include protos

abrown-RV:
## 📖 Description

We've learned that the go-way of doing generated files is to put them right in the repository. So this moves the protos to the path `sdks/go/pkg/protos` and deals with the fallout of this move.

---

## ✅ Definition of Done (DoD)
- [x] Move protos
- [x] Still works!
- [x] Update coverage to exclude protos

---

## 🛠️ Implementation Notes
Main thing is changing the output dir in the makefile to `sdks/go/pkg/protos`
Also had to put the import `github.com/rossvideo/sdks/go/pkg/protos` in a bunch of places instead of the `github.com/rossvideo/build/go`
Removed references to `build/go` from the go.mod.

---

## 🔗 Related Issues / Links

---
## 📝 Additional Context

---

## 🚧 Risks / Open Questions

I had to change the postCreateCommand.sh to remove the reference to `build/go` so you will need to recreate your dev container.
I believe we actually no longer require `go work` at all, so I'm inclined to remove it, but what it does do is let you run go cli commands from the root dir, instead of needing to cd into `sdks/go` so there is some convenience there. But also we don't need to run many go cmds directly, we have launch targets and scripts to help with those.... thoughts?

---
## ✨ Updates
- 2026-04-27 Better run the unit tests for coverage with ignoring the protos. Updated PR workflow to run coverage properly. Made issue to revamp the workflow: https://github.com/rossvideo/Catena/issues/1303


.vscode/launch.json
```diff
@@ -7,6 +7,7 @@
         "request": "launch",
         "mode": "debug",
         "program": "${workspaceFolder}/sdks/go/examples/${input:pickGoService}",
+        "preLaunchTask": "Generate Protobufs",
```
ctwarog-ross, line 10: I like this, good addition

ctwarog-ross (changes requested):
Image
Github UT action is catching the protos in coverage

nelgd (approved):

ctwarog-ross (approved):

Issue #1305, opened 2026-04-27
Constraint creation helpers


ctwarog-ross:
## 📖 Description
Create a constraint.go with methods to create every type of constraint from the protos.

---

## 🎯 Goal / Outcome
Methods for every constraint to create a new one given correct arguments

```NewConstraintInt32Choice(choices{{1,"one"},{2,"two"}})```
should return a protos.Int32ChoiceConstraint with the given choices

---

## ✅ Definition of Done (DoD)
- [ ] Feature is implemented and works as expected
- [ ] Edge cases are handled
- [ ] Tests are added or updated (if applicable)
- [ ] Documentation is updated (if needed)
- [ ] Code is reviewed and approved

---

## 🛠️ Implementation Notes (Optional)

---

## 🔗 Related Issues / Links

---
## 🚧 Risks / Open Questions (Optional)

---

abrown-RV:
when doing this make sure to update the Param.WithConstraint to use whatever this effort comes up with

PR #1309, opened 2026-04-29
[GO] Param Creation Helpers

### LINKED ISSUE #1306: Param creation helpers
## 📖 Description
Create a param.go with methods to create every type of param from the protos given a constraint.

---

## 🎯 Goal / Outcome
Methods for every constraint to create a new one given correct arguments

```NewParamInt32(fqoid, NewConstraint(choices{{1,"one"},{2,"two"}}))```
should return a protos.Int32Param with the given choices and fqoid

---

## ✅ Definition of Done (DoD)
- [ ] Edge cases are handled

---

If a constraint is invalid ex a string constraint on an int, log the error and don't apply the constraint. We don't want to return errors with these to interfere with chaining.
---

## 🔗 Related Issues / Links

---

---

### PR DESCRIPTION
**Pull request by nelgd**
## 📖 Description
Adds the `Param` builder for the Go Catena SDK for constructing `protos.Param` instances. Provides factory functions for every `ParamType`, chainable `With*` methods for metadata, and runtime `SetValue`/`GetValue` accessors with type validation.

---

## 🎯 Goal / Outcome
Build params using a builder pattern instead of manually constructing proto structs. Type mismatches, invalid constraints, and incompatible values are caught and logged at build time rather than silently producing broken protos.

---

## ✅ Definition of Done (DoD)
- [x] Factory function for every `ParamType` (INT32, FLOAT32, STRING, STRUCT, STRUCT_VARIANT, arrays, BINARY, DATA, EMPTY)
- [x] `NewParamFromValue` for dynamic/type-inferred param creation
- [x] Chainable `With*` methods: Name, Value, Constraint, ReadOnly, Widget, Precision, MaxLength, AccessScope, ClientHint, Param (sub-params), Response, Help, OidAliases, MinimalSet, Stateless, TemplateOid
- [x] `SetValue` / `GetValue` runtime accessors with type validation
- [x] Validation helpers: `isValueValidForParamType`, `isConstraintValidForParam`, `paramTypeFromValueKind`
- [x] Comprehensive test coverage

---

- Struct-type factories (`NewParamStruct`, `NewParamStructVariant`, `NewParamStructArray`, `NewParamStructVariantArray`) accept nil-able types rather than variadic args. This avoids the misleading `...` syntax that silently ignores extra arguments.
- `NewParamStructVariant` takes `*StructVariantValue` (pointer) since `StructVariantValue` is a value type — a pointer makes nil-ability idiomatic in Go.
- `WithParam` deep-clones sub-params via `proto.Clone` to prevent mutation of the original after attachment.
- `WithResponse` is noted with a TODO to migrate to a future `CatenaCommand` helper, since it only applies to command params.

---

## 🔗 Related Issues / Links

Closes issue #1306

### DIFF
--- a/sdks/go/pkg/catena/param.go
+++ b/sdks/go/pkg/catena/param.go
@@ -79,14 +79,14 @@ func NewParamString(value string) *Param {
 	}
 }
 
-func NewParamStruct(value ...map[string]any) *Param {
+func NewParamStruct(value map[string]any) *Param {
 	cp := &Param{
 		Proto: &protos.Param{
 			Type: protos.ParamType_STRUCT,
 		},
 	}
-	if len(value) > 0 {
-		pv, res := ToProto(value[0])
+	if value != nil {
+		pv, res := ToProto(value)
 		if res.Code != OK {
 			logger.Warning("NewParamStruct: failed to convert value; value left nil",
 				"error", res.Error)
@@ -97,14 +97,14 @@ func NewParamStruct(value ...map[string]any) *Param {
 	return cp
 }
 
-func NewParamStructVariant(value ...StructVariantValue) *Param {
+func NewParamStructVariant(value *StructVariantValue) *Param {
 	cp := &Param{
 		Proto: &protos.Param{
 			Type: protos.ParamType_STRUCT_VARIANT,
 		},
 	}
-	if len(value) > 0 {
-		pv, res := ToProto(value[0])
+	if value != nil {
+		pv, res := ToProto(*value)
 		if res.Code != OK {
 			logger.Warning("NewParamStructVariant: failed to convert value; value left nil",
 				"error", res.Error)
@@ -151,14 +151,14 @@ func NewParamBinary(value []byte) *Param {
 	}
 }
 
-func NewParamStructArray(value ...[]map[string]any) *Param {
+func NewParamStructArray(value []map[string]any) *Param {
 	cp := &Param{
 		Proto: &protos.Param{
 			Type: protos.ParamType_STRUCT_ARRAY,
 		},
 	}
-	if len(value) > 0 {
-		pv, res := ToProto(value[0])
+	if value != nil {
+		pv, res := ToProto(value)
 		if res.Code != OK {
 			logger.Warning("NewParamStructArray: failed to convert value; value left nil",
 				"error", res.Error)
@@ -169,14 +169,14 @@ func NewParamStructArray(value ...[]map[string]any) *Param {
 	return cp
 }
 
-func NewParamStructVariantArray(value ...[]StructVariantValue) *Param {
+func NewParamStructVariantArray(value []StructVariantValue) *Param {
 	cp := &Param{
 		Proto: &protos.Param{
 			Type: protos.ParamType_STRUCT_VARIANT_ARRAY,
 		},
 	}
-	if len(value) > 0 {
-		pv, res := ToProto(value[0])
+	if value != nil {
+		pv, res := ToProto(value)
 		if res.Code != OK {
 			logger.Warning("NewParamStructVariantArray: failed to convert value; value left nil",
 				"error", res.Error)
@@ -214,7 +214,7 @@ func NewParamData(payload DataPayload) *Param {
 
 // NewParamFromValue creates a Param by inferring the ParamType from the
 // Value's proto kind. This allows dynamic param creation regardless of type.
-// DataPayload values are mapped to ParamType_DATA (not BINARY) since the two
+// DataPayload values are mapped to ParamType_DATA since the two
 // cannot be distinguished from the value alone.
 func NewParamFromValue(v Value) *Param {
 	pt := paramTypeFromValueKind(v.Value)
--- a/sdks/go/pkg/catena/param.go
+++ b/sdks/go/pkg/catena/param.go
@@ -244,6 +244,12 @@ var paramTypesWithSubParams = map[protos.ParamType]bool{
 }
 
 func (cp *CatenaParam) WithParam(oid string, param *CatenaParam) *CatenaParam {
+	if param == nil {
+		logger.Warning("WithParam called with nil sub-param; ignoring",
+			"oid", oid,
+			"param_type", cp.param.Type.String())
+		return cp
+	}
 	if !paramTypesWithSubParams[cp.param.Type] {
 		logger.Warning("WithParam called on param type that does not support sub-params; ignoring",
 			"param_type", cp.param.Type.String())
--- a/sdks/go/pkg/catena/param.go
+++ b/sdks/go/pkg/catena/param.go
@@ -39,8 +39,6 @@
 package catena
 
 import (
-	"sync"
-
 	"google.golang.org/protobuf/proto"
 
 	"github.com/rossvideo/catena/sdks/go/pkg/logger"
@@ -48,9 +46,7 @@ import (
 )
 
 // CatenaParam wraps a protos.Param and exposes a fluent builder API.
-// All methods are safe for concurrent use.
 type CatenaParam struct {
-	mu    sync.RWMutex
 	proto *protos.Param
 }
 
@@ -236,15 +232,11 @@ func NewParamFromValue(v Value) *CatenaParam {
 // --- Chainable With* methods ---
 
 func (cp *CatenaParam) WithName(name PolyglotText) *CatenaParam {
-	cp.mu.Lock()
-	defer cp.mu.Unlock()
 	cp.proto.Name = &protos.PolyglotText{DisplayStrings: name}
 	return cp
 }
 
 func (cp *CatenaParam) WithValue(v Value) *CatenaParam {
-	cp.mu.Lock()
-	defer cp.mu.Unlock()
 	if !isValueValidForParamType(v.Value, cp.proto.Type) {
 		logger.Warning("WithValue: value kind incompatible with param type; ignoring",
 			"param_type", cp.proto.Type.String())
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: sdks/go/pkg/catena/param.go [hunk 1]
```diff
@@ -0,0 +1,317 @@
...
+
+func (cp *CatenaParam) WithClientHint(key, value string) *CatenaParam {
+	if cp.param.ClientHints == nil {
+		cp.param.ClientHints = map[string]string{}
+	}
+	cp.param.ClientHints[key] = value
+	return cp
+}
+
+var paramTypesWithSubParams = map[protos.ParamType]bool{
```
abrown-RV: more go way to do a set with a map is to do `map[protos,paramType]struct{}`

sdks/go/pkg/catena/param.go
```diff
@@ -0,0 +1,317 @@
...
+		return cp
+	}
+	if cp.param.Params == nil {
+		cp.param.Params = map[string]*protos.Param{}
+	}
+	cp.param.Params[oid] = proto.Clone(param.param).(*protos.Param)
+	return cp
+}
+
+func (cp *CatenaParam) WithResponse(response bool) *CatenaParam {
```
abrown-RV: response only matters for commands, we can probably do a CatenaCommand helper that composes from CatanaParam and the WithRepsonse goes there.... might be a bit more to it since there's probably other fields that have no meaning for command, so maybe there's an unexported `baseParam` that `Param` and `Command` can use. Save all that for when we do the Command helper.

sdks/go/pkg/catena/param.go
```diff
@@ -0,0 +1,317 @@
...
+	cp.param.MinimalSet = minimalSet
+	return cp
+}
+
+func (cp *CatenaParam) WithStateless(stateless bool) *CatenaParam {
+	cp.param.Stateless = stateless
+	return cp
+}
+
+func (cp *CatenaParam) WithTemplateOid(oid string) *CatenaParam {
```
abrown-RV: this should link to another param.... idk if there's a cool way to link automatically for people. Probably not, but give it a think.

sdks/go/pkg/catena/param.go
```diff
@@ -0,0 +1,317 @@
...
+	cp.param.TemplateOid = oid
+	return cp
+}
+
+// Build returns the underlying protos.Param.
+func (cp *CatenaParam) Build() *protos.Param {
```
abrown-RV: related: https://github.com/rossvideo/Catena/issues/1323

sdks/go/pkg/catena/param.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+// --- Chainable With* methods ---
+
+func (cp *CatenaParam) WithName(name PolyglotText) *CatenaParam {
+	cp.param.Name = &protos.PolyglotText{DisplayStrings: name}
+	return cp
+}
+
+func (cp *CatenaParam) WithConstraint(c *protos.Constraint) *CatenaParam {
```
abrown-RV: this should have the same `Constraint` helper treatment... new issue I guess

abrown-RV (changes requested):

sdks/go/pkg/catena/param.go
nelgd: I think we have an issue for it #1305

sdks/go/pkg/catena/param.go
abrown-RV: I would like `cp.param` to just be unexported `.Proto`. Starting the pattern now, so we don't have to refactor later.

In that case we can just remove .Build() which is a bit of a misnomer anyways, it doesn't build anything, and multiple calls return the same reference, which is not what is implied by "Build" at least to me.

sdks/go/pkg/catena/param.go
```diff
@@ -0,0 +1,325 @@
...
+		logger.Warning("NewParamData: failed to convert DataPayload; value left nil",
+			"error", res.Error)
+		return cp
+	}
+	cp.param.Value = &protos.Value{Kind: &protos.Value_DataPayload{DataPayload: pdp}}
+	return cp
+}
+
+// --- Chainable With* methods ---
+
```
abrown-RV, line 227: We need a way for the more complex types to get values struct, struct array, variant, variant array. Ideas:
- add a WithValue(value CatenaValue) ok, but would need to warn if the CatenaValue is a different type.
- could look at CatenaValue and match the type in the big switch. make for example, NewParamStruct(value map[string]any), etc. downside here is that it would need to get validated, and it might return an error... nothing stopping it from returning an err, its just a little different from the other NewFuncs.

- maybe a NewParamFromValue(value CatenaValue) this lets them convert to a CatenaValue earlier and deal with the error there. Regardless of other option, I kind of see this one as a real convienience cause it will allow for dynmaic param creation regardless of type.

abrown-RV (dismissed):

sdks/go/pkg/catena/param.go
ctwarog-ross: what about... all 3? Give the business logic the option to create however they choose.  My main concern is with thread safety. Manually making the devicerequest return map has allowed us to plug in getter methods with their own mutex handling to the "value" field of the param definitions. In this case where we're wrapping raw data then presumably unmarshalling it to create the devicerequest return, we could have a giant mutex lock before the request starts building but what if there are several mutexes at play?

nelgd: Removed the per param mutexs and added in Andrews suggestions for struct and variant stuff

sdks/go/pkg/catena/param.go
```diff
@@ -0,0 +1,499 @@
...
+import (
+	"google.golang.org/protobuf/proto"
+
+	"github.com/rossvideo/catena/sdks/go/pkg/logger"
+	"github.com/rossvideo/catena/sdks/go/pkg/protos"
+)
+
+// CatenaParam wraps a protos.Param and exposes a fluent builder API.
+type CatenaParam struct {
+	proto *protos.Param
```
abrown-RV:
```suggestion
	Proto *protos.Param
```
So a sophisticated BL can directly manipulate it.

sdks/go/pkg/catena/param.go
```diff
@@ -0,0 +1,499 @@
...
+
+import (
+	"google.golang.org/protobuf/proto"
+
+	"github.com/rossvideo/catena/sdks/go/pkg/logger"
+	"github.com/rossvideo/catena/sdks/go/pkg/protos"
+)
+
+// CatenaParam wraps a protos.Param and exposes a fluent builder API.
+type CatenaParam struct {
```
abrown-RV: don't forget its just `Param` now.

sdks/go/pkg/catena/param.go
```diff
@@ -0,0 +1,499 @@
...
+func NewParamStructVariant(value ...StructVariantValue) *CatenaParam {
+	cp := &CatenaParam{
+		proto: &protos.Param{
+			Type: protos.ParamType_STRUCT_VARIANT,
+		},
+	}
+	if len(value) > 0 {
+		pv, res := ToProto(value[0])
+		if res.Code != OK {
+			logger.Warning("NewParamStructVariant: failed to convert value; value left nil",
```
abrown-RV, line 107: hit coverage for all these err paths

nelgd:
All the err cases are hit with coverage now

sdks/go/pkg/catena/param.go
```diff
@@ -0,0 +1,493 @@
...
+func NewParamBinary(value []byte) *Param {
+	return &Param{
+		Proto: &protos.Param{
+			Type:  protos.ParamType_BINARY,
+			Value: &protos.Value{Kind: &protos.Value_DataPayload{DataPayload: &protos.DataPayload{Kind: &protos.DataPayload_Payload{Payload: value}}}},
+		},
+	}
+}
+
+func NewParamStructArray(value ...[]map[string]any) *Param {
```
abrown-RV: trying to decide how I feel about the ... which lets you basically have either a value or not (technically you can give it more than 1 value, but it would ignore the rest of them XD) Is there a reason for the variadic approach?

ctwarog-ross: I don't think we should have variadic args in these functions. There isn't really any purpose to being able to create instances without any args, and it ignores anything after the first arg anyway

nelgd: Makes sense, changed it to having required agreements

ctwarog-ross:
Remove the nil checks from things such as struct array. Change ToProto to correctly handle nil values. Add a check at the top of ToProto that returns an err if v is nil

ctwarog-ross:
We discovered more things.
Maps can be nil OR empty, and needs to be accounted for in the code as well as tested.
arrays can be empty but NOT nil.

What needs to be added:
change nil check in toProto in struct and struct variant to If len(val) is 0
Unit test exists for map[string] value nil, add one for map with no values (data := map[string]any{}, toProto(data))
Same thing add another unit test for empty struct_variant

ctwarog-ross (dismissed):

ctwarog-ross (dismissed):

sdks/go/pkg/catena/param.go
```diff
@@ -0,0 +1,487 @@
...
+// string, but Param doesn't track its own OID path (the OID is only
+// assigned externally via WithParam). Auto-linking would require params to
+// know their full mounted path, which adds significant complexity. Revisit if
+// we ever add path tracking to the param tree.
+func (cp *Param) WithTemplateOid(oid string) *Param {
+	cp.Proto.TemplateOid = oid
+	return cp
+}
+
+// --- Thread-safe runtime accessors ---
```
abrown-RV: not thread safe anymore

nelgd: forgot to remove that oops

abrown-RV (approved):
Thanks for your patience Nelson, sorry for the runaround on this one.

nelgd:
> Thanks for your patience Nelson, sorry for the runaround on this one.

no worries

PR #1312, opened 2026-05-01
[C++] Duplicate gRPC static into REST examples

abrown-RV:
A bunch of the REST examples didn't have their grid files copied over since dashboard doesn't support REST. But I wanted to get consistency for the testing I'm doing. Also Christian is working on the DB support, so this will be nice to have.

cue suggestion to unify our cpp examples like we have with go.


sdks/cpp/connections/REST/examples/one_of_everything_REST/device.one_of_everything.yaml
```diff
@@ -27,91 +26,122 @@ language_packs:
...
+    read_only: true
+    value: { string_value: "eo://catena_logo_16x16.png" }
 
   number_example: 
     type: INT32
     value: { int32_value: 0 }
-    name: { display_strings: { en: "Int Example" } }
+    name: { display_strings: { 
+      en: "Int Example", 
+      fr: "Exemple Entier",
```
NR-RV: `Exemple d'entier`

PR #1313, opened 2026-05-01
[Go] [Env] Fix creating build/go in the wrong place

abrown-RV:
`postCreateCommand.sh` messed up a little in #1302 . Fix creating the dir in the wrong spot.


ctwarog-ross (approved):
Confirmed no longer creates the dir on local build

Issue #1318, opened 2026-05-05
Heartbeat


ctwarog-ross:
## 📖 Description
Go catena services intermittently disconnect/reconnect in Dashboard/RPM. This is due to a timeout where if the client doesn't receive any messages for a certain period of time, it disconnects and reinitializes the connection.

---

## 🎯 Goal / Outcome
Implement a heartbeat like in cpp that sends a given param value update every specified interval, so that DB/RPM connections do not timeout
---

## ✅ Definition of Done (DoD)
- [ ] Feature is implemented and works as expected
- [ ] Edge cases are handled
- [ ] Tests are added or updated (if applicable)
- [ ] Documentation is updated (if needed)
- [ ] Code is reviewed and approved

---

## 🛠️ Implementation Notes (Optional)

---

## 🔗 Related Issues / Links

---
## 🚧 Risks / Open Questions (Optional)

---

PR #1319, opened 2026-05-05
[Go] Heartbeat introduced

ctwarog-ross:
## 📖 Description

Closes Issue #1318

Heartbeat pattern matches cpp
Can be interacted with through server api with Start/StopHeartbeat Can also be interacted with manually as detailed in Heartbeat.go comments

Every specified interval, a given param update will be sent to the client

---

## ✅ Definition of Done (DoD)
- [x] Edge cases are handled

---

## 🛠️ Implementation Notes

---

## 🔗 Related Issues / Links

---
## 📝 Additional Context

---

## 🚧 Risks / Open Questions

---
## ✨ Updates


NR-RV (dismissed):

ctwarog-ross:
bugbot run

ctwarog-ross:
bugbot run

sdks/go/pkg/catena/heartbeat.go
```diff
@@ -0,0 +1,165 @@
...
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.handlers = append(h.handlers, handler)
+	return h
+}
+
+// Start begins emitting tick events at the specified interval.
+// Returns true if the heartbeat was started, false if it was already running
+// or the interval is invalid (zero or negative).
+func (h *Heartbeat) Start(interval time.Duration) bool {
```
abrown-RV: I think more go idomatic is to return err if there's an error. and the do nothing if already running, does not need a special return? So that can just return nil, as can the final return and the the if bad internal can do the fmt.Errorf() or w/e that's called

sdks/go/pkg/catena/heartbeat.go
```diff
@@ -0,0 +1,165 @@
...
+	h.mu.Lock()
+	handlers := make([]HeartbeatHandler, len(h.handlers))
+	copy(handlers, h.handlers)
+	h.mu.Unlock()
+
+	for _, handler := range handlers {
+		func() {
+			defer func() {
+				if r := recover(); r != nil {
+					slog.Error("panic in heartbeat handler", "error", r)
```
abrown-RV: Should use our logger?

sdks/go/pkg/internal/base_server.go
```diff
@@ -209,3 +211,73 @@ func (bs *BaseServer) BroadcastUpdate(slot uint16, oid string, value any) {
 	}
 	bs.NotifyUpdate(update)
 }
+
```
abrown-RV, line 257: coverage on the new code please

sdks/go/pkg/catena/heartbeat.go
```diff
@@ -0,0 +1,165 @@
...
+	return &Heartbeat{
+		handlers: make([]HeartbeatHandler, 0),
+	}
+}
+
+// OnTick registers a handler to be called on each heartbeat tick.
+// Multiple handlers can be registered and will be called in order.
+// Handlers should not block for long periods.
+// Returns the Heartbeat for method chaining.
+func (h *Heartbeat) OnTick(handler HeartbeatHandler) *Heartbeat {
```
abrown-RV: should there be a way to remove handlers? I guess you just remake the heartbeat without the handler you don't want anymore....

sdks/go/examples/getDevice/getdevice.go
```diff
@@ -54,6 +56,19 @@ var Devices = map[uint16]map[string]any{
 		"access_scopes":     []string{"st2138:mon", "st2138:op", "st2138:cfg", "st2138:adm"},
 		"default_scope":     "st2138:op",
 		"params": map[string]any{
+			"version": map[string]any{
```
abrown-RV: missing the nested `product` struct. Dunno how much we care... the heartbeat doesn't care... but maybe attacted clients will? Unsure here... This is the joy of the detached nature of the handlers architecture...

sdks/go/examples/getDevice/getdevice.go
```diff
@@ -436,15 +452,21 @@ func RegisterHandlers(srv catena.CatenaServer) {
...
+				return catena.ReplyError[catena.CatenaDevice](catena.INTERNAL, "failed to convert device")
 			}
 
 			return catena.Reply(device)
 		})
 	}
+
+	// Start heartbeat — broadcasts the version param on slot 0 every 5 seconds.
+	// This mirrors the C++ SDK pattern: dm.setHeartbeatParam("/product/version"); dm.startHeartbeat();
+	srv.StartHeartbeat(0, "/product/version", func() any {
```
abrown-RV: Should tick for each slot. Heartbeat is not just a keeping the connection alive concern. It is also reassuring attached clients that each slot is still ok.

abrown-RV (changes requested):

sdks/go/examples/getDevice/getdevice.go
ctwarog-ross: Ill make the product struct, to show intended convention

ctwarog-ross:
bugbot run

sdks/go/examples/getDevice/getdevice.go
```diff
@@ -54,6 +56,29 @@ var Devices = map[uint16]map[string]any{
...
+					"version": map[string]any{
+						"name": map[string]any{
+							"display_strings": map[string]string{
+								"en": "Version",
+							},
+						},
+						"type":      catena.ParamTypeString,
+						"read_only": true,
+						"widget":    "TEXT",
+						"value": map[string]any{
```
abrown-RV: values in sub params of structs belong at the top level, so the `product` level should have:

```go
"value": map[string]any {
    "struct_value": map[string]any {
        "fields": map[string]any {
            "version": map[string]any {
                "string_value": "1.0.0"
            }
        }
    }
}
```
verbose, I know XD

ctwarog-ross:
bugbot run

abrown-RV (dismissed):

abrown-RV (approved):

PR #1331, opened 2026-05-08
GetPopulatedSlots

ctwarog-ross:
## 📖 Description
GetPopulatedSlots endpoint implemented for REST
default log level set to INFO

---

## ✅ Definition of Done (DoD)

---

## 🛠️ Implementation Notes

---

## 🔗 Related Issues / Links

---
## 📝 Additional Context

---

## 🚧 Risks / Open Questions

---
## ✨ Updates

### DIFF
--- a/sdks/go/pkg/rest/server.go
+++ b/sdks/go/pkg/rest/server.go
@@ -353,6 +353,9 @@ func (s *Server) RegisterRoutes() {
 	// Connect endpoint: GET /st2138-api/v1/connect (SSE streaming)
 	s.mux.HandleFunc("/st2138-api/v1/connect", s.handleConnect)
 
+	// Devices endpoint: GET /st2138-api/v1/devices (returns populated slots)
+	s.mux.HandleFunc("/st2138-api/v1/devices", s.handleGetPopulatedSlots)
+
 	// Catch-all for 404 - must be registered last
 	s.mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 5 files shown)


NR-RV (dismissed):

sdks/go/pkg/rest/server.go
```diff
@@ -370,6 +373,29 @@ func (s *Server) RegisterRoutes() {
...
+	slots := s.baseServer.Slots
+
+	response := map[string][]uint32{
+		"slots": slots,
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(response); err != nil {
+		logger.Error("failed to write slots response", "error", err)
```
abrown-RV, line 395: this should probably return an error if the body fails to encode, but can pass that off to someone else in a ticket for faster turnaround

ctwarog-ross: I don't think that's a use case that will ever be hit, at least from marshaling, since it will only ever be populated by the exact same members in the same structure

NR-RV (approved):

sdks/go/pkg/rest/server.go
abrown-RV: I agree, you're probably right, but those sound like famous last words XD

We already have json marshalling helpers, why not use those?

ctwarog-ross: I can implement a marshal checker manually, the other ones use catena values

PR #1333, opened 2026-05-08
[GO] Fixed Race Conditions

### LINKED ISSUE #1320: Race condition exposed by testing
## 📖 Description
pkg/rest | FAIL | Pre-existing: test reads httptest.ResponseRecorder buffer concurrently with SSE writer goroutine (server_test.go:962). Not heartbeat-related — it's a test harness synchronization bug.
-- | -- | --
pkg/grpc | Intermittent FAIL | Flaky — passed on re-run. When it fails, it's the same pattern (test assertion vs async server activity).

---

## 🎯 Goal / Outcome
Run tests to check for race condition using
```
go test -race -count=10 ./pkg/catena/... ./pkg/internal/...
```
Identify and fix the condition
---

## ✅ Definition of Done (DoD)
- [ ] Edge cases are handled

---

---

## 🔗 Related Issues / Links

---

---

## 📝 Additional Context (Optional)

[pkg/rest](vscode-file://vscode-app/c:/Users/christian.twarog/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html)	FAIL	Pre-existing: test reads httptest.ResponseRecorder buffer concurrently with SSE writer goroutine (server_test.go:962). Not heartbeat-related — it's a test harness synchronization bug.
[pkg/grpc](vscode-file://vscode-app/c:/Users/christian.twarog/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html)	Intermittent FAIL	Flaky — passed on re-run. When it fails, it's the same pattern (test assertion vs async server activity).

nelgd:
## 📖 Description
Fixes multiple data races in the Go SDK uncovered by `go test -race`, and adds a Makefile target to run the race detector as part of the development workflow.

---

## 🎯 Goal / Outcome
All Go SDK tests pass cleanly under the race detector (`go test -race`) with no reported data races, and the team has a convenient `make race` target to catch future regressions.

---

## ✅ Definition of Done (DoD)
- [x] `currentEnv` global in `config.go` is converted from a plain variable to `atomic.Pointer[Environment]` so concurrent readers and writers no longer race
- [x] `time.Sleep` startup waits removed from gRPC integration tests; replaced with synchronous `Start` + `grpc.WithBlock()` dial to guarantee the server is ready
- [x] `stubServerRuntime` test double protected with a `sync.Mutex` around `RegisterTransportConnection`, `DeregisterConnection`, and `ShutdownTransportConnections`
- [x] `setupSSEConnection` rewritten to return a cleanup function that cancels the context **and** waits for the SSE handler goroutine to exit, preventing goroutine leaks and races on shared counters
- [x] REST transport tests updated to use the new cleanup function instead of raw `context.CancelFunc`
- [x] `gRPC Start()` now skips listener creation when one is already set (supports test injection)
- [x] `make race` target added with configurable `RACE_PACKAGES` and `RACE_COUNT` variables

---

- **`atomic.Pointer[Environment]`** was chosen over a `sync.RWMutex` for `currentEnv` because the access pattern is overwhelmingly reads (gRPC interceptors on every RPC) with rare writes (test setup / `InitOptions`). An atomic pointer is lock-free on the read path and has zero contention.
- **Listener injection in `GrpcTransport.Start`**: the `if t.listener == nil` guard allows tests to pre-set a listener (e.g. `bufconn`) without `Start` clobbering it, while production callers that leave it `nil` still get the normal `net.Listen` path.
- **`setupSSEConnection` cleanup function**: the old pattern returned a bare `context.CancelFunc`, which cancelled the request but never waited for the handler goroutine to finish. This caused races on the shared `stubServerRuntime` counters (`registerCalls`, `deregisterCalls`). The new cleanup function blocks until the handler exits (with a 2-second safety timeout).
- The `make race` target defaults to `RACE_COUNT=10` iterations to surface intermittent races.

---

## 🔗 Related Issues / Links

Closes issue #1320

### DIFF
--- a/sdks/go/pkg/catena/server.go
+++ b/sdks/go/pkg/catena/server.go
@@ -204,6 +204,10 @@ type ServerRuntime interface {
 	DeregisterConnection(connID int)
 }
 
+// CatenaServer is an alias for Server, following the CatenaValue/CatenaDevice
+// naming convention used by the rest of the public API.
+type CatenaServer = Server
+
 var _ Server = (*server)(nil)
 var _ ServerRuntime = (*server)(nil)
 
--- a/sdks/go/pkg/catena/server.go
+++ b/sdks/go/pkg/catena/server.go
@@ -204,10 +204,6 @@ type ServerRuntime interface {
 	DeregisterConnection(connID int)
 }
 
-// CatenaServer is an alias for Server, following the CatenaValue/CatenaDevice
-// naming convention used by the rest of the public API.
-type CatenaServer = Server
-
 var _ Server = (*server)(nil)
 var _ ServerRuntime = (*server)(nil)
 
--- a/sdks/go/pkg/transports/grpc_transport.go
+++ b/sdks/go/pkg/transports/grpc_transport.go
@@ -71,28 +71,7 @@ type GrpcTransport struct {
 var _ catena.Transport = (*GrpcTransport)(nil)
 
 func NewGrpcTransport(port uint16, reflectionEnabled bool) *GrpcTransport {
-	transport := &GrpcTransport{
-		catenaService: &catenaService{},
-		grpcServer: grpc.NewServer(
-			grpc.UnaryInterceptor(unaryInterceptor),
-			grpc.StreamInterceptor(streamInterceptor),
-		),
-		port:       port,
-		reflection: reflectionEnabled,
-	}
-	transport.catenaService.transport = transport
-
-	// Register the CatenaService with the gRPC server
-	protos.RegisterCatenaServiceServer(transport.grpcServer, transport.catenaService)
-
-	// Register reflection service on gRPC server if enabled
-	if reflectionEnabled {
-		reflection.Register(transport.grpcServer)
-		logger.Info("gRPC server created with reflection enabled")
-	} else {
-		logger.Info("gRPC server created")
-	}
-	return transport
+	return newGrpcTransport(port, nil, reflectionEnabled)
 }
 
 // NewGrpcTransportWithListener creates a GrpcTransport that will use the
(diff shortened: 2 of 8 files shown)


nelgd:
Ran `go test` on `rest` and `grpc` packages multiple times with no failures.

---

image

abrown-RV (changes requested):
can you add a makefile target, similar to `make coverage` that runs this so its easy going forward?

nelgd:
> can you add a makefile target, similar to `make coverage` that runs this so its easy going forward?

yup, was going to ask about that

abrown-RV:
Great minds.... 😄

abrown-RV:
Looks good, but I'm going to say this can wait for the Transports merge, and then @nelgd and I can go through and resolve merge issues together

nelgd:
Rebased but the race condition test isnt passing. Looking into a fix for this

sdks/go/pkg/catena/server.go
```diff
@@ -136,6 +136,10 @@ type ServerRuntime interface {
 	DeregisterConnection(connID int)
 }
 
+// CatenaServer is an alias for Server, following the CatenaValue/CatenaDevice
+// naming convention used by the rest of the public API.
+type CatenaServer = Server
```
abrown-RV: don't need this anymore, we redid all the others to drop the Catena prefix

sdks/go/pkg/transports/grpc_transport.go
```diff
@@ -69,21 +69,32 @@ type GrpcTransport struct {
 var _ catena.Transport = (*GrpcTransport)(nil)
 
 func NewGrpcTransport(port uint16, reflectionEnabled bool) *GrpcTransport {
+	return newGrpcTransport(port, nil, reflectionEnabled)
+}
+
+// NewGrpcTransportWithListener creates a GrpcTransport that will use the
+// provided listener instead of binding a new one in Start(). This avoids
+// a TOCTOU race in tests where the port could be claimed between Close()
+// and rebind.
```
abrown-RV: This is a lot of faff, see if you can just do port=0 in the unit tests. The sprintf will give you `":0"` for `net.Listen()` which will just assign you a random port, like the TIme of Check part of TOCTOU that already exists. Then after calling start you can just do transport.listener.Addr() to get the port to hook up in the unit tests.

abrown-RV (changes requested):
Just not a fan of the new GrpcTransport ctor, I think I have a better suggestion in the comments below.

abrown-RV (approved):

PR #1334, opened 2026-05-08
[GO] Value return fix

nelgd:
## 📖 Description

Refactors the `oneOfEverything` Go example so its on-the-wire model and live updates actually showcase what an adopter would build.

- Slot 0's "counter dashboard" is split from a single struct param into two distinct `int32` params (`counter` and `running`), with `running` made `read_only` and constrained by an `INT_CHOICE` so it serves as a real demo of constraints + read-only.
- `GetDevice` now returns the **current** value of every param across every slot — slot 0 reads from `CounterState`, slots 1 and 2 read from the same `slotParams` store that `SetValue` writes to.
- The bundled REST web UI is updated to consume the new wire shape and gains an `⛶ Expand` button next to the slot picker, opening the device JSON in a wider, syntax-highlighted popup that reuses the asset-modal scaffold.

---

## 🎯 Goal / Outcome

- Address inline review feedback: split the struct-shaped counter, plug live state into the device descriptor, mark `running` as `read_only`, and give it an `INT_CHOICE` constraint so the example demonstrates more of the SDK's surface.
- Keep `CounterState` and the descriptor in sync so a fresh `GetDevice` always reflects the live state, never an init-time zero.
- Make the device JSON in the web UI easy to read without growing the inline section: one click expands it into a roomy modal.

---

## ✅ Definition of Done (DoD)

- [x] `counter` and `running` modelled as two distinct `int32` params on slot 0
- [x] `running` carries `read_only: true` and an `INT_CHOICE` constraint (`0 = Not Counting`, `1 = Counting`), with display name "Counter Running Status"
- [x] `GetDevice` for slot 0 serves live values via `CounterState` (`GetValue()` / `RunningInt32()`)
- [x] `GetDevice` for slots 1 & 2 serves live values via two new helpers (`int32ParamValue`, `stringParamValue`) reading from `slotParams`
- [x] `GetValue` returns the live `CounterState` value for both `counter` and `running`
- [x] `SetValue` on `running` is explicitly rejected with a hint to use `start` / `stop`
- [x] `start` / `stop` broadcast the running flag on the new `running` param
- [x] Web UI polls `counter` and `running` separately, routes SSE updates by oid, and drives the badge + glow from the new flag
- [x] Web UI gains an `⛶ Expand` button that opens the device JSON in a wide, syntax-highlighted modal (reuses `closeAssetModal` and the asset-modal scaffold)

---

## 🛠️ Implementation Notes

- `var Devices = …` was promoted to `func BuildDevices(counter *CounterState, slotParams map[uint16]*sync.Map)`, called once per `GetDevice`. Construction is cheap and the descriptor is small, so per-request rebuild is a non-issue.
- New `CounterState.RunningInt32()` provides the on-the-wire `0` / `1` representation, used consistently by `BuildDevices`, `GetValue`, and the broadcast helper.
- New `int32ParamValue` / `stringParamValue` helpers each return a `value{}` map (`{"int32_value": …}` or `{"string_value": …}`) directly from a `*sync.Map` lookup, with a typed fallback if the key is missing or the wrong type. `sync.Map.Load` is concurrency-safe so this happily races against `SetValue` writes.
- The legacy `BuildCounterResponse(counter)` helper (which produced a `{counter, running}` struct payload) is removed; commands and `GetValue` for slot 0 now return plain `int32`, and a small `broadcastRunning()` closure handles the flag separately.
- `SetValue` on slot 0's `running` is hard-rejected with `INVALID_ARGUMENT` ("use start/stop commands to change running state") to prevent the slot-param cache from drifting away from `CounterState`.

---

## 🔗 Related Issues / Links

---

## 🚧 Risks / Open Questions

- **`BuildDevices` rebuilds the entire slot map per request.** Negligible at three slots with ~10 params; if the example grows substantially we'd want to memoise the static portions and only re-evaluate the live `value` fields.

---

## 📝 Additional Context

Files changed (4 commits, +433 / -260 vs `develop`)

### DIFF
--- a/sdks/go/examples/oneOfEverything/oneofeverything.go
+++ b/sdks/go/examples/oneOfEverything/oneofeverything.go
@@ -401,7 +401,12 @@ func InitSlotParams() map[uint16]*sync.Map {
 	return slotParams
 }
 
-func BuildCounterResponse(counter *CounterState) map[string]any {
+// BuildCounterCommandResponse builds the command-result payload that includes
+// both the current counter value and the running flag. It is used as the
+// response body for start/stop/add10/reset commands so callers can observe
+// the side effects of a command in a single round trip. The "counter" param
+// itself is a plain int32 — see the GetValue handler and the broadcasts below.
+func BuildCounterCommandResponse(counter *CounterState) map[string]any {
 	var runningVal int32 = 0
 	if counter.IsRunning() {
 		runningVal = 1
@@ -439,9 +444,8 @@ func RegisterHandlers(srv catena.CatenaServer) {
 				counter.Start()
 				logger.Info("Counter started", "value", counter.GetValue())
 			}
-			resp := BuildCounterResponse(counter)
-			srv.BroadcastUpdate(0, "counter", resp)
-			val, _ := catena.ToCatenaValue(resp)
+			srv.BroadcastUpdate(0, "counter", counter.GetValue())
+			val, _ := catena.ToCatenaValue(BuildCounterCommandResponse(counter))
 			return catena.CommandReply(val)
 		},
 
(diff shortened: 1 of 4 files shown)


sdks/go/examples/oneOfEverything/oneofeverything.go
```diff
@@ -135,6 +144,17 @@ var Devices = map[uint16]map[string]any{
...
+				"name": map[string]any{
+					"display_strings": map[string]string{
+						"en": "Running",
+					},
+				},
+				"type": catena.ParamTypeInt32,
+				"value": map[string]any{
+					"int32_value": 0,
+				},
+			},
```
ctwarog-ross, line 212:
```
"running": map[string]any{
	"name": map[string]any{
		"display_strings": map[string]string{
			"en": "Counter Running Status",
		},
	},
	"type": catena.ParamTypeInt32,
	"value": map[string]any{
		"int32_value": RunningInt32(),
	},
	"constraint": map[string]any{
		"type": "INT_CHOICE",
		"int32_choice": map[string]any{
			"choices": []map[string]any{
				{
					"value": int32(0),
					"name": map[string]any{
						"display_strings": map[string]string{
							"en": "Not Counting",
						},
					},
				},
				{
					"value": int32(1),
					"name": map[string]any{
						"display_strings": map[string]string{
							"en": "Counting",
						},
					},
				},
			},
		},
	},
},
```
Change running to an int_choice like so.
The purpose of one of everything is to show an adopter an example of everything they can do

ctwarog-ross (changes requested):
Change device request to serve up current values of params. My requested change on running oid shows how to do that by plugging in the get method RunningInt32() to the value field

sdks/go/examples/oneOfEverything/oneofeverything.go
abrown-RV: Should also be `read_only:true` if `SetValue` prevents changing

ctwarog-ross (approved):

PR #1342, opened 2026-05-13
run as root

abrown-RV:
My advice to switch the container to non-root was a bad one. Switching back to root.


ctwarog-ross (approved):
hurrah

PR #1345, opened 2026-05-19
[GO] Add catena package documentation and update status codes

### LINKED ISSUE #1322: [Go] Unify status codes
## 📖 Description

Currently the status codes have specific to HTTP and specific to GRPC. There is already some overlap. For a BL that is trying to implement handlers that work with any transport, it would be confusing trying to decide on return codes.

They should also be renamed to follow common go naming conventions:
```go
type StatusCode int
const (
  StatusCodeOk StatusCode = 0
  StatusCodeCancelled = 1
  // etc.
)
```

---

## 🎯 Goal / Outcome
We should have a clear list of status codes with clear descriptions of their use cases. Based on the allowed responses from the openapi.yaml REST side. Not sure what to do about the grpc side, maybe we have to just support all of them? But maybe we can dispense with some that have no use for us.

We also need a clear table of the mappings between all of the statuses and their real codes in the different transports.

---

## ✅ Definition of Done (DoD)
- [x] Decide on minimal list of codes we actually need
- [x] Define mappings
- [x] Implement changes

---

For status codes like `HTTP 204`, which is the proper response for SetValue for example, the Handler should just return `StatusCodeOk`. The transport should be responsible for knowing that an `OK` from a `SetValue` request is actually 204, and not a 200. So the point I'm making is the mapping can be a little more involved.

---

## 🔗 Related Issues / Links

---

---

ZacharyDuaibesRV:
Unify the transport-neutral status code model used by Catena handlers, and document it in one place, so business logic that wants to work with either REST or gRPC has a clear contract.

The handler type stays `catena.StatusCode`. Each transport (REST, gRPC) is responsible for mapping a `StatusCode` to the right wire shape, so handlers no longer have to pick HTTP numbers (e.g. 200 vs 204 for `SetValue`) or gRPC codes directly.

- New `sdks/go/pkg/catena/StatusCode.md` capturing:
  - The transport-neutral `StatusCode` model and helpers.
  - Normative references to **SMPTE PCD ST 2138-11 §6.2 Table 2** (gRPC) and **ST 2138-12 §7.3 Table 1 + §7.4–§7.11** (REST).
  - Quick rules for handler authors.
  - Cheat sheet of `StatusCode` → HTTP / gRPC wire mappings.
  - Per-code semantics (cause / when to return / what clients infer), with examples drawn from ST 2138-11.
  - Legacy / transport-shaped constants table marking which ones have no basis in ST 2138-12 and what to use instead.
  - Naming divergence vs ST 2138-11 (`CANCELED` vs `CANCELLED`, `TIMEOUT EXCEEDED` vs `DEADLINE_EXCEEDED`).
  - Callout that **HTTP 412 Precondition Failed is unrelated to `FailedPrecondition`** (different concept despite the name).
- removed `doc.go`.
- `status_code.go` constant comments updated: legacy HTTP-shaped constants get `// Deprecated:` notices with what-to-use-instead guidance.
- References [SMPTE/st2138-12#5](https://github.com/SMPTE/st2138-12/issues/5) (proposal to add HTTP 400 to §7.3 Table 1).

To Do:
- [ ] Rewrite `ToHTTPStatus` (`pkg/transports/rest_transport.go`) so default failures are the ST 2138-12 §7.3 set `{401, 403, 404, 500, 503}` and legacy HTTP-shaped constants are no longer special-cased.
- [ ] Trim `ToGRPCCode` (`pkg/transports/grpc_transport.go`) to the 0–16 cases only; drop the REST-numeric branches.
- [ ] Migrate `pkg/transports/json_utils.go` from `catena.BAD_REQUEST` → `catena.INVALID_ARGUMENT` (+ update `pkg/transports/json_utils_test.go`).
- [ ] Update / add tests in `pkg/transports/rest_status_code_test.go` and `pkg/transports/grpc_status_code_test.go`.
- [ ] Rename 0–16 constants to Go-idiomatic style (`StatusCodeOk`, `StatusCodeInvalidArgument`, …); decide `Canceled` spelling vs ST 2138-11's `CANCELED`.
- [ ] Remove deprecated HTTP-shaped constants (`CREATED`, `ACCEPTED`, `METHOD_NOT_ALLOWED`, `CONFLICT`, `UNPROCESSABLE_ENTITY`, `TOO_MANY_REQUESTS`, `BAD_GATEWAY`, `GATEWAY_TIMEOUT`, `BAD_REQUEST`).
- [ ] Sweep call sites in `pkg/transports`, `examples/*`, all tests.
- [ ] Update `StatusCode.md` to reflect new names; drop the legacy table.


sdks/go/pkg/catena/doc.go
```diff
@@ -0,0 +1,286 @@
...
+//	OK                    success                            HTTP 200 / 204; gRPC OK
+//	CANCELLED             caller stopped the work            HTTP 499; gRPC Canceled
+//	UNKNOWN               unclassified failure               HTTP 500; gRPC Unknown
+//	INVALID_ARGUMENT      bad client input                   HTTP 400 (per-route); gRPC InvalidArgument
+//	DEADLINE_EXCEEDED     time ran out                       HTTP 504; gRPC DeadlineExceeded
+//	NOT_FOUND             resource missing                   HTTP 404; gRPC NotFound
+//	ALREADY_EXISTS        duplicate create                   HTTP 409; gRPC AlreadyExists
+//	PERMISSION_DENIED     authenticated but not allowed      HTTP 403; gRPC PermissionDenied
+//	RESOURCE_EXHAUSTED    quota / rate limit                 HTTP 429; gRPC ResourceExhausted
+//	FAILED_PRECONDITION   wrong system state for the call    HTTP 400; gRPC FailedPrecondition
```
abrown-RV: put a note somewhere that `HTTP 412 Precondition Failed` while it has the same name as the `GRPC FailedPrecondition` it has a completely different meaning, and so isn't used here. Just cause, like me, people are likely to wonder XD

sdks/go/pkg/catena/doc.go
abrown-RV, line 1: turn this into a markdown file

sdks/go/pkg/catena/34CS-PCD-ST-2138-11-Catena-gRPC-Connection-Type-2026-03-24.pdf
abrown-RV, line 1: don't include the pdks, fine to reference them, but they should only live in one place so they don't have to be kept up to date

sdks/go/pkg/catena/status_code.go
```diff
@@ -107,39 +109,70 @@ const (
 	// UNAUTHENTICATED indicates the request does not have valid authentication credentials.
 	UNAUTHENTICATED StatusCode = 16
 
-	// REST-specific status codes (do not use in gRPC contexts)
+	// REST-specific status codes (legacy; see package doc Phase B)
```
abrown-RV: What's phase B?

sdks/go/pkg/catena/doc.go
```diff
@@ -0,0 +1,286 @@
...
+//     the route, not the handler's StatusCode.
+//
+// # Naming divergence vs ST 2138-11 (resolve in Phase D)
+//
+//	Spec name (ST 2138-11)     Catena Go constant     Standard gRPC name
+//	-------------------------  ---------------------  ---------------------
+//	CANCELED (one L)           CANCELLED              Canceled
+//	TIMEOUT EXCEEDED           DEADLINE_EXCEEDED      DeadlineExceeded
+//
+// Recommendation: align Catena with the standard gRPC names (Canceled,
```
abrown-RV: some of this text is more of a recommendation to you from your llm, not actually something that belongs in the end product. Clean stuff like that up, there are other instances.

abrown-RV (changes requested):
Very nice Zach. Appreciate your work here a lot. There a few AI inconsistancies to clean up. For example it refers to Phase B and D but doesn't define B and we skip over A and C XD.

Please take out the pdfs. And turn the big doc.go into a markdown instead. We don't have a docs dir yet, mostly we've just been putting them right next to the code, so something StatusCode.md or something.

I like the phase idea at least. But they need tickets to work against. and definition One of those phases should be renaming them to make the idiomatic `go` style as well.

Or just do it all in this one PR, we're still in the time where stuff like that is OK, with no one outside the team using it... yet. Although hopefully that could change soon.

abrown-RV:
Made an issue on -12 to add `400` to the list of REST codes. Since as is pointed out here, there are many instances of openapi paths that return it. https://github.com/SMPTE/st2138-12/issues/5

sdks/go/pkg/catena/doc.go
ZacharyDuaibesRV: Added a callout block above the code-by-code section in `StatusCode.md`:
HTTP 412 Precondition Failed is unrelated to `FailedPrecondition`. HTTP 412 shares a name with gRPC's `FailedPrecondition` but means something different: HTTP 412 is the response when a request's conditional headers (e.g. `If-Match`,` If-Unmodified-Since`) do not match. Catena does not use HTTP 412, and `FailedPrecondition `does not map to it.

sdks/go/pkg/catena/doc.go
ZacharyDuaibesRV: New file: _sdks/go/pkg/catena/StatusCode.md_. _doc.go_ is now a short package comment that points at it.
Would we like to remove _doc.go_ entirely?

sdks/go/pkg/catena/34CS-PCD-ST-2138-11-Catena-gRPC-Connection-Type-2026-03-24.pdf
ZacharyDuaibesRV: Removed. They are referenced by title and section number from _StatusCode.md_.

sdks/go/pkg/catena/status_code.go
ZacharyDuaibesRV: Dropped "Phase B" reference. Now reads `// REST-specific status code (legacy; see StatusCode.md)`

sdks/go/pkg/catena/doc.go
ZacharyDuaibesRV: cleaned

sdks/go/pkg/catena/doc.go
abrown-RV: ya remove entirely

ZacharyDuaibesRV: removed

abrown-RV: need to push the change

ZacharyDuaibesRV: done

abrown-RV (approved):
Looks good Zach, on to the next part! keep your checklist going :D

PR #1348, opened 2026-05-20
[GO] Authz system implemented

### LINKED ISSUE #1328: Keycloak
## 📖 Description
Implement keycloak server to validate JWT tokens

---

## 🎯 Goal / Outcome

---

## ✅ Definition of Done (DoD)

---

---

## 🔗 Related Issues / Links

---

---

### LINKED ISSUE #1329: Authz callback system
## 📖 Description
Implement a system to register a callback method to handle jwt token validation

---

## 🎯 Goal / Outcome
Implement a RegisterAccessHandler to register a callback
Business logic uses that to register callback given a jwt token, endpoint and fqoid

---

## ✅ Definition of Done (DoD)
- [ ] Edge cases are handled

---

---

## 🔗 Related Issues / Links

---

---

ctwarog-ross:
## 📖 Description
Authz implemented

Each transport now captures authz and metadata in transportContext
transportContext is parsed into handlerContext in each endpoint
handlerContext contains parsed token, scopes and metadata
handlerContext is exposed to Business logic through handler

Each endpoint has base authz that follow st2138 spec
Business logic can escalate if it chooses

Connections store HandlerContext when registered
BroadcastUpdate only broadcasts to connections with specified scopes

Global AccessHandler can be registered
Called every endpoint

OneOfEverything contains example of how to use authz

Config authzEnabled option added
authzEnabled bypasses all authz checking when False

---

## ✅ Definition of Done (DoD)

- [x] Each transport capture authz and metadata
- [x] Authz endpoints follow st2138 spec
- [ ] Tests are added
- [ ] add example to OneOfEverything 

---

## 🛠️ Implementation Notes

---

## 🔗 Related Issues / Links

Closes issue https://github.com/rossvideo/Catena/issues/1329

---
## 📝 Additional Context

---

## 🚧 Risks / Open Questions

---
## ✨ Updates


sdks/go/pkg/catena/server.go
```diff
@@ -291,10 +319,40 @@ func (s *server) Shutdown(ctx context.Context) {
...
+		Metadata:    transportContext.Metadata,
+	}
+	return handlerContext, StatusWithCode(OK, "")
+}
+
+func (s *server) isAccessAllowed(endpointType EndpointType, handlerContext HandlerContext) bool {
 	s.mu.Lock()
-	defer s.mu.Unlock()
-	return s.getSlotsLocked()
+	accessHandler := s.accessHandler
```
abrown-RV, line 439: should put a nil check either here or in RegisterAccessHandler... probably in register, then if it gets nil, that resets it back to the default? idk trying to avoid an err return

sdks/go/pkg/catena/server.go
```diff
@@ -291,10 +319,40 @@ func (s *server) Shutdown(ctx context.Context) {
...
+
+func (s *server) GetSlots(transportContext TransportContext) ([]uint16, StatusResult) {
+	handlerContext, err := s.getHandlerContext(transportContext)
+	if err.Code != OK {
+		return nil, err
+	}
+
+	if !s.isAccessAllowed(EndpointGetSlots, handlerContext) {
+		return nil, StatusWithCode(PERMISSION_DENIED, "Permission denied")
+	}
```
abrown-RV, line 464: add a unit test that registers a return false access handler and then hits all the Invokes so that we get coverage on all these permission denied returns

sdks/go/pkg/catena/server.go
```diff
@@ -401,87 +459,157 @@ func (s *server) RegisterHeartbeatHandler(slot uint16, handler HeartbeatHandler)
...
-func (s *server) InvokeGetDeviceHandler(slot uint16) (CatenaDevice, StatusResult) {
+func (s *server) RegisterAccessHandler(handler AccessHandler) {
+	s.mu.Lock()
+	s.accessHandler = handler
+	s.mu.Unlock()
+}
+
+func (s *server) InvokeGetDeviceHandler(slot uint16, transportContext TransportContext) (CatenaDevice, StatusResult) {
+	handlerContext, err := s.getHandlerContext(transportContext)
+	if err.Code != OK {
```
abrown-RV: having if err.Code != OK feels weird calling it `err`, maybe `status`? or `res`? I'd like to reserve `err` for specifically go `error` vars

sdks/go/pkg/transports/grpc_transport.go
```diff
@@ -207,6 +208,36 @@ func streamInterceptor(srv interface{}, ss grpc.ServerStream, info *grpc.StreamS
 	return sanitizeGRPCError(err)
 }
 
+func (t *GrpcTransport) retrieveMetadataFromContext(ctx context.Context) catena.TransportContext {
+	var accessToken string
+	requestMetadata, ok := metadata.FromIncomingContext(ctx)
+	metadataMap := make(map[string]string)
+	if ok {
+		if vals := requestMetadata.Get("authorization"); len(vals) > 0 {
+			accessToken = vals[0]
```
abrown-RV, line 218: hit with coverage

sdks/go/pkg/transports/grpc_transport.go
```diff
@@ -207,6 +208,36 @@ func streamInterceptor(srv interface{}, ss grpc.ServerStream, info *grpc.StreamS
...
+		// Convert metadata.MD map[string][]string to map[string]string by taking first value
+		metadataMap := make(map[string]string)
+		for key, vals := range requestMetadata {
+			if len(vals) > 0 {
+				metadataMap[key] = vals[0]
+			}
+		}
+	} else {
+		logger.Debug("No metadata found in context")
+	}
```
abrown-RV, line 227: not sure if possible to cover this

sdks/go/pkg/transports/rest_transport.go
```diff
@@ -156,6 +157,31 @@ func (t *RestTransport) RegisterFallbackHandler(handler FallbackHandler) {
...
 
+func (t *RestTransport) retrieveMetadataFromRequest(r *http.Request) catena.TransportContext {
+	transportContext := catena.TransportContext{
+		AccessToken: r.Header.Get("Authorization"),
+		Metadata: map[string]string{
+			"Detail-Level": r.Header.Get("Detail-Level"),
+			"User-Agent":   r.Header.Get("User-Agent"),
+		},
+	}
+	requestMetadata, ok := metadata.FromIncomingContext(r.Context())
```
abrown-RV: why is this here? this is for grpc. You should loop through r.Headers and copy over them over. Interesting question on which ones, cause there's all kind of boilerplate ones. Maybe just all cause who are we to say which headers BL does or doesn't want? or maybe we should interpose ourselves here and only copy over headers that are designated for st2138.... I don't have a good perspective here.

abrown-RV (changes requested):
looking just like we've talked about. All makes sense.

I think we should have a bigger (and later) discussion about what we do for metadata/headers... copy everything? allowlist/denylist? I don't have a good perspective here, I'll have to develop one XD

I also noted in one place that calling a var that holds our StatusResult `err` feels wrong. Because you then ask the "error" if it is actually ok... which is confusing. I think `err` should be reserved for var that are actual go `error` types. we should name it `status` or `result`/`res` But this is all over the code so pretty picky of me, I know...

sdks/go/pkg/catena/server.go
ctwarog-ross: We don't do that with any of the other Registers, do we add it for all of them?

sdks/go/pkg/transports/rest_transport.go
ctwarog-ross: We should copy all the headers I think. It isn't computationally expensive and better to have them than not

abrown-RV:
Makes sense to me, maybe a nice job for Zach.

ctwarog-ross:
bugbot run

sdks/go/pkg/catena/auth.go
```diff
@@ -0,0 +1,206 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("discover jwks endpoint: %w", err)
+		}
+
+		keyFunc, err := keyfunc.NewDefaultCtx(ctx, []string{jwksUrl})
+		if err != nil {
+			return nil, fmt.Errorf("create keyfunc: %w", err)
+		}
+		v.keyfunc = keyFunc.Keyfunc
+		v.validateFn = v.validateSignature
```
ctwarog-ross: very confusing naming, was getting mixed up between its child object options' member ValidateSignature

sdks/go/pkg/catena/auth.go
```diff
@@ -0,0 +1,206 @@
...
+			jwt.SigningMethodRS384.Alg(),
+			jwt.SigningMethodRS512.Alg(),
+			jwt.SigningMethodPS256.Alg(),
+			jwt.SigningMethodPS384.Alg(),
+			jwt.SigningMethodPS512.Alg(),
+			jwt.SigningMethodES256.Alg(),
+			jwt.SigningMethodES384.Alg(),
+			jwt.SigningMethodES512.Alg(),
+			jwt.SigningMethodEdDSA.Alg(),
+			jwt.SigningMethodHS256.Alg(),
```
ctwarog-ross: Just leave ES256 as the default as that is the Ross way to Auth. The business logic can specify whatever options they want otherwise.

sdks/go/pkg/catena/auth.go
```diff
@@ -0,0 +1,295 @@
...
+		return scopes, StatusWithCode(OK, "")
+	}
+
+	scopeClaim, ok := claims["scope"]
+	if !ok {
+		return scopes, StatusWithCode(OK, "")
+	}
+
+	scopeString, ok := scopeClaim.(string)
+	if !ok {
```
abrown-RV: in cpp if a token doesn't have any scopes it doesn't deny it on the face. It just lets you in with a token with 0 scopes... so you'll end up hitting roadblocks later on. So not sure what to say here. To match c++ exactly, we'd just return an empty set of scopes, but also I think just stopping here practically matches c++. The spec says every param shall have a scope assigned (either directly or inheirited), so a token with no scopes should not be able to do anything.

Either way this block should get hit with coverage

sdks/go/pkg/catena/connection_queue.go
```diff
@@ -164,6 +169,14 @@ func (cq *connectionQueue) notifyUpdate(update *protos.PushUpdates) {
 	}
 }
 
+func isValueUpdate(update *protos.PushUpdates) bool {
+	if update == nil {
+		return false
```
abrown-RV, line 174: coverage

sdks/go/examples/oneOfEverything/main.go
```diff
@@ -618,10 +635,16 @@ func main() {
 	for _, slot := range slotList {
 		p := slotParams[slot]
 
-		srv.RegisterGetValueHandler(slot, func(slot uint16, fqoid string) (catena.Value, catena.StatusResult) {
+		srv.RegisterGetValueHandler(slot, func(slot uint16, fqoid string, ctx catena.HandlerContext) (catena.Value, catena.StatusResult) {
 			logger.Info("GetValue", "slot", slot, "fqoid", fqoid)
 			key := normalizeFqoid(fqoid)
 
+			if slot == 1 && !ctx.HasScope(catena.ScopeMon) && !ctx.HasScope(catena.ScopeMonWrite) {
```
abrown-RV: should use `HasReadScope`

sdks/go/examples/oneOfEverything/main.go
```diff
@@ -655,7 +678,7 @@ func main() {
 	for _, slot := range slotList {
 		p := slotParams[slot]
 
-		srv.RegisterSetValueHandler(slot, func(value any, slot uint16, fqoid string) catena.StatusResult {
+		srv.RegisterSetValueHandler(slot, func(value any, slot uint16, fqoid string, ctx catena.HandlerContext) catena.StatusResult {
```
abrown-RV, line 679: should check write scopes

All the handlers should check scopes... or maybe we just make 1 or 2 devices that "use authz"? idk...

sdks/go/pkg/catena/server.go
```diff
@@ -46,21 +46,91 @@ import (
...
+	return ctx.hasAnyScope(catenaReadScopes...) || ctx.HasAnyWriteScope()
+}
+
+func (ctx HandlerContext) hasReadScope(scopeName string) bool {
+	if ctx.HasScope(scopeName) {
+		return true
+	}
+
+	switch scopeName {
+	case ScopeOp:
```
abrown-RV: I think this should just check if we have `scope+":w"` so that this works with other scopes.

sdks/go/pkg/catena/server.go
```diff
@@ -46,21 +46,91 @@ import (
...
+
+func (ctx HandlerContext) HasAnyWriteScope() bool {
+	return ctx.hasAnyScope(catenaWriteScopes...)
+}
+
+func (ctx HandlerContext) HasAnyReadScope() bool {
+	return ctx.hasAnyScope(catenaReadScopes...) || ctx.HasAnyWriteScope()
+}
+
+func (ctx HandlerContext) hasReadScope(scopeName string) bool {
```
abrown-RV: Suggestion: right now the read/write versions of scopes are treated as basically separate scopes except in a few places where it is acknowledged that write implies read, but thats only internal right now.

In the cpp sdk r/w are treated together. A scope is just the base part of it `st2138:mon` and then theres both `HasRead` and `HasWrite` both of these take the base scope name. HasRead then does nothing to it, and `HasWrite` adds the `:w` The benefit being that in the device model a param is given a scope. And then `GetValue` calls `HasRead` and `SetValue` can call `HasWrite` without having to transform the scope.

So if we were to follow this, we'd only need the 4 consts, just for the base names. Then we make `HasRead` check `HasScope`(probably make `hasScope` unexported) and also call `HasWrite`. And `HasWrite` just uses `hasScope(scope+":w")`

sdks/go/pkg/catena/server.go
```diff
@@ -291,10 +385,84 @@ func (s *server) Shutdown(ctx context.Context) {
...
+
+func (s *server) hasReadAccess(handlerContext HandlerContext) bool {
+	return !s.authzEnabled || handlerContext.HasAnyReadScope()
+}
+
+func (s *server) hasWriteAccess(handlerContext HandlerContext) bool {
+	return !s.authzEnabled || handlerContext.HasAnyWriteScope()
+}
+
+func (s *server) GetSlots(transportContext TransportContext) ([]uint16, StatusResult) {
```
abrown-RV, line 453: should check ReadAccess

sdks/go/pkg/catena/server.go
```diff
@@ -401,87 +569,181 @@ func (s *server) RegisterHeartbeatHandler(slot uint16, handler HeartbeatHandler)
...
+}
+
+func (s *server) InvokeGetDeviceHandler(slot uint16, transportContext TransportContext) (Device, StatusResult) {
+	handlerContext, res := s.resolveHandlerContext(transportContext)
+	if res.Code != OK {
+		return ReplyError[Device](res.Code, res.Error)
+	}
+
+	if !s.isAccessAllowed(EndpointGetDevice, handlerContext) {
+		return ReplyError[Device](PERMISSION_DENIED, "Permission denied")
```
abrown-RV, line 593: I feel like these checks should be flipped, why bother the BL if its going to be denied for not having read access?

sdks/go/pkg/catena/server.go
```diff
@@ -401,87 +569,181 @@ func (s *server) RegisterHeartbeatHandler(slot uint16, handler HeartbeatHandler)
...
 	}
 	// TODO: lookup default handler for slot
 	logger.Warning("GetValueHandler called - no handler registered for this slot", "slot", slot, "fqoid", fqoid)
 	return ReplyError[Value](NOT_FOUND, "fqoid "+fqoid+" not found at slot "+strconv.Itoa(int(slot)))
 }
 
-func (s *server) InvokeSetValueHandler(value any, slot uint16, fqoid string) StatusResult {
+func (s *server) InvokeSetValueHandler(value any, slot uint16, fqoid string, transportContext TransportContext) StatusResult {
+	handlerContext, res := s.resolveHandlerContext(transportContext)
+	if res.Code != OK {
```
abrown-RV, line 638: we need another 2 unittests that run through every invoke with an invalid token, or something that makes coverage hit this err branch

sdks/go/pkg/catena/server.go
```diff
@@ -401,87 +569,181 @@ func (s *server) RegisterHeartbeatHandler(slot uint16, handler HeartbeatHandler)
...
+		return res
+	}
+
+	if !s.isAccessAllowed(EndpointSetValue, handlerContext) {
+		return StatusWithCode(PERMISSION_DENIED, "Permission denied")
```
abrown-RV, line 643: and one with a accessHandler of return false to hit this err branch for each invoke

sdks/go/pkg/catena/server_test.go
```diff
@@ -759,28 +1092,282 @@ func TestServer_ExecuteCommand_Route(t *testing.T) {
...
+	}
+	if status.Code != OK {
+		t.Errorf("expected OK, got %v", status.Code)
+	}
+}
+
+func TestServer_AuthzDisabledAllowsRequestsWithoutToken(t *testing.T) {
+	srv := newTestServer(t, false)
+	srv.slots[0] = struct{}{}
+	srv.RegisterGetValueHandler(0, func(slot uint16, fqoid string, ctx HandlerContext) (Value, StatusResult) {
```
abrown-RV: should probably do a loop style test that goes through each endpoint

sdks/go/pkg/catena/auth.go
ctwarog-ross: You're right we should let this get caught after. My thought was we would only ever get here if authz was enabled, and any endpoint has a base requirement of any read scope. However there is other data in the HandlerContext that doesn't get added if this returns !OK, which may be useful later on such as token and metadata

sdks/go/examples/oneOfEverything/main.go
ctwarog-ross: GetValue already implicitly checks if it has a read scope. The purpose of this check is to show how to only allow for specific scopes in business logic. I'll leave it in but add a comment explaining

sdks/go/examples/oneOfEverything/main.go
ctwarog-ross: refer to comment above

sdks/go/pkg/catena/server.go
ctwarog-ross: HasScope needs to be public for the business logic to be able to specify an exact scope for something, rather than just have the option of has a read or a write scope. I think removing the write scopes const to replace it with a hardcoded string concatenation adds cpu cycles and doesn't improve code legibility. Strings are immutable in Go, so that means every single scope write check would be creating a new string by adding +":w" and garbage collecting it after.

abrown-RV (approved):
Wahhoooooo!

PR #1351, opened 2026-05-25
[GO] Unify StatusCode constants and align REST/gRPC mappers with SMPTE PCDs

### LINKED ISSUE #1322: [Go] Unify status codes

ZacharyDuaibesRV:
## Summary

Completes the StatusCode unification work started in the docs PR. This is the code half: renames the constants to idiomatic Go, deletes the deprecated HTTP-shaped values, rewrites both transport mappers around the SMPTE PCD vocabularies, and fixes a bug along the way.

### 1. Idiomatic Go renames

The 17 transport-neutral codes are now named in idiomatic Go style. The old `XYZ_XYZ` names are gone.

| Old                   | New                            |
|-----------------------|--------------------------------|
| `OK`                  | `StatusCodeOk`                 |
| `CANCELLED`           | `StatusCodeCancelled`  |
| `UNKNOWN`             | `StatusCodeUnknown`            |
| `INVALID_ARGUMENT`    | `StatusCodeInvalidArgument`    |
| `DEADLINE_EXCEEDED`   | `StatusCodeDeadlineExceeded`   |
| `NOT_FOUND`           | `StatusCodeNotFound`           |
| `ALREADY_EXISTS`      | `StatusCodeAlreadyExists`      |
| `PERMISSION_DENIED`   | `StatusCodePermissionDenied`   |
| `RESOURCE_EXHAUSTED`  | `StatusCodeResourceExhausted`  |
| `FAILED_PRECONDITION` | `StatusCodeFailedPrecondition` |
| `ABORTED`             | `StatusCodeAborted`            |
| `OUT_OF_RANGE`        | `StatusCodeOutOfRange`         |
| `UNIMPLEMENTED`       | `StatusCodeUnimplemented`      |
| `INTERNAL`            | `StatusCodeInternal`           |
| `UNAVAILABLE`         | `StatusCodeUnavailable`        |
| `DATA_LOSS`           | `StatusCodeDataLoss`           |
| `UNAUTHENTICATED`     | `StatusCodeUnauthenticated`    |

### 2. Deprecated HTTP-shaped constants removed

The 10 legacy REST-numeric constants (`CREATED`, `ACCEPTED`, `NO_CONTENT`, `BAD_REQUEST`, `METHOD_NOT_ALLOWED`, `CONFLICT`, `UNPROCESSABLE_ENTITY`, `TOO_MANY_REQUESTS`, `BAD_GATEWAY`, `SERVICE_UNAVAILABLE`, `GATEWAY_TIMEOUT`) are deleted. They have no basis in ST 2138-12 and the docs PR (#1345) already marked them deprecated.

Internal call sites were migrated to behaviour-preserving replacements:

| Removed                | Replacement                                       |
|------------------------|---------------------------------------------------|
| `CREATED`, `ACCEPTED`, `NO_CONTENT` | `StatusCodeOk` (REST layer chooses 204 by route) |
| `BAD_REQUEST`, `UNPROCESSABLE_ENTITY` | `StatusCodeInvalidArgument` |
| `CONFLICT`             | `StatusCodeAlreadyExists`                         |
| `TOO_MANY_REQUESTS`    | `StatusCodeResourceExhausted`                     |
| `SERVICE_UNAVAILABLE`  | `StatusCodeUnavailable`                           |
| `GATEWAY_TIMEOUT`      | `StatusCodeDeadlineExceeded`                      |
| `METHOD_NOT_ALLOWED`, `BAD_GATEWAY` | deleted - transport-only, see below |

### 3. Mapper rewrites (`pkg/transports/{rest,grpc}_transport.go`)

- **`ToGRPCCode`** is trimmed to the ST 2138-11 §6.2 set (codes 0–16) plus a `default → codes.Unknown` fallback. All the REST-numeric special cases are gone.
- **`ToHTTPStatus`** drops the REST-numeric pass-through block. The default failure set is now exactly ST 2138-12 §7.3 `{401, 403, 404, 500, 503}` plus the per-case 200/400/409/429/499/501/504 mappings derived from the 17-code vocabulary. Unrecognised codes fall through to 500.

### 4. Bug fix: `BAD_REQUEST` → HTTP 500

`pkg/transports/json_utils.go` was returning `catena.BAD_REQUEST` for malformed Content-Type, malformed JSON bodies, and unreadable bodies. With the old mapper, `BAD_REQUEST` was handled by the REST-numeric pass-through block - but `ToGRPCCode` had no case for it, so gRPC callers got `codes.Unknown`. With the new mapper, the pass-through block is gone entirely; the same fall-through that bites gRPC today would bite REST too.

Fix: all four `BAD_REQUEST` returns in `json_utils.go` are now `StatusCodeInvalidArgument`. Bad JSON now correctly produces HTTP 400 and gRPC `InvalidArgument` instead of HTTP 500 / gRPC `Unknown`.

### 5. Method enforcement moved out of `StatusCode`

The 7 router-level `"only X allowed"` checks in `rest_transport.go` used to emit `METHOD_NOT_ALLOWED` and rely on the REST-numeric pass-through to land on HTTP 405. With the pass-through gone, those checks now call a new helper:

```go
func writeHTTPMethodNotAllowed(w http.ResponseWriter, msg string)

### CODE REVIEW

File: sdks/go/examples/getSetValue/getsetvalue.go [hunk 1]
```diff
@@ -117,34 +117,34 @@ func RegisterBasicParamHandlers(srv catena.CatenaServer, params *sync.Map, slot
...
 		if reflect.TypeOf(val) != reflect.TypeOf(value) {
 			logger.Error("SetParam type mismatch", "slot", slot, "fqoid", fqoid)
-			return catena.StatusWithCode(catena.INVALID_ARGUMENT, "type mismatch")
+			return catena.StatusWithCode(catena.StatusCodeInvalidArgument, "type mismatch")
 		}
 		params.Store(fqoid, value)
 
 		srv.BroadcastUpdate(slot, fqoid, value)
-		return catena.StatusWithCode(catena.NO_CONTENT, "")
+		return catena.StatusWithCode(catena.StatusCodeOk, "")
```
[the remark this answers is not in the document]
ZacharyDuaibesRV: - Added writeHTTPStatusResultNoBody helper in rest_transport.go. On StatusCodeOk + empty error it writes HTTP 204; on any other code it delegates to writeHTTPStatusResult so the failure mapping is unchanged.
- handleValueEndpoint PUT branch now calls writeHTTPStatusResultNoBody, restoring ST 2138-12 §7.7.2 SetValue success = 204.
- TestRestTransport_SetValue_Route updated from StatusOK to StatusNoContent (the StatusOK assertion was masking this bug).
- Added TestWriteHTTPStatusResultNoBody_SuccessIs204 and TestWriteHTTPStatusResultNoBody_ErrorPreservesMapping as regression tests. 

SetValues and the language-pack mutation routes aren't wired in the REST transport yet, but the helper is in place for when they are.

sdks/go/pkg/transports/rest_transport.go
```diff
@@ -187,6 +187,19 @@ func writeHTTPResult(w http.ResponseWriter, result catena.StatusResult, value in
...
+// 405 is not in ST 2138-12, but is emitted here for HTTP correctness when a
+// route is reached with the wrong method.
+func writeHTTPMethodNotAllowed(w http.ResponseWriter, msg string) {
+	w.WriteHeader(http.StatusMethodNotAllowed)
+	if catena.IsDev() {
+		_ = json.NewEncoder(w).Encode(map[string]string{"error": msg})
+	} else {
+		_ = json.NewEncoder(w).Encode(map[string]string{"error": http.StatusText(http.StatusMethodNotAllowed)})
+	}
+}
```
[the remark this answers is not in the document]
ZacharyDuaibesRV: moved the writeHTTPStatusResult comment back to its function, and the new writeHTTPMethodNotAllowed now has only its own doc block.

sdks/go/pkg/catena/status_code_test.go
```diff
@@ -44,40 +44,26 @@ import (
 )
 
 func TestStatusCodeConstants(t *testing.T) {
-	// Verify gRPC-compatible codes are in the expected range (0-16)
 	grpcCodes := []StatusCode{
-		OK, CANCELLED, UNKNOWN, INVALID_ARGUMENT, DEADLINE_EXCEEDED,
-		NOT_FOUND, ALREADY_EXISTS, PERMISSION_DENIED, RESOURCE_EXHAUSTED,
-		FAILED_PRECONDITION, ABORTED, OUT_OF_RANGE, UNIMPLEMENTED,
-		INTERNAL, UNAVAILABLE, DATA_LOSS, UNAUTHENTICATED,
+		StatusCodeOk, StatusCodeCanceled, StatusCodeUnknown, StatusCodeInvalidArgument, StatusCodeDeadlineExceeded,
```
abrown-RV: I think this test is just unnecessary now

sdks/go/pkg/catena/StatusCode.md
```diff
@@ -30,218 +30,195 @@ to add `400` to §7.3 Table 1 is tracked in
...
+### `StatusCodeCanceled`
 
 - **Cause:** the operation was canceled by either the client or the device.
 - **When to return:** you detected explicit cancellation. Not for plain
   timeouts.
 - **What clients infer:** the work was abandoned, no result.
-- **Naming:** ST 2138-11 spells this `CANCELED` (one L); standard gRPC uses
-  `Canceled`. The Catena Go constant is currently `CANCELLED`.
+- **Naming:** ST 2138-11 spells this `CANCELED` (one L); standard gRPC and
+  the Catena Go constant both use `Canceled`.
```
abrown-RV: now that they're both the same, is this note really needed? I wish we could spell it the proper *Canadian* way, 2 Ls but we should stay consistent with the standard

sdks/go/oneOfEverything
abrown-RV, line 1: whoops you included a bin file

abrown-RV (dismissed):
a few little things

sdks/go/pkg/catena/status_code_test.go
ZacharyDuaibesRV: ya its redundant since coverage happens in TestToGRPCCodes anyways

sdks/go/pkg/catena/StatusCode.md
ZacharyDuaibesRV: removed note. I wish we spelled it the Canadian way too lol but better to stay consistent with the standard

sdks/go/oneOfEverything
ZacharyDuaibesRV: whoops

abrown-RV (commented):
Ready to approve but you've got merge conflicts and it will just make me re-approve after resovling, so resolve those first please.

ctwarog-ross (dismissed):

ctwarog-ross (approved):

PR #1353, opened 2026-05-29
[GO] Constraint Creation Helpers

### LINKED ISSUE #1305: Constraint creation helpers

nelgd:
## 📖 Description
Add a `Constraint` wrapper type with factory functions for every constraint kind defined in the Catena protocol. Update `Param.WithConstraint` to accept the new `*Constraint` type instead of raw `*protos.Constraint`, and migrate all existing tests to use the new builders.

---

## 🎯 Goal / Outcome
SDK users can construct any constraint without importing or manually wiring up `protos` types. Constraint creation is reduced from 5–10 lines of nested proto struct literals to a single, type-safe factory call (e.g. `NewConstraintInt32Choice([]Int32Choice{{1, name}})`), consistent with the existing `NewParamXxx` builder pattern.

---

## ✅ Definition of Done (DoD)
- [x] `Constraint` wrapper struct added with `Proto` field
- [x] Go-native input types exported: `Int32Choice`, `StringStringChoice`, `AlarmEntry`, `AlarmSeverity`
- [x] `Param.WithConstraint` signature updated from `*protos.Constraint` → `*Constraint`
- [x] All 18 existing constraint tests in `param_test.go` migrated to use builders
- [x] Dedicated `constraint_test.go` with tests for every factory (including edge cases: empty slices, multi-language names, nil descriptions, strict flag, severity mapping)
- [x] All tests pass (`go test ./pkg/catena/...`)

---

- `Constraint` follows the same wrapper pattern as `Param` and `Value` (struct with a `Proto` field).
- `AlarmSeverity` is a type alias (`=`) for `protos.Alarm_Severity`, not a new type, so it's directly interchangeable with the proto enum without casts.
- Range constraints have two variants: a 3-arg version (`min, max, step`) and a `WithDisplay` 5-arg version that also sets `display_min`/`display_max`.
- `NewConstraintStringChoice` uses variadic `choices ...string` for ergonomic inline calls.
- `WithConstraint(nil)` remains a safe no-op; the nil check now also covers `c.Proto == nil`.

---

## 🔗 Related Issues / Links

Closes issue #1305


sdks/go/pkg/catena/param.go
```diff
@@ -240,17 +240,17 @@ func (cp *Param) WithValue(v Value) *Param {
...
 }
 
-func (cp *Param) WithConstraint(c *protos.Constraint) *Param {
-	if c == nil {
+func (cp *Param) WithConstraint(c *Constraint) *Param {
+	if c == nil || c.Proto == nil {
 		return cp
 	}
-	if !isConstraintValidForParam(c, cp.Proto.Type) {
+	if !isConstraintValidForParam(c.Proto, cp.Proto.Type) {
```
ctwarog-ross: Change isConstraintValidForParam to take in Constraint instead of protos.Constraint. Then have it unwrap the proto.Constraint inside isConstraintValidForParam

ctwarog-ross (changes requested):

ctwarog-ross (approved):

PR #1358, opened 2026-06-02
[GO] Examples cleanup

### LINKED ISSUE #1357: [Go] Example revamp
## 📖 Description
We need to cleanup the examples as well as make them more useful for adopters.
- one-of-everything needs its authz logic fixed
- one-of-everything needs struct example
- hello_world basic catena device setup example
- remove all the other examples

---

## 🎯 Goal / Outcome
An adopter should be able to get set up easily with the hello_world example, as well as have any possible functionality shown in one_of_everything.

---

## ✅ Definition of Done (DoD)
- [ ] Edge cases are handled

---

---

## 🔗 Related Issues / Links

---

---

ctwarog-ross:
## 📖 Description
Closes Issue #1357 

Removed all examples other than one-of-everything
one-of-everything has struct implementation
one-of-everything authz fixed
one-of-everything now has all the implemented endpoint edge cases

hello_world example
hello_world contains a string with get/setvalue endpoint exposed
runs on rest and grpc

---

## ✅ Definition of Done (DoD)

---

## 🛠️ Implementation Notes

---

## 🔗 Related Issues / Links

---
## 📝 Additional Context

---

## 🚧 Risks / Open Questions

---
## ✨ Updates


sdks/go/examples/hello_world/main.go
```diff
@@ -0,0 +1,190 @@
...
+
+func buildDevice() map[string]any {
+	return map[string]any{
+		"slot":              uint32(slot),
+		"detail_level":      catena.DetailLevelFull,
+		"multi_set_enabled": true,
+		"subscriptions":     true,
+		"access_scopes":     []string{"st2138:mon", "st2138:op", "st2138:cfg", "st2138:adm"},
+		"default_scope":     "st2138:op",
+		"params": map[string]any{
```
abrown-RV, line 84: should get the mandatory params... but this will be taken care of by the helpers, so meh?

sdks/go/examples/hello_world/main.go
```diff
@@ -0,0 +1,190 @@
...
+		logger.Info("GetDevice", "slot", slot)
+		device, err := catena.ToDevice(buildDevice())
+		if err != nil {
+			logger.Error("Failed to convert device", "slot", slot, "error", err)
+			return catena.ReplyError[catena.Device](catena.INTERNAL, "failed to convert device")
+		}
+		return catena.Reply(device)
+	})
+
+	srv.RegisterGetValueHandler(slot, func(slot uint16, fqoid string, ctx catena.HandlerContext) (catena.Value, catena.StatusResult) {
```
abrown-RV, line 120: should check for `op` scope, since that's the default, above and `hello_world` does not specify a different scope. Or just hardcode AuthzEnabled to false since this is a basic example

sdks/go/examples/hello_world/main.go
```diff
@@ -0,0 +1,190 @@
...
+			return catena.StatusWithCode(catena.NOT_FOUND, "parameter not found: "+fqoid)
+		}
+
+		stringValue, ok := value.(string)
+		if !ok {
+			return catena.StatusWithCode(catena.INVALID_ARGUMENT, "hello_world requires a string value")
+		}
+
+		helloWorld.Set(stringValue)
+		srv.BroadcastUpdate(slot, helloWorldOID, stringValue, catena.ScopeMon)
```
abrown-RV: should check for `ScopeOp` write at the top of `SetValue`, and broadcast on `ScopeOp`

sdks/go/examples/oneOfEverything/main.go
```diff
@@ -157,6 +158,252 @@ func stringParamValue(p *sync.Map, key string, fallback string) map[string]any {
...
+		if !ok {
+			return nil, false
+		}
+		params = childParamMap(paramInfo)
+	}
+
+	return paramInfo, true
+}
+
+func paramInfosForRequest(params map[string]any, oid string, recursive bool) ([]catena.ParamInfo, catena.StatusResult) {
```
abrown-RV: all this extra plumbing for the param infos is maybe an indicator that we could be doing them nicer, but something to think about.

sdks/go/examples/oneOfEverything/main.go
abrown-RV, line 1: We should get better commenting on these examples for them to be good for people to look at

abrown-RV (commented):
Do you want to spit up oneofeverthing's main.go into different files for the different endpoints here? or is that another PR?

sdks/go/examples/hello_world/main.go
ctwarog-ross: Authzenabled to false is a good idea

sdks/go/examples/hello_world/main.go
ctwarog-ross: We'll leave authz out of the hello_world example

ctwarog-ross:
bugbot run

ctwarog-ross:
bugbot run

sdks/go/examples/hello_world/main.go
abrown-RV: makes sense, should just pass "" for the scope then? ... we need a good convention for what to pass when authz is disabled. It doesn't matter but like empty string at least saves some bytes? XD

sdks/go/pkg/catena/param_info.go
```diff
@@ -89,6 +91,157 @@ func ToParamInfo(m map[string]any) (ParamInfo, error) {
...
+			name.With(lang, s)
+		}
+	}
+	if len(name) == 0 {
+		return nil
+	}
+	return name
+}
+
+func paramArrayLength(paramInfo map[string]any) uint32 {
```
abrown-RV, line 134: where is this coming from, array_length and array_len aren't valid fields in a device definition?

sdks/go/examples/oneOfEverything/access_handler.go
```diff
@@ -0,0 +1,22 @@
...
+func registerAccessHandler(srv catena.Server) {
+	// AccessHandler is the coarse authorization gate for SDK endpoints. It runs
+	// after the transport has built HandlerContext, so adopters can check scopes,
+	// identity, or endpoint type before any endpoint-specific handler runs. Use
+	// this for cross-cutting policy; use individual handlers for per-param rules.
+	srv.RegisterAccessHandler(func(endpointType catena.EndpointType, ctx catena.HandlerContext) bool {
+		logger.Info("Access request", "endpointType", endpointType)
+
+		// Shows how to restrict getPopulatedSlots to only op and adm read scopes.
+		if endpointType == catena.EndpointGetSlots {
```
abrown-RV: I don't think we should do this. Struggling to come up with a good demo to put here, maybe its enough to just have it return true and have all the comments?

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+		} else if strings.HasPrefix(key, "sample_float") && key == "sample_float" {
+			v, ok = state.sampleFloat, true
+		}
+		state.mu.RUnlock()
+
+		return replyValue(fqoid, v, ok)
+	})
+
+	// Slot 0: writes use the same manual switch strategy as reads. Writes are
+	// restricted to callers with configuration-scope write access.
```
abrown-RV: might be good to have examples of params with individual scopes instead of blanket accross a whole slot.

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+
+		var v any
+		var ok bool
+		if strings.HasPrefix(key, "product") {
+			v, ok = state.slotTwoProductValue(key)
+			return replyValue(fqoid, v, ok)
+		}
+
+		state.mu.RLock()
+		if strings.HasPrefix(key, "volume") && key == "volume" {
```
abrown-RV: isn't the HasPrefix redundant?

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+		} else if strings.HasPrefix(key, "struct_example") {
+			switch key {
+			case "struct_example":
+				v, ok = cloneStringAnyMap(state.structExample), true
+			case "struct_example/number":
+				v, ok = state.structExample["number"], true
+			case "struct_example/text":
+				v, ok = state.structExample["text"], true
+			}
+		} else if strings.HasPrefix(key, "sample_struct_variant_array") && key == "sample_struct_variant_array" {
```
abrown-RV: all of the rest of these need to be broken down by sub-params like `struct_example`

friendly reminder that array has /index

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+			v, ok = cloneStructVariantArray(state.sampleStructVariantArray), true
```
abrown-RV: why clone? In catena.ToValue() it will be cloned there. Right? putting it into a proto clones it? maybe not... ok looking at the proto code right, its only the scalars that are copied by value, and therefore "cloned".

Maybe it would be helpful to make tovalue clone? since that would be a pretty safe assumption that you don't want it to follow the changing reference? or have it be an option or add a ToValueClone or something?

abrown-RV: why clone? In catena.ToValue() it will be cloned there. Right? putting it into a proto clones it? maybe not... ok putting it into the proto doesn't clone it necessarily, but for the most part ToProto will end up cloning everything by how it walks through the heirarchy. I think only for the arrays, it doesn't actually clone.

Maybe it would be helpful to make tovalue consistantly clone? since that would be a pretty safe assumption that you don't want it to follow the changing reference?

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+			v, ok = state.sampleFloat, true
+		}
+		state.mu.RUnlock()
+
+		return replyValue(fqoid, v, ok)
+	})
+
+	// Slot 0: writes use the same manual switch strategy as reads. Writes are
+	// restricted to callers with configuration-scope write access.
+	srv.RegisterSetValueHandler(0, func(value any, slot uint16, fqoid string, ctx catena.HandlerContext) catena.StatusResult {
```
abrown-RV: the setvalues should check for `product` and reject `PermissionDenied` (its readonly)

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+		if value == nil {
+			logger.Error("SetValue nil value received", "slot", slot, "fqoid", fqoid)
+			return catena.StatusWithCode(catena.StatusCodeInvalidArgument, "nil value received")
+		}
+
+		key := normalizeFqoid(fqoid)
+		switch key {
+		case "running":
+			logger.Warning("SetValue rejected for running param", "slot", slot, "fqoid", fqoid)
+			return catena.StatusWithCode(catena.StatusCodeInvalidArgument, "use start/stop commands to change running state")
```
abrown-RV: this should be PermissionDenied, `running` is readonly

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+		case "counter":
+			expectedType := reflect.TypeOf(counter.GetValue())
```
abrown-RV: does `if v, ok := value.(int32); ok {` work instead of reflecting? I assume reflecting is slower than casting

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+			state.sampleIntArray = append([]int32(nil), value.([]int32)...)
+			state.mu.Unlock()
+		default:
+			logger.Error("SetValue param not found", "slot", slot, "fqoid", fqoid)
+			return catena.StatusWithCode(catena.StatusCodeNotFound, "param not found: "+fqoid)
+		}
+
+		logger.Info("Parameter updated", "fqoid", fqoid, "value", value)
+		srv.BroadcastUpdate(slot, fqoid, value, catena.ScopeMon)
```
abrown-RV: should be `cfg` scope

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+
+		key := normalizeFqoid(fqoid)
+		switch key {
+		case "resolution", "brightness", "contrast", "saturation":
+			current, ok := state.slotOneParams.Load(key)
+			if !ok {
+				logger.Error("SetValue param not found", "slot", slot, "fqoid", fqoid)
+				return catena.StatusWithCode(catena.StatusCodeNotFound, "param not found: "+fqoid)
+			}
+			if reflect.TypeOf(current) != reflect.TypeOf(value) {
```
abrown-RV: reflection makes sense here 👍

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+
+		logger.Info("Parameter updated", "fqoid", fqoid, "value", value)
+		srv.BroadcastUpdate(slot, fqoid, value, catena.ScopeMon)
+		return catena.StatusWithCode(catena.StatusCodeOk, "")
+	})
+
+	// Slot 2: writes use prefix-based dispatch and keep product params read-only.
+	// Like the slot 2 read handler, this example does not add a per-handler
+	// scope check so adopters can compare scoped and unscoped endpoint patterns.
+	srv.RegisterSetValueHandler(2, func(value any, slot uint16, fqoid string, ctx catena.HandlerContext) catena.StatusResult {
```
abrown-RV: I think technically the spec does not allow for unscoped params. Currently the schema doesn't actually enforce it so should check with John, but I think scopes are unescapable, at least when authz is on.

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+			return catena.StatusWithCode(catena.StatusCodeNotFound, "param not found: "+fqoid)
+		}
+		if key != "sample_binary" && reflect.TypeOf(current) != reflect.TypeOf(value) {
+			state.mu.Unlock()
+			logger.Error("SetValue type mismatch", "slot", slot, "fqoid", fqoid,
+				"expected", reflect.TypeOf(current), "got", reflect.TypeOf(value))
+			return catena.StatusWithCode(catena.StatusCodeInvalidArgument, "type mismatch")
+		}
+
+		if strings.HasPrefix(key, "volume") && key == "volume" {
```
abrown-RV: doubled code?

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,306 @@
...
+		} else if strings.HasPrefix(key, "struct_example") {
+			switch key {
+			case "struct_example":
+				current, ok = state.structExample, true
+			case "struct_example/number":
+				current, ok = state.structExample["number"], true
+			case "struct_example/text":
+				current, ok = state.structExample["text"], true
+			}
+		} else if strings.HasPrefix(key, "sample_struct_variant_array") && key == "sample_struct_variant_array" {
```
abrown-RV: these complex types all need to handle having their sub-items set directly

sdks/go/examples/oneOfEverything/command_handler.go
```diff
@@ -0,0 +1,57 @@
+package main
+
+import (
+	"github.com/rossvideo/catena/sdks/go/pkg/catena"
+	"github.com/rossvideo/catena/sdks/go/pkg/logger"
+)
+
+func registerCommandHandler(srv catena.Server, counter *CounterState, broadcastRunning func(), counterScope string) {
```
abrown-RV, line 8: commands have scope too

sdks/go/examples/oneOfEverything/device_handler.go
```diff
@@ -0,0 +1,364 @@
...
+			WithName(catena.NewPolyglotText("en", "Contrast")).ToMap()
+		saturationParam := catena.NewParamInt32(saturation).
+			WithName(catena.NewPolyglotText("en", "Saturation")).ToMap()
+		return map[string]any{
+			"slot":              uint32(1),
+			"detail_level":      catena.DetailLevelFull,
+			"multi_set_enabled": false,
+			"subscriptions":     true,
+			"access_scopes":     []string{"st2138:mon", "st2138:op", "st2138:cfg", "st2138:adm"},
+			"default_scope":     "st2138:op",
```
abrown-RV: its mon in value_handlers

sdks/go/examples/oneOfEverything/device_handler.go
```diff
@@ -0,0 +1,364 @@
...
+			WithName(catena.NewPolyglotText("en", "Add 10 to Counter")).ToMap()
+		resetCommand := catena.NewParamEmpty().
+			WithName(catena.NewPolyglotText("en", "Reset Counter")).ToMap()
+		return map[string]any{
+			"slot":              uint32(0),
+			"detail_level":      catena.DetailLevelFull,
+			"multi_set_enabled": true,
+			"subscriptions":     true,
+			"access_scopes":     []string{"st2138:mon", "st2138:op", "st2138:cfg", "st2138:adm"},
+			"default_scope":     "st2138:op",
```
abrown-RV: value_handlers has it as cfg

sdks/go/examples/oneOfEverything/device_handler.go
```diff
@@ -0,0 +1,364 @@
...
+		}, true
+	case 2:
+		// Slot 2: prefix-dispatched identity/audio params plus FLOAT32, arrays,
+		// BINARY, STRUCT_VARIANT, STRUCT_ARRAY, STRUCT_VARIANT_ARRAY examples.
+		state.mu.RLock()
+		productValue := cloneStringAnyMap(state.slotTwoProduct)
+		volume := state.volume
+		muted := state.muted
+		deviceName := state.deviceName
+		structExample := cloneStringAnyMap(state.structExample)
```
abrown-RV: same cloning thoughts as in value_handlers

sdks/go/examples/oneOfEverything/device_handler.go
```diff
@@ -0,0 +1,364 @@
...
+		sampleFloat := state.sampleFloat
+		sampleFloatArray := cloneFloat32Slice(state.sampleFloatArray)
+		sampleStringArray := cloneStringSlice(state.sampleStringArray)
+		sampleBinary := cloneBytes(state.sampleBinary)
+		sampleStructVariant := cloneStructVariant(state.sampleStructVariant)
+		sampleStructArray := cloneStructArray(state.sampleStructArray)
+		sampleStructVariantArray := cloneStructVariantArray(state.sampleStructVariantArray)
+		state.mu.RUnlock()
+		productParam := catena.NewParamStruct(productValue).
```
abrown-RV, line 254: this is good illustration of actually using the param helpers... and I'm not totally happy... need to think more...

sdks/go/examples/oneOfEverything/heartbeat_handler.go
```diff
@@ -0,0 +1,23 @@
...
+	srv.RegisterHeartbeatHandler(1, func(slot uint16) {
+		// Example ticking on the canonical "product/version" param.
+		srv.BroadcastUpdate(1, "product/version", "1.0.0", catena.ScopeMon)
+	})
+	srv.RegisterHeartbeatHandler(2, func(slot uint16) {
+		// Example ticking on a different param to show any can be used for heartbeats.
+		state.mu.RLock()
+		volume := state.volume
+		state.mu.RUnlock()
+		srv.BroadcastUpdate(2, "volume", volume, catena.ScopeMon)
```
abrown-RV, line 24: I think broadcast update should also be made threadsafe with making sure it clones the value its given (it already mostly does cause it also calls ToProto) but then you can just lock and unlock safely around BroadcastUpdate directly

sdks/go/examples/oneOfEverything/main.go
```diff
@@ -911,13 +536,27 @@ func main() {
 
 		logger.Info("Web UI available at:")
 		logger.Info("  http://localhost:8080/")
+		logger.Info("Asset endpoints:")
```
abrown-RV: these endpoints lists for grpc/rest should either do all the endpoint or none of them

abrown-RV (changes requested):
That sure is a lot of code, looking really good. I'm sure after we've talked all this to death there will be more XD

abrown-RV:
we should remove `normalizeFqoid` from everywhere, since it should be properly doing it now.

sdks/go/examples/oneOfEverything/access_handler.go
ctwarog-ross: that's a good point, especially since we use getpopulatedslots for other things like generating jwt in postman. Having our flagship demo device having getPopulatedSlots be locked behind specific authz is not ideal

sdks/go/examples/oneOfEverything/value_handlers.go
ctwarog-ross: The boilerplate for this got crazy, not sure what to do about this

sdks/go/examples/oneOfEverything/value_handlers.go
ctwarog-ross: You are correct, reflection takes significantly longer

sdks/go/examples/oneOfEverything/value_handlers.go
ctwarog-ross: its not technically unscoped, the comment is a bit misleading, it just means any read/write scope, not particular ones like in slot 0/1

sdks/go/examples/oneOfEverything/value_handlers.go
ctwarog-ross: Thewre have been changes to ToProto and unit tests to ensure it always deep copies data. Cloning help methods removed from example.

sdks/go/examples/hello_world/main.go
```diff
@@ -0,0 +1,183 @@
...
+	return s.value
+}
+
+func (s *helloWorldState) Set(value string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.value = value
+}
+
+func normalizeFqoid(fqoid string) string {
```
abrown-RV: remove here too

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,611 @@
...
+		var v any
+		var ok bool
+		switch fqoid {
+		case "product", "product/name", "product/vendor", "product/version":
+			v, ok = state.slotZeroProductValue(fqoid)
+		case "counter":
+			v, ok = counter.GetValue(), true
+		case "running":
+			v, ok = counter.RunningInt32(), true
+		case "sample_int_array":
```
abrown-RV: should this array go live with the other arrays in slot 2?

sdks/go/examples/oneOfEverything/value_handlers.go
abrown-RV: That's still not to spec: every param must only have a single scope associtated with it, either directly on the param or inheirted from the parent/device default. The idea is that users have multiple scopes assigned to them to acheive broad access instead of params having multiple scopes. It does feel backwards at first, but thats the way it is. HOWEVER, here in L1 it does get muddy, and I fully expect devices will exist that don't really follow this. It just conflicts with the DeviceRequest/GetParam reporting the single scope on a param. So all that being said I don't know what to say for doing scopes here in the handlers, maybe it really is better to just do the device level default, check that at the top of the handler and call it a day, cause otherwise *every* if branch needs its own check. If we do that route, that the GetValue also needs the same scope checks and the DeviceRequest should report the same scope for each.

abrown-RV (changes requested):
All these var types in slot 2 is going to be great for the Check Your Catena tool

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,611 @@
...
+			if idx, field, _, okPath := parseArrayElementPath(fqoid, "sample_int_array"); okPath && field == "" {
+				arr := state.sampleIntArrayValue()
+				if idx < len(arr) {
+					v, ok = arr[idx], true
+				}
+			}
+		}
+
+		return replyValue(fqoid, v, ok)
+	})
```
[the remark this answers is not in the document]
ctwarog-ross: / was removed from spec. Disregard

sdks/go/examples/oneOfEverything/command_handler.go
```diff
@@ -0,0 +1,58 @@
...
+			return catena.CommandReply(val)
+		case "reset":
+			counter.Reset()
+			logger.Info("Counter reset", "value", counter.GetValue())
+			srv.BroadcastUpdate(0, "counter", counter.GetValue(), counterScope)
+			val, _ := catena.ToValue(counter.GetValue())
+			return catena.CommandReply(val)
+		default:
+			logger.Warning("Command not found", "slot", slot, "command", commandFqoid)
+			return catena.CommandExceptionResult("Invalid Command", "Command not found: "+commandFqoid, nil)
```
abrown-RV: should return a `StatusCodeNotFound` this is a protocol error not a command error.

sdks/go/examples/oneOfEverything/command_handler.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/rossvideo/catena/sdks/go/pkg/logger"
+)
+
+func registerCommandHandler(srv catena.Server, counter *CounterState, broadcastRunning func(), counterScope string) {
+	// Slot 0: switch per command FQOID and build command responses manually.
+	// This handler does not reduce allowed scopes beyond the base level of
+	// any command scope (mon, op, conf, or adm), however it broadcasts updates
+	// to the counter value only to specified counterScope.
+	srv.RegisterExecuteCommandHandler(0, func(slot uint16, commandFqoid string, payload any, ctx catena.HandlerContext) (catena.CommandResult, catena.StatusResult) {
+		logger.Info("ExecuteCommand", "slot", slot, "command", commandFqoid)
```
abrown-RV, line 14:
```suggestion
		logger.Info("ExecuteCommand", "slot", slot, "command", commandFqoid)
		if !ctx.HasWriteScope(counterScope) {
			logger.Warning("Unauthorized command execution attempt", "slot", slot, "command", commandFqoid)
			return catena.ReplyError[catena.CommandResult](catena.StatusCodePermissionDenied, "Caller does not have required scope for this command")
		}
```
Should check scope

sdks/go/pkg/catena/value.go
```diff
@@ -126,11 +127,11 @@ func ToProto(v any) (*protos.Value, StatusResult) {
 	case string:
 		return &protos.Value{Kind: &protos.Value_StringValue{StringValue: val}}, StatusResult{Code: StatusCodeOk}
 	case []int32:
-		return &protos.Value{Kind: &protos.Value_Int32ArrayValues{Int32ArrayValues: &protos.Int32List{Ints: val}}}, StatusResult{Code: StatusCodeOk}
+		return &protos.Value{Kind: &protos.Value_Int32ArrayValues{Int32ArrayValues: &protos.Int32List{Ints: slices.Clone(val)}}}, StatusResult{Code: StatusCodeOk}
```
abrown-RV, line 138: ToValue and BroadcastUpdate should get documentation indicating their cloning behaviour. Here's what copilot gave me:

```go
// ToValue converts v into a Value with its own copied backing data.
//
// For supported input types, the returned Value does not alias caller-owned
// slices, maps, or payload buffers, so callers may release locks and freely
// mutate or discard the original input after ToValue returns.
//
// Callers must still prevent concurrent mutation of v while this function is
// running.
func ToValue(v any) (Value, StatusResult) {
```

sdks/go/examples/hello_world/main.go
```diff
@@ -0,0 +1,178 @@
...
+
+	logger.Info("=======================================================")
+	logger.Info("Hello World Example")
+	logger.Info("=======================================================")
+	logger.Info("REST transport listening on default port 8080")
+	logger.Info("gRPC transport listening on default port 6254")
+	logger.Info("Try REST:")
+	logger.Info("  GET  http://localhost:8080/st2138-api/v1/0/value/hello_world")
+	logger.Info("Try gRPC:")
+	logger.Info("  grpcurl -plaintext -d '{\"slot\":0,\"oid\":\"hello_world\"}' localhost:6254 st2138.CatenaService/ValueRequest")
```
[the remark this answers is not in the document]
abrown-RV: bugbot is right, what a deep cut

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,678 @@
...
+				state.mu.RUnlock()
+				return status
+			}
+		}
+		state.mu.RUnlock()
+
+		state.mu.Lock()
+		defer state.mu.Unlock()
+		for _, entry := range entries {
+			if status := slotTwoApplySet(entry.Fqoid, entry.Value, state); status.Code != catena.StatusCodeOk {
```
abrown-RV: this should not fail, validate should ensure that apply cannot fail

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -0,0 +1,678 @@
...
+	// type. This slot only requires monitor-scope write access, demonstrating
+	// that each handler can enforce a different authorization policy.
+	srv.RegisterSetValueHandler(1, func(slot uint16, entries []catena.SetValueEntry, ctx catena.HandlerContext) catena.StatusResult {
+		logger.Info("SetValue", "slot", slot, "count", len(entries))
+		if !ctx.HasWriteScope(catena.ScopeMon) {
+			return catena.StatusWithCode(catena.StatusCodePermissionDenied, "monitor scope required")
+		}
+
+		for _, entry := range entries {
+			if entry.Value == nil {
```
abrown-RV, line 160: wonder if we can do something in the sdk that guarentees it will never be called with nil? so this boilerplate isn't needed everywhere? I guess since its an any, probably not...

abrown-RV (commented):
some more bugbot things too

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -199,11 +198,12 @@ func registerValueHandlers(srv catena.Server, counter *CounterState, state *Exam
...
-		defer state.mu.Unlock()
 		for _, entry := range entries {
-			if status := slotTwoApplySet(entry.Fqoid, entry.Value, state); status.Code != catena.StatusCodeOk {
-				return status
-			}
+			slotTwoApplySet(entry.Fqoid, entry.Value, state)
+		}
+		state.mu.Unlock()
+
+		for _, entry := range entries {
```
abrown-RV, line 212: Shouldn't BroadcastUpdate happen under lock? So it can clone the unmodified value

sdks/go/examples/oneOfEverything/value_handlers.go
```diff
@@ -199,11 +198,12 @@ func registerValueHandlers(srv catena.Server, counter *CounterState, state *Exam
 		state.mu.RUnlock()
 
 		state.mu.Lock()
-		defer state.mu.Unlock()
 		for _, entry := range entries {
-			if status := slotTwoApplySet(entry.Fqoid, entry.Value, state); status.Code != catena.StatusCodeOk {
-				return status
-			}
+			slotTwoApplySet(entry.Fqoid, entry.Value, state)
```
abrown-RV, line 210: Isn't there technically a window here between the RUnlock and the Lock where something could step in and change things?

sdks/go/hello_world
abrown-RV, line 1: accidental binary

sdks/go/examples/oneOfEverything/value_handlers.go
ctwarog-ross: yes, a lock should be carried through the whole handler

abrown-RV (dismissed):

ctwarog-ross:
bugbot run

abrown-RV (approved):

PR #1359, opened 2026-06-03
[GO] Reserved Update Command

### LINKED ISSUE #1339: [GO] Update commands
## 📖 Description
Implement https://github.com/rossvideo/Catena/issues/1337 in the Go SDK

---

## 🎯 Goal / Outcome

---

## ✅ Definition of Done (DoD)
- [ ] Edge cases are handled

---

---

## 🔗 Related Issues / Links

---

---

nelgd:
## 📖 Description
Introduces two SDK-reserved command OIDs, `upload_update` and `apply_update`, to the Go SDK. These commands are now auto-injected into **every** device model produced by `ToDevice`, so they are always present without any application or example code declaring them. A device may **not** define either OID itself — attempting to do so now causes `ToDevice` to fail with an explicit error naming the reserved OID.

---

## 🎯 Goal / Outcome
- `upload_update` and `apply_update` exist by default on every Catena device exposed through the Go SDK.
- The two OIDs are owned by the SDK and cannot be overridden or redefined by device authors.
- Device authors get a clear, early error if they accidentally collide with a reserved OID.

---

## ✅ Definition of Done (DoD)
- [x] `upload_update` / `apply_update` auto-injected into every device via `injectReservedCommands` in `toProtoDevice`
- [x] Reserved OIDs cannot be defined by a device — `ToDevice` returns an error naming the offending OID
- [x] Tests cover injection (with and without other commands) and rejection of user-defined reserved OIDs
- [x] `gofmt`, `go vet`, and `./pkg/...` tests pass

---

## 🔗 Related Issues / Links
- Closes issue #1339

### DIFF
--- a/sdks/go/pkg/catena/device.go
+++ b/sdks/go/pkg/catena/device.go
@@ -91,6 +91,12 @@ func toProtoDevice(m map[string]any) (*protos.Device, error) {
 		return nil, fmt.Errorf("failed to unmarshal JSON to Device proto: %w", err)
 	}
 
+	// Reserved software-update commands are auto-injected into every device so
+	// upload_update / apply_update are always present and cannot be removed.
+	if err := injectReservedCommands(device); err != nil {
+		return nil, fmt.Errorf("failed to inject reserved commands: %w", err)
+	}
+
 	return device, nil
 }
 
--- a/sdks/go/pkg/catena/server.go
+++ b/sdks/go/pkg/catena/server.go
@@ -228,15 +228,12 @@ type server struct {
 	connectionQueue        connectionQueueInterface
 	heartbeat              *Heartbeat
 	transports             []Transport
-	updateManager          *updateManager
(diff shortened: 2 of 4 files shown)


nelgd:
PR is ready to be reviewed. Both reserved commands work but there is no logic for the actual update system yet

abrown-RV:
I think as we discussed in the team sync, we can handle this with the upcoming device creation helper. Cause the reserved commands, are just reserved, they're not mandatory. This PR makes them mandatory. Just when a BL wants they they should conform to a schema.

nelgd:
> I think as we discussed in the team sync, we can handle this with the upcoming device creation helper. Cause the reserved commands, are just reserved, they're not mandatory. This PR makes them mandatory. Just when a BL wants they they should conform to a schema.

Sounds good, ill close out this pr and reference it in the issue

PR #1360, opened 2026-06-09
[GO] ConnectionProps

### LINKED ISSUE #1336: ConnectionProps
## 📖 Description
Implement connectionprops message handling in the same way as cpp

---

## 🎯 Goal / Outcome
Clicking "Detect Settings" on DB/RPM new connection should populate the dialogue with all required values to connect to the service.

---

## ✅ Definition of Done (DoD)

---

---

## 🔗 Related Issues / Links
Refer to https://github.com/rossvideo/Catena/issues/1222 for implementation notes

---

---

nelgd:
## 📖 Description

Adds a **DashBoard connection-props HTTP server** to the Go SDK. The new `catena.ConnectionProps` type runs a lightweight HTTP server that serves a single `/connect/connection-props.xml` endpoint, advertising the device so DashBoard's "Detect Frame Information" can resolve and auto-populate its connection dialog. It works in front of both REST and gRPC Catena devices.

The change wires this into the configuration system (new `DashboardOptions` with env/flag support), integrates it into the `oneOfEverything` example.

---

## 🎯 Goal / Outcome

Once merged, a Go-based Catena device can advertise its connection properties over HTTP so DashBoard can automatically discover and connect to it, matching behavior already available in the C++ SDK.

---

## ✅ Definition of Done (DoD)

- [x] `catena.ConnectionProps` server implemented synchronous bind for early error reporting and graceful shutdown,
- [x] Configuration support added (`DashboardOptions`, defaults, env/flag wiring in `load.go`)
- [x] `oneOfEverything` example integrated with start/stop lifecycle and improved startup summary
- [x] Unit tests added (`connection_props_test.go`) and existing `config` tests updated
- [x] All `pkg/` tests pass (`catena`, `config`, `logger`, `transports`); `run_coverage.sh` green
- [x] Committed `oneOfEverything` binary removed from the repo

---

- **Synchronous listener bind**: `Start()` calls `net.Listen` before launching the serving goroutine, so failures (privileged port, address in use) are returned to the caller rather than only being logged asynchronously after `Start` already reported success.
- **Transport-agnostic**: a `ConnectionProtocol` enum (`st2138-rest` / `st2138-grpc`) selects what gets advertised; the same server fronts either transport.
- **Defaults**: advertised hostname `localhost`, HTTP server on port `8080`, advertised Catena service port `6254`, refresh interval `30000ms`, endpoint `/connect/connection-props.xml`.
- **XML generation** happens once at construction; values are XML-escaped via `strings.NewReplacer`.
- During development the port moved `80` → `8080`, an early fallback REST handler approach was removed in favor of the dedicated server, and a round of Bugbot findings were addressed.

---

## 🔗 Related Issues / Links

- Closes issue #1336

### DIFF
--- a/sdks/go/examples/oneOfEverything/main.go
+++ b/sdks/go/examples/oneOfEverything/main.go
@@ -848,13 +848,10 @@ func main() {
 	}
 	connectionProps := catena.NewConnectionProps(catena.ConnectionPropsOptions{
 		Protocol:        connectionProtocol,
-		Port:            options.Dashboard.Port,
-		ServicePort:     options.Dashboard.ServicePort,
-		Hostname:        options.Dashboard.Hostname,
+		Dashboard:       options.Dashboard,
 		RefreshInterval: 30000,
 		NodeName:        "One of Everything Demo",
 		NodeID:          "one-of-everything-a4:bb:6d:6a:6f:a3",
-		TLSEnabled:      options.Dashboard.TLSEnabled,
 	})
 	connectionPropsURL := fmt.Sprintf("http://localhost:%d%s", options.Dashboard.Port, connectionProps.Endpoint())
 	if err := connectionProps.Start(); err != nil {
@@ -943,7 +940,7 @@ func main() {
 	logger.Info("")
 	logger.Info("[ REST transport ]", "status", status(options.UseRest))
 	if options.UseRest {
-		logger.Info("    web ui", "url", "http://localhost:8080/")
+		logger.Info("    web ui", "url", "http://localhost:9080/")
 	}
 
 	logger.Info("")
--- a/sdks/go/examples/oneOfEverything/main.go
+++ b/sdks/go/examples/oneOfEverything/main.go
@@ -874,16 +874,6 @@ func main() {
 		restTransport := transports.NewDefaultRestTransport()
 
 		restTransport.RegisterFallbackHandler(func(w http.ResponseWriter, r *http.Request) (catena.Value, catena.StatusResult) {
-			// Also serve the DashBoard connection-props XML on the REST port
-			if r.URL.Path == connectionProps.Endpoint() {
-				if r.Method != http.MethodGet {
-					return catena.ReplyError[catena.Value](catena.StatusCodeInvalidArgument, "method not allowed: "+r.Method)
-				}
-				w.Header().Set("Content-Type", "application/xml")
-				w.Write([]byte(connectionProps.Content()))
-				return catena.Reply(catena.Value{})
-			}
-
 			if r.URL.Path == "/assets-list" {
 				var assetList []map[string]any
 				assets.Range(func(key, value any) bool {
--- a/sdks/go/examples/oneOfEverything/main.go
+++ b/sdks/go/examples/oneOfEverything/main.go
@@ -842,17 +842,15 @@ func main() {
 	// DashBoard "Detect Frame Information" connection-props HTTP server.
 	// Advertises this device so DashBoard can resolve and populate the
 	// connection dialog. Works for both REST and gRPC devices.
-	connectionProtocol := catena.ProtocolST2138Rest
+	dashboardOpts := options.Dashboard
+	dashboardOpts.Protocol = catena.ProtocolST2138Rest
 	if options.UseGrpc {
-		connectionProtocol = catena.ProtocolST2138Grpc
+		dashboardOpts.Protocol = catena.ProtocolST2138Grpc
 	}
-	connectionProps := catena.NewConnectionProps(catena.ConnectionPropsOptions{
-		Protocol:        connectionProtocol,
-		Dashboard:       options.Dashboard,
-		RefreshInterval: 30000,
-		NodeName:        "One of Everything Demo",
-		NodeID:          "one-of-everything-a4:bb:6d:6a:6f:a3",
-	})
+	dashboardOpts.RefreshInterval = 30000
+	dashboardOpts.NodeName = "One of Everything Demo"
+	dashboardOpts.NodeID = "one-of-everything-a4:bb:6d:6a:6f:a3"
+	connectionProps := catena.NewConnectionProps(dashboardOpts)
 	connectionPropsURL := fmt.Sprintf("http://localhost:%d%s", options.Dashboard.Port, connectionProps.Endpoint())
 	if err := connectionProps.Start(); err != nil {
 		logger.Warning("Failed to start connection props server", "port", options.Dashboard.Port, "error", err)
--- a/sdks/go/examples/oneOfEverything/main.go
+++ b/sdks/go/examples/oneOfEverything/main.go
@@ -839,6 +839,30 @@ func main() {
 		os.Exit(1)
 	}
 
+	// DashBoard "Detect Frame Information" connection-props HTTP server.
+	// Advertises this device so DashBoard can resolve and populate the
+	// connection dialog. Works for both REST and gRPC devices.
+	connectionProtocol := catena.ProtocolST2138Rest
+	if options.UseGrpc {
+		connectionProtocol = catena.ProtocolST2138Grpc
+	}
+	connectionProps := catena.NewConnectionProps(catena.ConnectionPropsOptions{
+		Protocol:        connectionProtocol,
+		Port:            options.Dashboard.Port,
+		ServicePort:     options.Dashboard.ServicePort,
+		Hostname:        options.Dashboard.Hostname,
+		RefreshInterval: 30000,
+		NodeName:        "One of Everything Demo",
+		NodeID:          "one-of-everything-a4:bb:6d:6a:6f:a3",
+		TLSEnabled:      options.Dashboard.TLSEnabled,
+	})
+	if err := connectionProps.Start(); err != nil {
+		logger.Warning("Failed to start connection props server", "port", options.Dashboard.Port, "error", err)
+	} else {
+		logger.Info("Connection props available at:",
+			"url", fmt.Sprintf("http://localhost:%d%s", options.Dashboard.Port, connectionProps.Endpoint()))
+	}
+
 	logger.Info("=======================================================")
 	logger.Info("One of Everything gRPC Example")
 	logger.Info("=======================================================")
@@ -866,6 +890,16 @@ func main() {
 		restTransport := transports.NewDefaultRestTransport()
 
 		restTransport.RegisterFallbackHandler(func(w http.ResponseWriter, r *http.Request) (catena.Value, catena.StatusResult) {
+			// Also serve the DashBoard connection-props XML on the REST port
+			if r.URL.Path == connectionProps.Endpoint() {
+				if r.Method != http.MethodGet {
+					return catena.ReplyError[catena.Value](catena.StatusCodeInvalidArgument, "method not allowed: "+r.Method)
+				}
+				w.Header().Set("Content-Type", "application/xml")
+				w.Write([]byte(connectionProps.Content()))
+				return catena.Reply(catena.Value{})
+			}
+
 			if r.URL.Path == "/assets-list" {
 				var assetList []map[string]any
 				assets.Range(func(key, value any) bool {
@@ -926,6 +960,9 @@ func main() {
 
 	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), time.Second)
 	defer shutdownCancel()
+	if err := connectionProps.Stop(shutdownCtx); err != nil {
+		logger.Warning("Error stopping connection props server", "error", err)
+	}
 	srv.Shutdown(shutdownCtx)
 	logger.Info("Server shutdown complete")
 }
--- a/sdks/go/examples/oneOfEverything/main.go
+++ b/sdks/go/examples/oneOfEverything/main.go
@@ -856,37 +856,21 @@ func main() {
 		NodeID:          "one-of-everything-a4:bb:6d:6a:6f:a3",
 		TLSEnabled:      options.Dashboard.TLSEnabled,
 	})
+	connectionPropsURL := fmt.Sprintf("http://localhost:%d%s", options.Dashboard.Port, connectionProps.Endpoint())
 	if err := connectionProps.Start(); err != nil {
 		logger.Warning("Failed to start connection props server", "port", options.Dashboard.Port, "error", err)
-	} else {
-		logger.Info("Connection props available at:",
-			"url", fmt.Sprintf("http://localhost:%d%s", options.Dashboard.Port, connectionProps.Endpoint()))
+		connectionPropsURL = ""
 	}
 
-	logger.Info("=======================================================")
-	logger.Info("One of Everything gRPC Example")
-	logger.Info("=======================================================")
-
-	// register the transports we want to serve on.
+	// Register the enabled transports.
 	if options.UseGrpc {
-		logger.Info("gRPC transport starting")
-		logger.Info("")
-		// Register gRPC transport if enabled in config
 		if err := srv.RegisterTransport(transports.NewDefaultGrpcTransport()); err != nil {
 			logger.Error("Failed to register gRPC transport", "error", err)
 			os.Exit(1)
 		}
-		logger.Info("Use grpcurl or a gRPC client to interact with the server:")
-		logger.Info("  grpcurl -plaintext localhost:6254 list")
-	} else {
-		logger.Info("gRPC transport disabled by config")
 	}
-	logger.Info("")
-	logger.Info("=======================================================")
 
 	if options.UseRest {
-		logger.Info("REST transport starting")
-		logger.Info("")
 		restTransport := transports.NewDefaultRestTransport()
 
 		restTransport.RegisterFallbackHandler(func(w http.ResponseWriter, r *http.Request) (catena.Value, catena.StatusResult) {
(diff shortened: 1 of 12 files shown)


sdks/go/examples/oneOfEverything/main.go
```diff
@@ -837,30 +837,38 @@ func main() {
...
+		TLSEnabled:      options.Dashboard.TLSEnabled,
+	})
+	connectionPropsURL := fmt.Sprintf("http://localhost:%d%s", options.Dashboard.Port, connectionProps.Endpoint())
+	if err := connectionProps.Start(); err != nil {
+		logger.Warning("Failed to start connection props server", "port", options.Dashboard.Port, "error", err)
+		connectionPropsURL = ""
+	}
 
-	// register the transports we want to serve on.
+	// Register the enabled transports.
```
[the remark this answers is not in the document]
nelgd: Matched connectionprop with the c++ version and it uses 8080. If there is a different port I should use let me know

nelgd:
You need to have authz off for this to work

sdks/go/pkg/config/config.go
```diff
@@ -81,6 +82,20 @@ type JwtValidationOptions struct {
 	Http *http.Client
 }
 
+// DashboardOptions configures the DashBoard connection-props HTTP server that
+// advertises this device for the "Detect Frame Information" workflow. It is
+// transport-agnostic and can front either a REST or gRPC Catena device.
+type DashboardOptions struct {
```
abrown-RV, line 115: prefix the env and cli with dashboard or db or something

sdks/go/examples/oneOfEverything/main.go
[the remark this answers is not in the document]
abrown-RV: ya in cpp rest defaults to 6254 like grpc cause there's no multiple transports thing, it just uses a shared port var that is grpc or rest. On the cpp side we default to 8080 cause that nicer in dev, since 80 is often squatted on by something else, and all of our use cases are in containers where the port doesn't actually matter, you can always remap at the container level.

I think we should pick a new default discovery port. Most of the time it will be remapped anyways by k8s ingress, and when not, well they can still set it to whatever they want.

How about `9080`? @ctwarog-ross?

sdks/go/examples/oneOfEverything/main.go
```diff
@@ -839,30 +839,38 @@ func main() {
...
-	logger.Info("One of Everything gRPC Example")
-	logger.Info("=======================================================")
+	// DashBoard "Detect Frame Information" connection-props HTTP server.
+	// Advertises this device so DashBoard can resolve and populate the
+	// connection dialog. Works for both REST and gRPC devices.
+	connectionProtocol := catena.ProtocolST2138Rest
+	if options.UseGrpc {
+		connectionProtocol = catena.ProtocolST2138Grpc
+	}
+	connectionProps := catena.NewConnectionProps(catena.ConnectionPropsOptions{
```
abrown-RV: ideally we just have the 1 options struct for connectionprops, I think most of the fields on ConnectionPropsOptions could actually come from InitOptions

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,319 @@
...
+		scheme = "https"
+		connectionType = "SSL"
+	}
+
+	var b strings.Builder
+	b.WriteString("<properties version=\"1.0\">\n")
+	b.WriteString("    <comment>DashBoard Device Connection Settings</comment>\n")
+	writeEntry(&b, "base-url", fmt.Sprintf("%s://%s/", scheme, c.opts.Hostname))
+	writeEntry(&b, "serviceUrl", c.opts.ServiceName)
+	writeEntry(&b, "equipmentType", "catena")
```
abrown-RV: shouldn't this be connection protocol?

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,319 @@
...
+// ConnectionPropsOptions configures a ConnectionProps server
+type ConnectionPropsOptions struct {
+	// Protocol is the transport being advertised (used for logging).
+	Protocol ConnectionProtocol
+	// Port is the port the connection-props HTTP server listens on (default 8080).
+	Port int
+	// ServicePort is the Catena service port advertised to DashBoard (default 6254).
+	ServicePort int
+	// Hostname is the address advertised to DashBoard (default "localhost").
+	Hostname string
```
abrown-RV: `ServiceHostname` for consistancy

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,319 @@
...
+	// NodeName is the optional human-readable node name.
+	NodeName string
+	// NodeID is the optional unique node identifier.
+	NodeID string
+	// ServiceName is the advertised service URL (default "service:catena-device").
+	ServiceName string
+	// Endpoint is the path served (default "/connect/connection-props.xml").
+	Endpoint string
+	// TLSEnabled controls whether the advertised connection uses TLS/SSL.
+	TLSEnabled bool
```
abrown-RV: `ServiceTLS` for consistancy

abrown-RV (changes requested):

ctwarog-ross:
it won't let me reply to Andrew's comment https://github.com/rossvideo/Catena/pull/1360#discussion_r3404310971
I think we change the default rest transport to 9080 and leave connectionprops on 8080.
I specifically told Nelson to do that cuz 8080 is one of the ports Dashboard pings by default, if you use another one you have to specify it in the device URL which isnt intuitive

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,319 @@
...
+)
+
+// ConnectionProtocol identifies the Catena transport advertised to DashBoard.
+type ConnectionProtocol int
+
+const (
+	// ProtocolST2138Rest advertises a ST 2138 REST device.
+	ProtocolST2138Rest ConnectionProtocol = iota
+	// ProtocolST2138Grpc advertises a ST 2138 gRPC device.
+	ProtocolST2138Grpc
```
ctwarog-ross: change this to just 2 string consts and add another ProtocolST2138Catena = "catena" to support legacy devices

sdks/go/examples/oneOfEverything/main.go
nelgd: ConnectionPropsOptions now uses config.DashboardOptions for the shared connection settings

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,315 @@
...
+	defaultConnectionPropsEndpoint    = "/connect/connection-props.xml"
+	defaultConnectionPropsService     = "service:catena-device"
+	defaultConnectionPropsRefreshMs   = 30000
+	defaultConnectionPropsHostname    = "localhost"
+	defaultConnectionPropsPort        = 8080
+	defaultConnectionPropsServicePort = 6254
+)
+
+// ConnectionPropsOptions configures a ConnectionProps server
+type ConnectionPropsOptions struct {
```
abrown-RV: merge this with `DashboardOptions` in config. The vision is all options structs just live in `config`. Not all fields need to have clv/env but if its easy to map, may as well.

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,315 @@
...
+	// ProtocolST2138Rest advertises a ST 2138 REST device.
+	ProtocolST2138Rest ConnectionProtocol = "st2138-rest"
+	// ProtocolST2138Grpc advertises a ST 2138 gRPC device.
+	ProtocolST2138Grpc ConnectionProtocol = "st2138-grpc"
+	// ProtocolST2138Catena advertises a legacy "catena" device.
+	ProtocolST2138Catena ConnectionProtocol = "catena"
+)
+
+// String returns the canonical protocol identifier used in logs.
+func (p ConnectionProtocol) String() string {
```
abrown-RV: is this needed? the basic type is already a string?

sdks/go/pkg/config/load.go
```diff
@@ -98,6 +98,11 @@ func InitOptionsPrefix(appName, prefix string, args []string) (RuntimeOptions, e
 		extractString("JWT_ISSUER", "jwt-issuer", "Expected JWT issuer for validating incoming requests", &opts.Server.JwtOptions.Issuer).
 		extractString("JWT_AUDIENCE", "jwt-audience", "Expected JWT audience for validating incoming requests", &opts.Server.JwtOptions.Audience).
 		extractBool("JWT_VALIDATE_SIGNATURE", "jwt-validate-signature", "Whether to validate the JWT signature or just the claims", &opts.Server.JwtOptions.ValidateSignature).
+		// DashBoard connection-props options
+		extractString("DASHBOARD_HOSTNAME", "dashboard-hostname", "Advertised hostname/address for DashBoard connection props", &opts.Dashboard.ServiceHostname).
+		extractInt("DASHBOARD_PORT", "dashboard-port", "Port for the DashBoard connection-props HTTP server", &opts.Dashboard.Port).
+		extractInt("DASHBOARD_SERVICE_PORT", "dashboard-service-port", "Catena service port advertised to DashBoard", &opts.Dashboard.ServicePort).
+		extractBool("DASHBOARD_TLS_ENABLED", "dashboard-tls-enabled", "Whether the advertised DashBoard connection uses TLS", &opts.Dashboard.ServiceTLS).
```
abrown-RV: should have service in the cli/env names as well

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,315 @@
...
+func (c *ConnectionProps) Start() error {
+	c.mu.Lock()
+	if c.running {
+		c.mu.Unlock()
+		logger.Warning("Connection props server already running", "port", c.opts.Dashboard.Port)
+		return fmt.Errorf("connection props server already running")
+	}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", c.handle)
```
abrown-RV: do we even need a mux with a single handler? could just register the handler directly on the server.
 orrrrr since we do know endpoint here, could register 2 handlers, health and endpoint

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,315 @@
...
+
+	if !running || srv == nil {
+		return nil
+	}
+
+	logger.Info("Stopping connection props server", "port", c.opts.Dashboard.Port)
+	err := srv.Shutdown(ctx)
+	if err != nil {
+		// Best-effort hard close so we never leak the listener.
+		_ = srv.Close()
```
abrown-RV:
```suggestion
		logger.Warning("HTTP server shutdown timed out, forcing close")
		closeErr := srv.Close()
		if closeErr != nil {
			logger.Error("failed to force close HTTP server", "error", closeErr)
		}
```
give a little more logging here

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", c.handle)
+
+	addr := fmt.Sprintf(":%d", c.opts.Dashboard.Port)
+	// Bind synchronously so that errors (privileged port, address already in
+	// use, etc.) are returned to the caller instead of only being logged
+	// asynchronously after Start has already reported success.
+	listener, err := net.Listen("tcp", addr)
```
abrown-RV, line 116: this is a better startup routine for http server, should replicate over in RestTransport, make a ticket

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,315 @@
...
+		w.Header().Set("Content-Type", "text/plain")
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte("OK"))
+	default:
+		http.Error(w, "Not Found", http.StatusNotFound)
+	}
+}
+
+// generateXML renders the DashBoard connection-props XML payload.
+func (c *ConnectionProps) generateXML() string {
```
abrown-RV, line 208: may as well use the built in go xml marshalling, bonus points if you can use an anonymous struct for less package clutter.

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,315 @@
...
+	scheme := "http"
+	connectionType := "TCP"
+	if c.opts.Dashboard.ServiceTLS {
+		scheme = "https"
```
abrown-RV: don't think dashboard recognizes `scheme https`

abrown-RV (changes requested):
don't forget to re-export the `DashboardOptions` struct in types.go

sdks/go/pkg/catena/connection_props.go
nelgd: Made an issue for this

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,289 @@
...
+	"net/http"
+	"strings"
+	"sync"
+
+	"github.com/rossvideo/catena/sdks/go/pkg/config"
+	"github.com/rossvideo/catena/sdks/go/pkg/logger"
+)
+
+// Default values
+const (
```
abrown-RV: these defaults should just exist in DefaultDashboardOptions

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,289 @@
...
+	opts config.DashboardOptions
+
+	mu      sync.Mutex
+	server  *http.Server
+	content string
+	running bool
+}
+
+// NewConnectionProps builds a ConnectionProps server, applying defaults for any
+// unset option. The XML payload is generated once at construction time.
```
abrown-RV: I've been running with a different philophacy: options structs are assumed to be well filled, so instead of defaulting unset fields, it should error if anything truely unrecoverable.... which I'm not sure anything is unrecoverable... except maybe for protocol, if that's not a real value then something is deffinately wrong. Main driving force for this is defaults living in multiple places are brittle.

Can add something like this to the comments:
```
// DashboardOptions must be initialized before use.
// Prefer config.InitOptions or config.DefaultDashboardOptions().
type DashboardOptions struct { ... }
```

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,289 @@
...
+import (
+	"context"
+	"encoding/xml"
+	"fmt"
+	"net"
+	"net/http"
+	"strings"
+	"sync"
+
+	"github.com/rossvideo/catena/sdks/go/pkg/config"
```
abrown-RV: all the types are re-exported, shouldn't need this import anymore

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,289 @@
...
+		{Key: "connectionType", Value: connectionType},
+	}
+	if c.opts.NodeID != "" {
+		entries = append(entries, entry{Key: "node-id", Value: c.opts.NodeID})
+	}
+	if c.opts.NodeName != "" {
+		entries = append(entries, entry{Key: "node-name", Value: c.opts.NodeName})
+	}
+	entries = append(entries,
+		entry{Key: "index-url", Value: "connect/connection-props.xml"},
```
abrown-RV: should be endpoint I think

sdks/go/pkg/catena/connection_props.go
```diff
@@ -0,0 +1,289 @@
...
+func NewConnectionProps(opts config.DashboardOptions) *ConnectionProps {
+	if opts.Protocol == "" {
+		opts.Protocol = config.ProtocolST2138Rest
+	}
+	if opts.Port == 0 {
+		opts.Port = defaultConnectionPropsPort
+	}
+	if opts.ServicePort == 0 {
+		opts.ServicePort = defaultConnectionPropsServicePort
+	}
```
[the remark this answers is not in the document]
abrown-RV: Suggestion: if we make ServicePort default 0, then document that if ServicePort is given to NewConnectionProps as 0 then it will set itself based on the selected protocoltype

sdks/go/pkg/config/config.go
```diff
@@ -93,6 +94,45 @@ func (o JwtValidationOptions) ResolvedAllowedAlgs() []string {
...
+	ServiceHostname string `env:"DASHBOARD_SERVICE_HOSTNAME" flag:"dashboard-service-hostname"`
+	// Port is the port the connection-props HTTP server listens on (default 8080).
+	Port int `env:"DASHBOARD_PORT" flag:"dashboard-port"`
+	// ServicePort is the Catena service port advertised to DashBoard (default 6254).
+	ServicePort int `env:"DASHBOARD_SERVICE_PORT" flag:"dashboard-service-port"`
+	// ServiceTLS controls whether the advertised connection uses TLS/SSL (default false).
+	ServiceTLS bool `env:"DASHBOARD_SERVICE_TLS_ENABLED" flag:"dashboard-service-tls-enabled"`
+	// Protocol is the transport being advertised (used for logging and the
+	// advertised equipmentType).
+	Protocol ConnectionProtocol
```
abrown-RV: may as well add these 2 to InitOptions as well?  Protocol is a little tough cause it needs validation

sdks/go/pkg/config/config.go
```diff
@@ -112,10 +152,22 @@ type LoggerOptions struct {
...
+		Logger:    DefaultLoggerOptions(),
+		Dashboard: DefaultDashboardOptions(),
+	}
+}
+
+// DefaultDashboardOptions returns sensible defaults for the DashBoard
+// connection-props server.
+func DefaultDashboardOptions() DashboardOptions {
+	return DashboardOptions{
+		ServiceHostname: "localhost",
```
abrown-RV, line 170: should match the defaults from the docs

abrown-RV (commented):
Very nice work on the last round, bonus points for the xml marshalling! :D

sdks/go/pkg/catena/connection_props.go
nelgd: C++ uses index-url. I can try and see if endpoint will work too

sdks/go/pkg/config/config.go
nelgd: Is the default 0.0.0.0? I had that originally but Dashboard wouldnt detect the device

abrown-RV: localhost is a fine default

0.0.0.0 is used for port listeners to tell it which interface to bind to. 0.0.0.0 means every interface. You can make a port only listen for local-only connections with 127.0.0.1, and if a machine has multiple networks, you can put in the machines ip, on a specific network to only open the port on that network.

sdks/go/pkg/catena/connection_props.go
abrown-RV: I mean the value, can be c.opts.Endpoint

abrown-RV: or c.Endpoint()

sdks/go/pkg/config/config.go
```diff
@@ -112,10 +155,26 @@ type LoggerOptions struct {
...
+		Port:            8080,
+		ServicePort:     6254,
+		ServiceTLS:      false,
+		Protocol:        ProtocolST2138Rest,
```
abrown-RV: default protocol grpc to match the default service port

sdks/go/pkg/config/config.go
```diff
@@ -93,6 +94,48 @@ func (o JwtValidationOptions) ResolvedAllowedAlgs() []string {
...
+)
+
+// DashboardOptions configures the DashBoard connection-props HTTP server that
+// advertises this device for the "Detect Frame Information" workflow. It is
+// transport-agnostic and can front either a REST or gRPC Catena device.
+//
+// DashboardOptions must be initialized before use.
+// Prefer config.InitOptions or config.DefaultDashboardOptions().
+type DashboardOptions struct {
+	// ServiceHostname is the address advertised to DashBoard (default "localhost").
```
abrown-RV, line 116: Needs a LogValuer

abrown-RV (changes requested):
couple little tweaks

sdks/go/pkg/config/config.go
nelgd: so just leave it as is then?

abrown-RV: yup localhost is fine, basically will always have to be changed... but also we have no way of knowing the real value in prod. And its convenient in dev

abrown-RV (approved):

PR #1361, opened 2026-06-10
[GO] better config log

abrown-RV:
## 📖 Description

Quick little patch to allow logging of any of the options structs

---

## ✅ Definition of Done (DoD)
- [x] Each options struct gets its own `LogValue` func
- [x] Tests are added or updated

---

## 🛠️ Implementation Notes
split the LogValue func into funcs for each options level and chain them together

---

## 🔗 Related Issues / Links

---
## 📝 Additional Context

---

## 🚧 Risks / Open Questions

---
## ✨ Updates

### DIFF
--- a/sdks/go/pkg/config/config.go
+++ b/sdks/go/pkg/config/config.go
@@ -43,6 +43,8 @@ import (
 	"net/http"
 	"strings"
 	"time"
+
+	"github.com/golang-jwt/jwt/v5"
 )
 
 type RuntimeOptions struct {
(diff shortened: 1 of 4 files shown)


sdks/go/pkg/config/config.go
```diff
@@ -81,6 +81,16 @@ type JwtValidationOptions struct {
 	Http *http.Client
 }
 
+const defaultJwtSigningAlg = "ES256"
```
ctwarog-ross: Why are we hardcoding this when we could use jwt.SigningMethodES256.Alg()

ctwarog-ross (approved):

PR #1365, opened 2026-06-15
[GO] MultiSetValue Endpoint

### LINKED ISSUE #1363: [GO] GRPC MultiSetValue
## 📖 Description
implement MultisetValue endpoint hit.
It should attach to the mutisetvalue GRPC endpoint hook in grpc_stransport.go.

---

## 🎯 Goal / Outcome
Multisetvalue example in one-of-everything
Multisetvalue endpoint handling
unit tests

---

## ✅ Definition of Done (DoD)
A list of TODOs that need to be done for the issue to be done:
- [ ] Edge cases are handled

---

---

## 🔗 Related Issues / Links

---

---

### LINKED ISSUE #1364: [GO] REST MultiSetValue
## 📖 Description
Implement REST MultiSetValue PUT endpoint
Map to url/fqoid/values.

---

## 🎯 Goal / Outcome
Multisetvalue example in one-of-everything
Multisetvalue endpoint handling
unit tests

---

## ✅ Definition of Done (DoD)
A list of TODOs that need to be done for the issue to be done:
- [ ] Edge cases are handled

---

---

## 🔗 Related Issues / Links

---

---

nelgd:
## 📖 Description

Implements the `MultiSetValue` endpoint for the Go SDK, exposed over REST as `PUT /st2138-api/v1/{slot}/values`. The change introduces a new dedicated runtime handler for applying multiple parameter values, and wires both the REST and gRPC transports to use it. When a dedicated handler is registered the operation is atomic. Otherwise it falls back to the existing single-value handler applied per value.

---

## 🎯 Goal / Outcome

Clients can set multiple parameters in a single request per the ST 2138 spec, with a path for true all-or-nothing semantics. 

---

## ✅ Definition of Done (DoD)

- [x] `PUT /st2138-api/v1/{slot}/values` implemented in the REST transport, returning `204 No Content` on success
- [x] New runtime handler surface added: `SetValueEntry`, `MultiSetValueHandler`, `RegisterMultiSetValueHandler`, `InvokeMultiSetValueHandler`, `EndpointMultiSetValue`
- [x] gRPC `MultiSetValue` rewired to the new runtime handler (removed per-value loop + TODO)
- [x] Atomic path via a dedicated handler; non-atomic fallback over the single `SetValueHandler` for backward compatibility
- [x] Edge cases handled: invalid content type, malformed body, value conversion error, wrong method, handler errors, no handler registered
- [x] Tests added across REST, gRPC, and runtime layers (success, fallback, fallback-stops-on-error, NotFound, write-access denial)
- [x] Documentation updated (REST.md, README.md, GRPC.md)

---

- `catena/server.go`: adds `EndpointMultiSetValue`, the `SetValueEntry` type, `MultiSetValueHandler`, a `multiSetValueHandlers` map, plus `RegisterMultiSetValueHandler` / `InvokeMultiSetValueHandler`. `InvokeMultiSetValueHandler` performs the same write + access-control checks as single SetValue, then: uses the dedicated handler if registered (atomic); else loops the single handler per entry and stops on first error (non-atomic); else returns `NotFound`.
- `transports/json_utils.go`: adds `ReadMultiSetValuesRequestJSON`, which validates `application/json`, parses `{"values":[{"fqoid","value"}]}` (note the body uses `fqoid`, not the proto `oid`), unmarshals each value via protojson, and converts via `catena.FromProto`.
- `transports/rest_transport.go`: adds the `values` route case and `handleValuesEndpoint` (PUT only, 204 on success, errors mapped via `ToHTTPStatus`).
- `transports/grpc_transport.go`: `MultiSetValue` now builds `[]catena.SetValueEntry` and calls the shared handler.
- The test stub mirrors the runtime's fallback so pre-existing gRPC MultiSetValue tests remain valid.

---

## 🔗 Related Issues / Links

Closes issue #1363 and #1364 

---

### DIFF
--- a/sdks/go/pkg/catena/server.go
+++ b/sdks/go/pkg/catena/server.go
@@ -63,6 +63,7 @@ const (
 	EndpointGetDevice
 	EndpointGetValue
 	EndpointSetValue
+	EndpointMultiSetValue
 	EndpointGetAsset
 	EndpointExecuteCommand
 	EndpointParamInfo
@@ -122,6 +123,15 @@ func (ctx HandlerContext) HasAnyReadScope() bool {
 type DeviceHandler func(slot uint16, ctx HandlerContext) (Device, StatusResult)
 type GetValueHandler func(slot uint16, fqoid string, ctx HandlerContext) (Value, StatusResult)
 type SetValueHandler func(value any, slot uint16, fqoid string, ctx HandlerContext) StatusResult
+
+// SetValueEntry is a single fqoid/value pair within a MultiSetValue request.
+type SetValueEntry struct {
+	Fqoid string
+	Value any
+}
+
+// MultiSetValueHandler applies multiple parameter values atomically (all-or-nothing).
+type MultiSetValueHandler func(values []SetValueEntry, slot uint16, ctx HandlerContext) StatusResult
 type GetAssetHandler func(slot uint16, fqoid string, ctx HandlerContext) (Asset, StatusResult)
 type ExecuteCommandHandler func(slot uint16, commandFqoid string, payload any, ctx HandlerContext) (CommandResult, StatusResult)
 type ParamInfoHandler func(slot uint16, oidPrefix string, recursive bool, ctx HandlerContext) ([]ParamInfo, StatusResult)
@@ -177,6 +187,7 @@ type Server interface {
 	RegisterGetDeviceHandler(slot uint16, handler DeviceHandler)
 	RegisterGetValueHandler(slot uint16, handler GetValueHandler)
 	RegisterSetValueHandler(slot uint16, handler SetValueHandler)
+	RegisterMultiSetValueHandler(slot uint16, handler MultiSetValueHandler)
 	RegisterGetAssetHandler(slot uint16, handler GetAssetHandler)
 	RegisterExecuteCommandHandler(slot uint16, handler ExecuteCommandHandler)
 	RegisterParamInfoHandler(slot uint16, handler ParamInfoHandler)
@@ -198,6 +209,7 @@ type ServerRuntime interface {
 	InvokeGetDeviceHandler(slot uint16, transportContext TransportContext) (Device, StatusResult)
 	InvokeGetValueHandler(slot uint16, fqoid string, transportContext TransportContext) (Value, StatusResult)
 	InvokeSetValueHandler(value any, slot uint16, fqoid string, transportContext TransportContext) StatusResult
+	InvokeMultiSetValueHandler(values []SetValueEntry, slot uint16, transportContext TransportContext) StatusResult
 	InvokeGetAssetHandler(slot uint16, fqoid string, transportContext TransportContext) (Asset, StatusResult)
 	InvokeExecuteCommandHandler(slot uint16, commandFqoid string, payload any, transportContext TransportContext) (CommandResult, StatusResult)
 	InvokeParamInfoHandler(slot uint16, oidPrefix string, recursive bool, transportContext TransportContext) ([]ParamInfo, StatusResult)
@@ -223,6 +235,7 @@ type server struct {
 	getDeviceHandlers      map[uint16]DeviceHandler
 	getValueHandlers       map[uint16]GetValueHandler
 	setValueHandlers       map[uint16]SetValueHandler
+	multiSetValueHandlers  map[uint16]MultiSetValueHandler
 	getAssetHandlers       map[uint16]GetAssetHandler
 	executeCommandHandlers map[uint16]ExecuteCommandHandler
 	paramInfoHandlers      map[uint16]ParamInfoHandler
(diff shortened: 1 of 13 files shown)


sdks/go/pkg/catena/server.go
```diff
@@ -660,6 +685,49 @@ func (s *server) InvokeSetValueHandler(value any, slot uint16, fqoid string, tra
...
+// InvokeMultiSetValueHandler applies multiple parameter values for a slot.
+// If a dedicated MultiSetValueHandler is registered for the slot, it is used to
+// guarantee all-or-nothing (atomic) semantics. Otherwise, it falls back to
+// looping over the single SetValueHandler per entry, returning on the first
+// error, however it is not atomic; partial application can occur if a later
+// entry fails.
+func (s *server) InvokeMultiSetValueHandler(values []SetValueEntry, slot uint16, transportContext TransportContext) StatusResult {
+	handlerContext, res := s.resolveHandlerContext(transportContext)
+	if res.Code != StatusCodeOk {
+		return res
```
abrown-RV: add multisetvalue to the aggregated unit tests that goes through each endpoint hitting these extra checks

sdks/go/pkg/transports/json_utils.go
```diff
@@ -202,6 +202,58 @@ func ReadRequestJSON(r *http.Request) (*protos.Value, catena.StatusResult) {
...
 }
 
+// ReadMultiSetValuesRequestJSON reads and unmarshals a SetValues request body of
+// the form {"values":[{"fqoid":"...","value":{<Value oneof>}}, ...]} into a slice of catena.SetValueEntry.
+func ReadMultiSetValuesRequestJSON(r *http.Request) ([]catena.SetValueEntry, catena.StatusResult) {
+	defer r.Body.Close()
+
+	contentType := r.Header.Get("Content-Type")
+	if contentType == "" {
+		return nil, catena.StatusResult{Code: catena.StatusCodeInvalidArgument, Error: "missing Content-Type header"}
```
abrown-RV, line 215: hit these err blocks

sdks/go/pkg/transports/json_utils.go
```diff
@@ -202,6 +202,58 @@ func ReadRequestJSON(r *http.Request) (*protos.Value, catena.StatusResult) {
...
+	if mediaType != "application/json" {
+		return nil, catena.StatusResult{Code: catena.StatusCodeInvalidArgument, Error: fmt.Sprintf("unsupported content type: %s, expected application/json", mediaType)}
+	}
+
+	data, err := io.ReadAll(r.Body)
+	if err != nil {
+		return nil, catena.StatusResult{Code: catena.StatusCodeInvalidArgument, Error: fmt.Sprintf("failed to read request body: %v", err)}
+	}
+
+	var body struct {
```
abrown-RV: Should parse the whole body with `protos.MultiSetValuePayload` probably faster than parsing the values individually after first parsing into this struct

sdks/go/pkg/transports/stub_server_test.go
```diff
@@ -118,6 +119,25 @@ func (s *stubServerRuntime) InvokeSetValueHandler(value any, slot uint16, fqoid
 	return catena.StatusResult{Code: catena.StatusCodeInternal}
 }
 
+func (s *stubServerRuntime) InvokeMultiSetValueHandler(values []catena.SetValueEntry, slot uint16, ctx catena.TransportContext) catena.StatusResult {
+	if s.multiSetValueFn != nil {
+		return s.multiSetValueFn(values, slot, ctx)
+	}
+	// Mirror the real server's non-atomic fallback: apply each entry via the
```
abrown-RV: this is not how the real server works, the real server *is* atomic. This should just work the same as the rest of the panicy fns.

abrown-RV (changes requested):
Awesome, glad to see this going in

abrown-RV (changes requested):
Ok, and apologies for shifting the goalposts on this. We've decided to simplify things by only having the `SetValueHandler` change to taking the array of entries, and then single `SetValue` endpoint calls are wrapped into an array with the one entry. And then MutliSetValue endpoints calls SetValueHandler with their full array. Does that make sense?

nelgd:
> Ok, and apologies for shifting the goalposts on this. We've decided to simplify things by only having the `SetValueHandler` change to taking the array of entries, and then single `SetValue` endpoint calls are wrapped into an array with the one entry. And then MutliSetValue endpoints calls SetValueHandler with their full array. Does that make sense?

Yeah makes sense I think. So multi and single set value will both technically use the same endpoint but multiset will loop through its array of values to change?

abrown-RV:
> > Ok, and apologies for shifting the goalposts on this. We've decided to simplify things by only having the `SetValueHandler` change to taking the array of entries, and then single `SetValue` endpoint calls are wrapped into an array with the one entry. And then MutliSetValue endpoints calls SetValueHandler with their full array. Does that make sense?
> 
> Yeah makes sense I think. So multi and single set value will both technically use the same endpoint but multiset will loop through its array of values to change?

They'll use the same handler, that is inherently the mutli handler, single set value can just very easily be converted into a multiple by wrapping in the array

abrown-RV (approved):

PR #1440, opened 2026-07-08
[GO] Languages Endpoint

### LINKED ISSUE #1367: [GO] Languages endpoint
## 📖 Description
languages endpoint
```
get:
  summary: Gets a list of language codes that are supported by the device model
  operationId: Languages

  # at least one scope with read access is required.
  security:
    - oauth2:
      - st2138:op
    - oauth2:
      - st2138:cfg
    - oauth2:
      - st2138:mon
    - oauth2:
      - st2138:adm

  parameters:
    - name: slot
      in: path
      required: true
      description: The slot of the device model to get language codes from
      schema:
        $ref: "../../schemata/device.json#/$defs/slot"
      example: 1  

  tags:
    - Languages

  responses:
    "200":
      description: A list of the language codes that are supported by the device model
      content:
        application/json:
          schema:
            type: array
            items:
              type: string
          example: ["en", "fr", "es", "de"]
    
    "500":
      $ref: '../openapi.yaml#/components/responses/InternalServerError'

    "400":
      $ref: '../openapi.yaml#/components/responses/BadRequest'
          
    "401":
      $ref: '../openapi.yaml#/components/responses/Unauthorized'

    "403":
      $ref: '../openapi.yaml#/components/responses/Forbidden'

    "404":
      $ref: '../openapi.yaml#/components/responses/NotFound'

    "503":
      $ref: '../openapi.yaml#/components/responses/ServiceUnavailable'
```

---

## 🎯 Goal / Outcome
url/slot/languages endpoint to serve a list of languages for that slot
unit tests

---

## ✅ Definition of Done (DoD)
A list of TODOs that need to be done for the issue to be done:
- [ ] Edge cases are handled

---

---

## 🔗 Related Issues / Links

---

---

### LINKED ISSUE #1377: [GO[ GRPC ListLanguages
## 📖 Description
Implement ListLanguage endpoint
```
ListLanguages(Slot) returns (LanguageList);
```

---

## 🎯 Goal / Outcome
A client can query the GRPC server to get a Language List like so (querying one-of-everything in cpp on port 6254):
```
grpcurl -plaintext -import-path ~/Catena/smpte/interface/proto -proto service.proto   -d '{"slot":0}' localhost:6254 st2138.CatenaService/ListLanguages
```
```
{
    "languages": [
        "de",
        "fr",
        "en",
        "es"
    ]
}
```

---

## ✅ Definition of Done (DoD)

---

---

## 🔗 Related Issues / Links

---

---

nelgd:
## 📖 Description

Adds the **Languages** endpoint to the Go SDK, exposing the language codes a device model supports at a given slot. The work follows the SDK's established per-endpoint pattern (handler type → `Server` registration → `ServerRuntime` invoke with access checks → transport route → test stub) and implements it across both transports:

- **REST:** `GET /st2138-api/v1/{slot}/languages` returns a JSON array of language codes (e.g. `["en","fr","es","de"]`).
- **gRPC:** the previously `Unimplemented` `ListLanguages` RPC now returns a populated `protos.LanguageList`.

A shared `ListLanguagesHandler` (`func(slot uint16, ctx HandlerContext) ([]string, StatusResult)`) backs both transports, and the `oneOfEverything` example registers a sample handler.

---

## 🎯 Goal / Outcome

Clients can query the set of supported language codes for a device slot over REST or gRPC, matching the ST 2138 Languages endpoint specification, instead of receiving an `Unimplemented` error.

---

## ✅ Definition of Done (DoD)

- [x] `ListLanguagesHandler` type, `EndpointListLanguages`, and `Register/InvokeListLanguagesHandler` added to `server.go` (with read-scope access checks)
- [x] REST route `GET /{slot}/languages` implemented (`handleLanguagesEndpoint`), serializing results as a JSON array
- [x] gRPC `ListLanguages` implemented against the shared runtime handler
- [x] Sample handler registered in the `oneOfEverything` example
- [x] Edge cases handled: unregistered slot → `404 NotFound`; empty/nil result → `200 []` ("no multi-lingual support"); wrong HTTP method → `405`
- [x] Unit tests added for REST, gRPC, and the server runtime; obsolete `ListLanguages` "Unimplemented" test entries removed
- [x] `go build ./...` and `go test ./...` pass; no linter errors

---

## 🛠️ Implementation Notes

- The handler returns a plain `[]string`; REST encodes it directly via `json.NewEncoder`, while gRPC wraps it in `&protos.LanguageList{Languages: ...}` (reusing the existing proto type).
- Access control mirrors `GetDevice`/`GetValue`: read scope required, gated by a new `EndpointListLanguages` enum value.
- **Behavioral distinction:** an *unregistered* slot yields `404` (consistent with other endpoints), whereas a *registered* handler returning an empty slice yields `200 []`. In REST, nil is coerced to `[]string{}` so the body is always a valid JSON array.
- The test `stubServerRuntime` gained a `listLanguagesFn` + `InvokeListLanguagesHandler` method to keep the `ServerRuntime` interface assertion compiling.

---

## 🔗 Related Issues / Links

Closes issues:
- #1367 
- #1377

### DIFF
--- a/sdks/go/examples/oneOfEverything/languages_handler.go
+++ b/sdks/go/examples/oneOfEverything/languages_handler.go
@@ -9,7 +9,7 @@ import (
 // handler returns the language codes the device model supports; adopters would
 // derive this from their own language packs. Here a small static list is used.
 func registerLanguagesHandlers(srv catena.Server) {
-	supported := []string{"en", "fr", "es", "de"}
+	supported := []string{"en", "fr"}
 	for _, slot := range slotList {
 		srv.RegisterListLanguagesHandler(slot, func(slot uint16, ctx catena.HandlerContext) ([]string, catena.StatusResult) {
 			logger.Info("ListLanguages", "slot", slot)
--- a/sdks/go/examples/oneOfEverything/languages_handler.go
+++ b/sdks/go/examples/oneOfEverything/languages_handler.go
@@ -1,0 +1,19 @@
+package main
+
+import (
+	"github.com/rossvideo/catena/sdks/go/pkg/catena"
+	"github.com/rossvideo/catena/sdks/go/pkg/logger"
+)
+
+// registerLanguagesHandlers wires the Languages endpoint for each slot. The
+// handler returns the language codes the device model supports; adopters would
+// derive this from their own language packs. Here a small static list is used.
+func registerLanguagesHandlers(srv catena.Server) {
+	supported := []string{"en", "fr", "es", "de"}
+	for _, slot := range slotList {
+		srv.RegisterListLanguagesHandler(slot, func(slot uint16, ctx catena.HandlerContext) ([]string, catena.StatusResult) {
+			logger.Info("ListLanguages", "slot", slot)
+			return supported, catena.StatusWithCode(catena.StatusCodeOk, "")
+		})
+	}
+}
--- a/sdks/go/pkg/catena/server.go
+++ b/sdks/go/pkg/catena/server.go
@@ -68,6 +68,7 @@ const (
 	EndpointExecuteCommand
 	EndpointParamInfo
 	EndpointConnect
+	EndpointListLanguages
 )
 
 type TransportContext struct {
@@ -139,6 +140,10 @@ type SetValueHandler func(slot uint16, entries []SetValueEntry, ctx HandlerConte
 type GetAssetHandler func(slot uint16, fqoid string, ctx HandlerContext) (Asset, StatusResult)
 type ExecuteCommandHandler func(slot uint16, commandFqoid string, payload any, ctx HandlerContext) (CommandResult, StatusResult)
 type ParamInfoHandler func(slot uint16, oidPrefix string, recursive bool, ctx HandlerContext) ([]ParamInfo, StatusResult)
+
+// ListLanguagesHandler returns the language codes supported by the device model
+// at a slot (e.g. ["en", "fr"]). An empty slice indicates no multi-lingual support.
+type ListLanguagesHandler func(slot uint16, ctx HandlerContext) ([]string, StatusResult)
 type HeartbeatHandler func(slot uint16)
 type AccessHandler func(endpointType EndpointType, ctx HandlerContext) bool
 
@@ -195,6 +200,7 @@ type Server interface {
 	RegisterGetAssetHandler(slot uint16, handler GetAssetHandler)
 	RegisterExecuteCommandHandler(slot uint16, handler ExecuteCommandHandler)
 	RegisterParamInfoHandler(slot uint16, handler ParamInfoHandler)
+	RegisterListLanguagesHandler(slot uint16, handler ListLanguagesHandler)
 	RegisterHeartbeatHandler(slot uint16, handler HeartbeatHandler)
 	RegisterAccessHandler(handler AccessHandler)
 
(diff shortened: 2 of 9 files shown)


sdks/go/pkg/catena/server.go
```diff
@@ -793,6 +813,31 @@ func (s *server) InvokeParamInfoHandler(slot uint16, oidPrefix string, recursive
 	return nil, StatusWithCode(StatusCodeNotFound, "ParamInfo "+oidPrefix+" not found at slot "+strconv.Itoa(int(slot)))
 }
 
+func (s *server) InvokeListLanguagesHandler(slot uint16, transportContext TransportContext) ([]string, StatusResult) {
+	handlerContext, res := s.resolveHandlerContext(transportContext)
+	if res.Code != StatusCodeOk {
+		return nil, res
```
abrown-RV, line 819: This should get covered, but actually I have helpers for these funcs that rolls up the coverage, so I'll let it by

sdks/go/examples/oneOfEverything/languages_handler.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/rossvideo/catena/sdks/go/pkg/catena"
+	"github.com/rossvideo/catena/sdks/go/pkg/logger"
+)
+
+// registerLanguagesHandlers wires the Languages endpoint for each slot. The
+// handler returns the language codes the device model supports; adopters would
+// derive this from their own language packs. Here a small static list is used.
+func registerLanguagesHandlers(srv catena.Server) {
+	supported := []string{"en", "fr", "es", "de"}
```
abrown-RV: we don't support all these langs

sdks/go/pkg/transports/grpc_transport.go
```diff
@@ -593,9 +593,23 @@ func (s *catenaService) LanguagePackRequest(ctx context.Context, req *protos.Lan
 	return nil, status.Error(codes.Unimplemented, "LanguagePackRequest not implemented")
 }
 
-// ListLanguages returns the list of available languages
+// ListLanguages returns the language codes supported by the device model at a slot.
 func (s *catenaService) ListLanguages(ctx context.Context, req *protos.Slot) (*protos.LanguageList, error) {
-	return nil, status.Error(codes.Unimplemented, "ListLanguages not implemented")
+	slot, err := catena.ValidateSlot(req.GetSlot())
+	if err.Code != catena.StatusCodeOk {
+		return nil, status.Error(ToGRPCCode(err.Code), err.Error)
```
abrown-RV, line 600: coverage

sdks/go/pkg/transports/rest_transport.go
```diff
@@ -514,6 +516,35 @@ func (t *RestTransport) handleGetPopulatedSlots(w http.ResponseWriter, r *http.R
...
+
+	if languages == nil {
+		languages = []string{}
+	}
+
+	response := &protos.LanguageList{Languages: languages}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(response); err != nil {
```
abrown-RV: annoyingly this endpoint breaks the pattern with everything else and doesn't return a proto-like json. Its just the bare array `["en", "fr"]`

abrown-RV (changes requested):

sdks/go/pkg/transports/rest_transport.go
nelgd: Is that how it should be? or should it return a proto-json? I went over it again and got cursor to compare against the spec and update it if needed.

abrown-RV (approved):

PR #1446, opened 2026-07-14
[GO] Child Param read_only bug

### LINKED ISSUE #1445: [GO] Children Params Read-Only Status
## 📖 Description

Children params from the product struct do not have the correct read_only status applied to them.

Image

---

## 🎯 Goal / Outcome

---

## ✅ Definition of Done (DoD)

---

---

## 🔗 Related Issues / Links

---

---

nelgd:
## 📖 Description
Fixes read-only inheritance for sub-params in the Go SDK. Previously, marking a parent param `read_only` did not propagate to its children, so a child of a read-only struct could still be writable. `WithReadOnly` and `WithParam` now
enforce downward inheritance while still allowing a child to be independently read-only under a writable parent.

---

## 🎯 Goal / Outcome
When a parent param is read-only, all of its sub-params (recursively) are guaranteed to be read-only too — regardless of the order in which `WithReadOnly` and `WithParam` are chained. A writable parent (`read_only: false`) leaves children untouched, so a child can still be explicitly marked read-only.

---

## ✅ Definition of Done (DoD)
- [x] `WithReadOnly(true)` propagates `read_only` to all existing sub-params recursively
- [x] `WithParam` forces a newly added child (and its descendants) read-only when the parent is already read-only
- [x] A writable parent does not clear or override a child's own `read_only` flag
- [x] Behavior is order-independent across the fluent builder chain
- [x] Tests added covering both chaining orders and the writable-parent edge cases
- [x] SDK-managed `product` param golden test updated to reflect inherited read-only field descriptors

---

- Added an unexported recursive helper `propagateReadOnly(params map[string]*protos.Param)` that walks the sub-param tree and sets `ReadOnly = true` (nil-safe).
- `WithReadOnly(true)` now calls `propagateReadOnly` over existing children; `WithReadOnly(false)` intentionally leaves children as-is.
- `WithParam` applies `propagateReadOnly` to the cloned child subtree only when the parent is already read-only, preserving the "writable parent can hold a read-only child" semantic.
- Read-only is treated as a one-way (downward, true-only) propagation; it never forces a child back to writable.

---

## 🔗 Related Issues / Links

Closes issue:
- #1445 

---

### DIFF
--- a/sdks/go/pkg/catena/param.go
+++ b/sdks/go/pkg/catena/param.go
@@ -254,9 +254,29 @@ func (cp *Param) WithConstraint(c *Constraint) *Param {
 	return cp
 }
 
+// WithReadOnly sets the param's read_only flag. Read-only is inherited
+// downward: when set to true, every existing sub-param (recursively) is also
+// marked read-only, since a child of a read-only parent cannot be writable.
+// Setting it to false leaves sub-params untouched, so a child may still be
+// independently marked read-only under a writable parent.
 func (cp *Param) WithReadOnly(readOnly bool) *Param {
 	cp.Proto.ReadOnly = readOnly
+	if readOnly {
+		propagateReadOnly(cp.Proto.Params)
+	}
 	return cp
+}
+
+// propagateReadOnly recursively marks every param in the map (and their
+// sub-params) as read-only.
+func propagateReadOnly(params map[string]*protos.Param) {
+	for _, p := range params {
+		if p == nil {
+			continue
+		}
+		p.ReadOnly = true
+		propagateReadOnly(p.Params)
+	}
 }
 
 func (cp *Param) WithWidget(widget string) *Param {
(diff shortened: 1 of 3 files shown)


nelgd:
Got it working and the product child params now show `read_only: true`

Does not pass the js validator so im just working on getting that done right now

nelgd:
Got it to pass the validator, only error im getting now is related to struct variants but its only on slot 2. Not sure if this is related or will be dealt with in a future issue.

image

sdks/go/pkg/catena/param.go
```diff
@@ -254,11 +254,31 @@ func (cp *Param) WithConstraint(c *Constraint) *Param {
...
+	}
 	return cp
 }
 
+// propagateReadOnly recursively marks every param in the map (and their
+// sub-params) as read-only.
+func propagateReadOnly(params map[string]*protos.Param) {
+	for _, p := range params {
+		if p == nil {
+			continue
```
abrown-RV, line 275: cover, I know its dumb, but its low hanging fruit

abrown-RV (changes requested):

abrown-RV (approved):

PR #1476, opened 2026-07-27
[Go] Param Helper Improvements

### LINKED ISSUE #1441: [Go] Param Helper Improvements
The `NewParamX` can be friendlier. Right now if you want to make a struct you must give it the full value *and* define all of the fields. Worst part is that you have to give the sub fields values that are just ignored which looks messy.

```go
NewParamStruct(map[string]any{
	"field1": "value1",
	"field2": 42,
}).
WithParam("field1", NewParamString("")).
WithParam("field2", NewParamInt32(0))
```
It would be nice if you could so something like this
```go
NewParamStruct().
WithParam("field1", NewParamString("value1")).
WithParam("field2", NewParamInt32(42))
```
much cleaner. So to do this we need to change all of the NewParamX funcs to become variadic in accepting a value. If the value is not given then the param is created with no value. If the value is given then it is used to create the param. And importantly it must also save the fact that it was created with a value so we can tell the different between a param with a value by default and a param with a intentionally default value.
```go
type Param struct {
	Proto    *protos.Param
	hasValue bool
}

func NewParamString(value ...string) *Param {
	p := &Param{
		Proto: &protos.Param{
			Type:  protos.ParamType_STRING,
			Value: &protos.Value{Kind: &protos.Value_StringValue{StringValue: ""}},
		},
	}
	if len(value) > 0 {
		p.Value.Value_String_Value.StringValue = value[0]
		p.hasValue = true
	}
	return p
}
```
What this does is allow the other part of this change, which is that `WithParam` should apply the param to the struct only if it has a value. This allows you to build up a struct with only the fields you want to set, and not have to worry about the other fields being set to default values. Also NewParamStruct can automatically make the trivial fields for you if you give it a map of values. So you can do this:	
```go
NewParamStruct(map[string]any{
	"field1": "value1",
	"field2": 42,
	"field3": 0.0
}).
// for fields with more complexity you can still use WithParam to set them up
// note no value in NewParamInt32() so it will not override the value from the map
WithParam("field2", NewParamInt32().WithName(PolyglotText{"en": "Field 2"})).
WithParam("field3", NewParamFloat32(1.2)) // give it a different value for some reason
```

Small discussion can be around what to do with extra variadic values. Maybe they could be ignored, maybe they could be combined together in some way depending on the type. Maybe a logger whine.

nelgd:
## 📖 Description

Makes the `NewParamX` builder API friendlier by making values optional. All `NewParamX` factory functions in `pkg/st2138` are now variadic: calling one with no arguments creates a param without a value, which is ideal for sub-param descriptors that previously required messy dummy values (e.g. `NewParamString("")`).

On a STRUCT parent, `WithParam` now moves a valued sub-param's value into the struct's own `Value.Fields[oid]` and stores the descriptor valueless. `NewParamStruct(map)` also auto-creates typed, valueless field descriptors from the map, so trivial fields need no explicit `WithParam` calls.

Before:

```go
NewParamStruct(map[string]any{
	"field1": "value1",
	"field2": 42,
}).
	WithParam("field1", NewParamString("")).
	WithParam("field2", NewParamInt32(0))
```

After:

```go
NewParamStruct().
	WithParam("field1", NewParamString("value1")).
	WithParam("field2", NewParamInt32(42))
```

or, letting the map auto-create the trivial descriptors and refining one without overriding its value:

```go
NewParamStruct(map[string]any{
	"field1": "value1",
	"field2": int32(42),
}).
	WithParam("field2", NewParamInt32().WithName(PolyglotText{"en": "Field 2"}))
```

---

## 🎯 Goal / Outcome

Struct params can be built field-by-field with only the fields you want to set, without supplying a full value map up front or giving sub-param descriptors throwaway dummy values. Whether a param carries a value is a simple `Proto.Value` nil check — no extra state.

---

## ✅ Definition of Done (DoD)

- [x] All `NewParamX` factories accept an optional (variadic) value; no-arg calls create valueless params
- [x] Extra variadic values: first value wins, extras ignored with a logger warning
- [x] `WithParam` on a STRUCT moves a valued sub-param's value into the parent struct's Value and keeps the descriptor valueless
- [x] Valueless `WithParam` refines a descriptor (name, constraint, ...) without overriding a map-provided value
- [x] `NewParamStruct(map)` auto-creates typed field descriptors (recursive for nested maps)
- [x] Tests added/updated; `go build ./...`, `go vet ./...`, and `go test ./...` all pass
- [x] `ProductParam` and the `oneOfEverything` example updated to the new style

---

- `sdks/go/pkg/st2138/param.go`
  - Converted all factories to variadic (`NewParamInt32(value ...int32)`, etc.) with a shared `pickValue` helper that warns when extra values are ignored
  - `NewParamStructVariant` signature changed from `*StructVariantValue` to variadic `StructVariantValue` (the nil-pointer state is no longer needed)
  - `NewParamStruct(map)` auto-creates valueless sub-param descriptors per field via the new `structFieldDescriptors` helper (recursive for nested structs), with types inferred by `paramTypeFromValueKind`
  - `WithParam` on a STRUCT parent propagates the cloned sub-param's value into `Value.Fields[oid]` and strips it from the stored descriptor; gated on a simple `cloned.Value != nil` check (per review — no `hasValue` field)
- `sdks/go/pkg/catena/product.go`
  - `ProductParam` simplified to a single `NewParamStruct(ProductValues(p))` call; removed the manual descriptor loop and `productFieldDescriptor()`
- `sdks/go/examples/oneOfEverything/device_handler.go`
  - `struct_example` drops its manual sub-params (auto-descriptors cover them); struct-variant/array sub-param descriptors now use valueless `NewParamInt32()` / `NewParamString()`
- `sdks/go/pkg/st2138/param_test.go`, `sdks/go/pkg/catena/param_info_test.go`
  - Updated call sites and assertions to the new value-location semantics; new coverage for no-arg factories, extra-value handling, auto-descriptor creation (incl. nested and the empty-fields guard), value move/strip on `WithParam`, valueless refine vs. valued override, and a `proto.Equal` check that the clean builder style matches passing the full map up front

---

Closes issue:
- #1441

### DIFF
--- a/sdks/go/pkg/st2138/param.go
+++ b/sdks/go/pkg/st2138/param.go
@@ -48,20 +48,8 @@ import (
 )
 
 // Param wraps a protos.Param and exposes a fluent builder API.
-//
-// hasValue records whether the param was intentionally given a value (via a
-// factory argument, WithValue, or SetValue), which distinguishes an
-// intentional default value from a param that was simply created without one.
-// WithParam uses this to decide whether a sub-param contributes its value to
-// a parent STRUCT's Value.
 type Param struct {
-	Proto    *protos.Param
-	hasValue bool
-}
-
-// HasValue reports whether the param was intentionally given a value.
-func (cp *Param) HasValue() bool {
-	return cp.hasValue
+	Proto *protos.Param
 }
 
 // --- Factory functions (one per ParamType) ---
@@ -85,7 +73,6 @@ func NewParamInt32(value ...int32) *Param {
 	cp := &Param{Proto: &protos.Param{Type: protos.ParamType_INT32}}
 	if len(value) > 0 {
 		cp.Proto.Value = &protos.Value{Kind: &protos.Value_Int32Value{Int32Value: pickValue("NewParamInt32", value)}}
-		cp.hasValue = true
 	}
 	return cp
 }
@@ -94,7 +81,6 @@ func NewParamFloat32(value ...float32) *Param {
 	cp := &Param{Proto: &protos.Param{Type: protos.ParamType_FLOAT32}}
 	if len(value) > 0 {
 		cp.Proto.Value = &protos.Value{Kind: &protos.Value_Float32Value{Float32Value: pickValue("NewParamFloat32", value)}}
-		cp.hasValue = true
 	}
 	return cp
 }
@@ -103,7 +89,6 @@ func NewParamString(value ...string) *Param {
 	cp := &Param{Proto: &protos.Param{Type: protos.ParamType_STRING}}
 	if len(value) > 0 {
 		cp.Proto.Value = &protos.Value{Kind: &protos.Value_StringValue{StringValue: pickValue("NewParamString", value)}}
-		cp.hasValue = true
 	}
 	return cp
 }
@@ -124,7 +109,6 @@ func NewParamStruct(value ...map[string]any) *Param {
 		return cp
 	}
 	cp.Proto.Value = pv
-	cp.hasValue = true
 	cp.Proto.Params = structFieldDescriptors(pv.GetStructValue())
 	return cp
 }
@@ -161,7 +145,6 @@ func NewParamStructVariant(value ...StructVariantValue) *Param {
 		return cp
 	}
 	cp.Proto.Value = pv
-	cp.hasValue = true
 	return cp
 }
 
@@ -169,7 +152,6 @@ func NewParamInt32Array(value ...[]int32) *Param {
 	cp := &Param{Proto: &protos.Param{Type: protos.ParamType_INT32_ARRAY}}
 	if len(value) > 0 {
 		cp.Proto.Value = &protos.Value{Kind: &protos.Value_Int32ArrayValues{Int32ArrayValues: &protos.Int32List{Ints: pickValue("NewParamInt32Array", value)}}}
-		cp.hasValue = true
 	}
 	return cp
 }
@@ -178,7 +160,6 @@ func NewParamFloat32Array(value ...[]float32) *Param {
 	cp := &Param{Proto: &protos.Param{Type: protos.ParamType_FLOAT32_ARRAY}}
 	if len(value) > 0 {
 		cp.Proto.Value = &protos.Value{Kind: &protos.Value_Float32ArrayValues{Float32ArrayValues: &protos.Float32List{Floats: pickValue("NewParamFloat32Array", value)}}}
-		cp.hasValue = true
 	}
 	return cp
 }
@@ -187,7 +168,6 @@ func NewParamStringArray(value ...[]string) *Param {
 	cp := &Param{Proto: &protos.Param{Type: protos.ParamType_STRING_ARRAY}}
 	if len(value) > 0 {
 		cp.Proto.Value = &protos.Value{Kind: &protos.Value_StringArrayValues{StringArrayValues: &protos.StringList{Strings: pickValue("NewParamStringArray", value)}}}
-		cp.hasValue = true
 	}
 	return cp
 }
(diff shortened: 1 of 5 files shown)


sdks/go/pkg/st2138/param.go
```diff
@@ -48,138 +48,188 @@ import (
...
+//
+// hasValue records whether the param was intentionally given a value (via a
+// factory argument, WithValue, or SetValue), which distinguishes an
+// intentional default value from a param that was simply created without one.
+// WithParam uses this to decide whether a sub-param contributes its value to
+// a parent STRUCT's Value.
 type Param struct {
-	Proto *protos.Param
+	Proto    *protos.Param
+	hasValue bool
```
abrown-RV: actually looking at this, I've realized my internal "params must always have values" assumption is not so true, so this hasValue field I suggested is not actually necessary and can be a simple p.Value == nil check instead.

sdks/go/pkg/st2138/param.go
```diff
@@ -48,138 +48,188 @@ import (
...
+			p.Params = structFieldDescriptors(nested)
 		}
-		cp.Proto.Value = pv
+		params[oid] = p
+	}
+	return params
+}
+
+func NewParamStructVariant(value ...StructVariantValue) *Param {
+	cp := &Param{Proto: &protos.Param{Type: protos.ParamType_STRUCT_VARIANT}}
```
abrown-RV, line 137: this should get some of the same treatment as struct... but fine to make into a follow-on issue
```go
NewParamStructVariant(StructVariantValue{
    StructVariantType: "left",
    Value: 0.5,
})
```
should produce
```json
{
    "type": "STRUCT_VARIANT",
    "params": {
        "left": {
            "type": "FLOAT",
        }
    },
    "value": {
        "struct_variant_value": {
            "struct_variant_type": "left",
            "value": {
                "float_value": 0.5
            }
        }
    }
}
```
right now it would get everything but the `params`. Again, fine to do in a follow-on issue, just establishing the starting for struct is enough for 1 PR.

sdks/go/pkg/st2138/param.go
```diff
@@ -48,138 +48,188 @@ import (
...
 	}
+	return cp
 }
 
-func NewParamStructArray(value []map[string]any) *Param {
-	cp := &Param{
-		Proto: &protos.Param{
-			Type: protos.ParamType_STRUCT_ARRAY,
-		},
+func NewParamStructArray(value ...[]map[string]any) *Param {
```
abrown-RV, line 183: same filling in of `params` as NewStruct, but bundle in a follow-on like variant

sdks/go/pkg/st2138/param.go
```diff
@@ -48,138 +48,188 @@ import (
...
-				"error", err)
-			return cp
+// structFieldDescriptors builds a valueless sub-param descriptor for each
+// field of a struct value, inferring each descriptor's type from the field's
+// value kind. Nested struct fields recurse so their fields get descriptors
+// too.
+func structFieldDescriptors(sv *protos.StructValue) map[string]*protos.Param {
+	fields := sv.GetFields()
+	if len(fields) == 0 {
+		return nil
```
abrown-RV, line 123: coverage, or maybe can't happen?

sdks/go/pkg/st2138/param.go
```diff
@@ -326,20 +379,42 @@ var paramTypesWithSubParams = map[protos.ParamType]struct{}{
...
-
 	if _, ok := paramTypesWithSubParams[cp.Proto.Type]; !ok {
 		logger.Warning("WithParam called on param type that does not support sub-params; ignoring",
 			"param_type", cp.Proto.Type.String())
 		return cp
 	}
+
+	cloned := proto.Clone(param.Proto).(*protos.Param)
+
+	if cp.Proto.Type == protos.ParamType_STRUCT && param.hasValue && cloned.Value != nil {
```
abrown-RV: case in point for why hasValue isn't needed, you can just check cloned.Value

abrown-RV (changes requested):
very nice, thanks for taking this on. really makes the building code simpler

abrown-RV (approved):

PR #1489, opened 2026-07-31
[GO] Streamed Responses

### LINKED ISSUE #1272: [Go] DeviceRequest stream
## 📖 Description

DeviceRequest should give the BL the ability to stream back chunks like the CPP does. Works fine as is, but the st2138 allows the device to send back the full device definition in pieces instead of all at once:
```
  oneof kind {
    Device device = 2;
    ComponentParam param = 3;
    ComponentConstraint shared_constraint = 4;
    ComponentMenu menu = 5;
    ComponentCommand command = 6;
    ComponentLanguagePack language_pack = 7;
  }
```

---

## 🎯 Goal / Outcome

streaming of device chunks

---

## ✅ Definition of Done (DoD)
- [ ] Handler can stream back device chunks
- [ ] GRPC streams
- [ ] REST /{slot}/stream
- [ ] Code is reviewed and approved -->

---

---

## 🔗 Related Issues / Links

---

---

### LINKED ISSUE #1273: [Go] Streamed responses
## 📖 Description

Bunch of handlers need to handle streaming a response:
- asset
- device request (https://github.com/rossvideo/Catena/issues/1272)
- execute command

---

## 🎯 Goal / Outcome

implement streaming properly

---

## ✅ Definition of Done (DoD)
- [ ] Asset
- [ ] DeviceRequest
- [ ] ExecuteCommand
- [ ] ParamInfo
- [ ] Unit Tests

---

Need to come up with a common way for handlers to yield their response instead of a single return. Common pattern I found with a little looking was a yield function (go doesn't have the yield keyword) so you make your handler accept a func that they can call to send data.
```go
func Handler(yield func(payload: PayloadType)) error {
    while streaming {
        data := NewPayloadType()
        yield(data)
    }
    return nil
}

func yield(payload: PayloadType) {
    # handle streaming to grpc/rest
}
# call handler
err := Handler(yield)
```

---

## 🔗 Related Issues / Links

---

---

nelgd:
## 📖 Description

Converts the Go SDK's `DeviceRequest`  and `Asset` handlers from single-return signatures to the streaming `Stream[T]` "yield" pattern already used by `ExecuteCommand` and `ParamInfo`. Handlers now receive a stream and call `stream.Send(chunk)` for each response chunk, which flows end-to-end through both the gRPC and REST transports.

Key pieces:
- New `st2138.DeviceComponent` chunk type with constructors mirroring the proto oneof (`ComponentDevice`, `ComponentParam`, `ComponentConstraint`, `ComponentMenu`, `ComponentCommand`, `ComponentLanguagePack`), and `Wire()` on `st2138.Asset` so both satisfy `Message`.
- `DeviceHandler` / `ReadAssetHandler` signatures replaced in place; `InvokeGetDeviceHandler` / `InvokeReadAssetHandler` now take a stream, wrap it in `shutdownStream`, and return only a `StatusResult`.
- SDK-managed product-struct injection moved from a post-return mutation into a `productDeviceStream` wrapper that enforces the `st2138:mon` scope policy per chunk (inject/overwrite on Device chunks, rewrite or drop product `ComponentParam` chunks).
- gRPC: `DeviceRequest` and `ExternalObjectRequest` forward each `Send` as one gRPC message via `grpcStream` adapters.
- REST: unary routes aggregate the stream into one JSON body (`deviceAggregateStream` merges components into a full device, order-independently; `assetAggregateStream` concatenates payload bytes, first chunk wins for metadata/digest/cachable). New SSE routes `GET /{slot}/stream` and `GET /{slot}/asset/{oid}/stream`. `?compression=` still applies to the assembled unary asset payload.

---

## 🎯 Goal / Outcome

All four server-streaming endpoints (`DeviceRequest`, `ExternalObjectRequest`, `ExecuteCommand`, `ParamInfoRequest`) expose a consistent streaming handler API, so business logic can deliver large device models and assets in chunks instead of building one monolithic response — completing the remaining work from #1273 (DeviceRequest DoD in #1272).

---

## ✅ Definition of Done (DoD)

- [x] `DeviceHandler` and `ReadAssetHandler` converted to `Stream[T]` signatures matching the ExecuteCommand/ParamInfo precedent
- [x] gRPC `DeviceRequest` and `ExternalObjectRequest` stream one proto per handler `Send`
- [x] REST unary routes aggregate multi-chunk streams into a single JSON response; new `/stream` SSE routes for device and asset
- [x] Product-struct injection/stripping preserved through the streaming path (mon-scope policy on Device and product param chunks)
- [x] Tests added or updated across `st2138`, `catena`, gRPC, REST, and the transport test stub
- [x] Documentation updated (`GRPC.md`, `REST.md`, transports `README.md`, `Server` interface doc comments) and examples reworked (`hello_world`, `oneOfEverything`)

---

- **Breaking API change (intentional):** the unary handler signatures were replaced in place rather than adding parallel streaming variants, per the ExecuteCommand/ParamInfo precedent. Existing adopters must migrate their `RegisterGetDeviceHandler` / `RegisterReadAssetHandler` callbacks.
- REST unary aggregation lives in new stream adapters (`deviceAggregateStream`, `assetAggregateStream`) alongside the existing `firstStream`/`lastStream`, keeping the transport's "adapt the stream to the route shape" pattern.
- `setNestedParam` proto-merges into existing entries so component chunk order doesn't matter (a parent arriving after a nested child keeps the synthesized child) — hardened in the follow-up commit with order-independence tests.
- SSE device frames use a new `MarshalDeviceComponentJSON` with the same SMPTE cleanup rules as the unary device marshaller (meaningful zeros survive, schema-forbidden defaults stripped).
- The `oneOfEverything` example demonstrates both chunked forms: device skeleton + per-param/command components, and 64 KiB-chunked asset payloads; `hello_world` shows the simple single-`ComponentDevice` case.

---

## 🔗 Related Issues / Links

Closes issues:
- #1272 
- #1273

### DIFF
--- a/sdks/go/examples/oneOfEverything/asset_handler.go
+++ b/sdks/go/examples/oneOfEverything/asset_handler.go
@@ -35,15 +35,17 @@ func registerAssetHandlers(srv catena.Server, assets *sync.Map) {
 			// Stream the asset in chunks to demonstrate large-object delivery:
 			// the first chunk carries the metadata, digest, encoding, and
 			// cachable flag plus the first slice of payload bytes; later
-			// chunks carry only payload bytes, which the client concatenates
-			// in send order. Assets up to assetChunkSize (and URL-kind assets,
-			// which have no embedded bytes) go out as a single chunk.
+			// chunks carry only payload bytes (never the URL: a payload can
+			// have one or the other, and continuation chunks only ever exist
+			// for embedded payloads), which the client concatenates in send
+			// order. Assets up to assetChunkSize (and URL-kind assets, which
+			// have no embedded bytes) go out as a single chunk.
 			stored := val.(storedAsset)
 			data := stored.payload.Payload
 			sent := 0
 			for first := true; first || sent < len(data); first = false {
 				end := min(sent+assetChunkSize, len(data))
-				dp := st2138.DataPayload{Url: stored.payload.Url, Payload: data[sent:end]}
+				dp := st2138.DataPayload{Payload: data[sent:end]}
 				if first {
 					dp = stored.payload
 					dp.Payload = data[sent:end]
--- a/sdks/go/examples/oneOfEverything/asset_handler.go
+++ b/sdks/go/examples/oneOfEverything/asset_handler.go
@@ -16,26 +16,52 @@ type storedAsset struct {
 	cachable bool
 }
 
+// assetChunkSize is the maximum number of embedded payload bytes streamed per
+// ReadAsset chunk. Kept small enough to bound per-message memory but large
+// enough that small assets go out in a single chunk.
+const assetChunkSize = 64 * 1024
+
 func registerAssetHandlers(srv catena.Server, assets *sync.Map) {
 	// Slot 0-2: direct map-backed lookup, matching the sync.Map data-model example.
 	for _, slot := range slotList {
-		srv.RegisterReadAssetHandler(slot, func(slot uint16, fqoid string, ctx catena.HandlerContext) (st2138.Asset, catena.StatusResult) {
+		srv.RegisterReadAssetHandler(slot, func(slot uint16, fqoid string, ctx catena.HandlerContext, stream catena.Stream[st2138.Asset]) catena.StatusResult {
 			logger.Info("Asset download request", "slot", slot, "fqoid", fqoid)
 			val, ok := assets.Load(fqoid)
 			if !ok {
 				logger.Warning("Asset not found", "slot", slot, "fqoid", fqoid)
-				return catena.ReplyError[st2138.Asset](catena.StatusCodeNotFound, "asset not found: "+fqoid)
+				return catena.StatusWithCode(catena.StatusCodeNotFound, "asset not found: "+fqoid)
 			}
 
+			// Stream the asset in chunks to demonstrate large-object delivery:
+			// the first chunk carries the metadata, digest, encoding, and
+			// cachable flag plus the first slice of payload bytes; later
+			// chunks carry only payload bytes, which the client concatenates
+			// in send order. Assets up to assetChunkSize (and URL-kind assets,
+			// which have no embedded bytes) go out as a single chunk.
 			stored := val.(storedAsset)
-			catenaAsset, err := st2138.ToAsset(stored.payload, stored.cachable)
-			if err != nil {
-				logger.Error("Failed to convert payload to asset", "slot", slot, "fqoid", fqoid, "error", err)
-				return catena.ReplyError[st2138.Asset](catena.StatusCodeInternal, "failed to convert asset: "+err.Error())
+			data := stored.payload.Payload
+			sent := 0
+			for first := true; first || sent < len(data); first = false {
+				end := min(sent+assetChunkSize, len(data))
+				dp := st2138.DataPayload{Url: stored.payload.Url, Payload: data[sent:end]}
+				if first {
+					dp = stored.payload
+					dp.Payload = data[sent:end]
+				}
+				chunk, err := st2138.ToAsset(dp, stored.cachable)
+				if err != nil {
+					logger.Error("Failed to convert payload to asset", "slot", slot, "fqoid", fqoid, "error", err)
+					return catena.StatusWithCode(catena.StatusCodeInternal, "failed to convert asset: "+err.Error())
+				}
+				if err := stream.Send(chunk); err != nil {
+					logger.Warning("Asset download stream closed", "slot", slot, "fqoid", fqoid, "error", err)
+					return catena.StatusWithCode(catena.StatusCodeInternal, "failed to send asset: "+err.Error())
+				}
+				sent = end
 			}
 
-			logger.Info("Asset download complete", "slot", slot, "fqoid", fqoid, "size", len(stored.payload.Payload))
-			return catena.Reply(catenaAsset)
+			logger.Info("Asset download complete", "slot", slot, "fqoid", fqoid, "size", len(data))
+			return catena.StatusWithCode(catena.StatusCodeOk, "")
 		})
 
 		// POST / CreateAsset: create a new asset, conflict if one already exists.
--- a/sdks/go/examples/oneOfEverything/device_handler.go
+++ b/sdks/go/examples/oneOfEverything/device_handler.go
@@ -17,17 +17,46 @@ func registerProductStructs(srv catena.Server, state *ExampleState) {
 }
 
 func registerDeviceHandlers(srv catena.Server, counter *CounterState, state *ExampleState) {
-	// GetDeviceHandler returns the complete device descriptor for a slot.
-	// Build the descriptor from the same data model your app uses at runtime;
-	// this example rebuilds per request so "value" fields stay current.
+	// GetDeviceHandler streams the device descriptor for a slot as
+	// DeviceComponent chunks. Build the descriptor from the same data model
+	// your app uses at runtime; this example rebuilds per request so "value"
+	// fields stay current.
 	for _, slot := range slotList {
-		srv.RegisterGetDeviceHandler(slot, func(slot uint16, ctx catena.HandlerContext) (st2138.Device, catena.StatusResult) {
+		srv.RegisterGetDeviceHandler(slot, func(slot uint16, ctx catena.HandlerContext, stream catena.Stream[st2138.DeviceComponent]) catena.StatusResult {
 			logger.Info("GetDevice", "slot", slot)
 			device, ok := buildDeviceDefinition(slot, counter, state)
 			if !ok {
-				return catena.ReplyError[st2138.Device](catena.StatusCodeNotFound, "device not found")
+				return catena.StatusWithCode(catena.StatusCodeNotFound, "device not found")
 			}
-			return catena.Reply(*device)
+
+			// Stream the model in components to demonstrate the chunked form
+			// of DeviceRequest: first the device skeleton (everything except
+			// params and commands), then one chunk per param and command.
+			// Clients merge the components back into one device model; a small
+			// model could equally be sent as a single ComponentDevice chunk
+			// carrying the whole device (see the hello_world example).
+			params := device.Proto.Params
+			commands := device.Proto.Commands
+			device.Proto.Params = nil
+			device.Proto.Commands = nil
+
+			if err := stream.Send(st2138.ComponentDevice(device)); err != nil {
+				logger.Warning("GetDevice stream closed", "slot", slot, "error", err)
+				return catena.StatusWithCode(catena.StatusCodeInternal, "failed to send device: "+err.Error())
+			}
+			for oid, param := range params {
+				if err := stream.Send(st2138.ComponentParam(oid, &st2138.Param{Proto: param})); err != nil {
+					logger.Warning("GetDevice stream closed", "slot", slot, "oid", oid, "error", err)
+					return catena.StatusWithCode(catena.StatusCodeInternal, "failed to send param: "+err.Error())
+				}
+			}
+			for oid, command := range commands {
+				if err := stream.Send(st2138.ComponentCommand(oid, &st2138.Param{Proto: command})); err != nil {
+					logger.Warning("GetDevice stream closed", "slot", slot, "oid", oid, "error", err)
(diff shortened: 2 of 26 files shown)


sdks/go/examples/oneOfEverything/device_handler.go
```diff
@@ -17,17 +17,46 @@ func registerProductStructs(srv catena.Server, state *ExampleState) {
...
 			}
-			return catena.Reply(*device)
+
+			// Stream the model in components to demonstrate the chunked form
+			// of DeviceRequest: first the device skeleton (everything except
+			// params and commands), then one chunk per param and command.
+			// Clients merge the components back into one device model; a small
+			// model could equally be sent as a single ComponentDevice chunk
+			// carrying the whole device (see the hello_world example).
+			params := device.Proto.Params
```
abrown-RV: can you make a `DeviceComponentsForRequest` or something, similar to param info's `ParamInfosForRequest` helper. BL shouldn't need to touch protos unless they're being extra fancy, so a helper to bridge our "no device model, but actually we kind of have one" gap.

Also make at least one slot (probably 0?) actually call the st2138.ComponentXxxxxx wrappers to show that method as well.... hopefully not too big a fallout with incompatibility with buildDeviceDefintion.... we can talk about it.

sdks/go/pkg/catena/stream.go
```diff
@@ -98,3 +99,53 @@ var _ Stream[Message] = (*nullStream[Message])(nil)
...
+	switch kind := chunk.Proto.Kind.(type) {
+	case *protos.DeviceComponent_Device:
+		if device := kind.Device; device != nil {
+			if s.hasMon {
+				(&st2138.Device{Proto: device}).WithParam(ProductOid, ProductParam(s.product))
+			} else {
+				delete(device.Params, ProductOid)
+			}
+		}
+	case *protos.DeviceComponent_Param:
```
abrown-RV, line 144: this can be simpler, just swallow all `isProductOid` true, since you already sent it with the device. doesn't need to switch on mon.

sdks/go/pkg/catena/stream.go
```diff
@@ -98,3 +99,53 @@ var _ Stream[Message] = (*nullStream[Message])(nil)
...
+}
+
+var _ Stream[st2138.DeviceComponent] = productDeviceStream{}
+
+// Send applies the product policy to the chunk, then delivers it to the inner
+// stream (or drops it, returning nil, when the policy says the caller may not
+// see it).
+func (s productDeviceStream) Send(chunk st2138.DeviceComponent) error {
+	if chunk.Proto == nil {
+		return s.inner.Send(chunk)
```
abrown-RV, line 130: coverage

sdks/go/examples/oneOfEverything/asset_handler.go
```diff
@@ -16,26 +16,52 @@ type storedAsset struct {
 	cachable bool
 }
 
+// assetChunkSize is the maximum number of embedded payload bytes streamed per
+// ReadAsset chunk. Kept small enough to bound per-message memory but large
+// enough that small assets go out in a single chunk.
+const assetChunkSize = 64 * 1024
```
abrown-RV, line 22: test this with dashboard before we roll it out with examples, we don't want to break compatibility there.

sdks/go/pkg/transports/rest/json_utils.go
```diff
@@ -119,6 +119,32 @@ func MarshalAssetJSON(asset *protos.ExternalObjectPayload) ([]byte, error) {
...
 }
 
+// MarshalDeviceComponentJSON marshals one streamed DeviceComponent chunk to
+// JSON with the same SMPTE-compliant treatment as MarshalDeviceJSON: it uses
+// EmitUnpopulated so meaningful proto3 defaults survive (e.g. a constraint's
+// min_value:0, or a current value of 0/0.0/""), then strips the
+// schema-forbidden extras from whichever oneof shape the chunk carries.
+func MarshalDeviceComponentJSON(component *protos.DeviceComponent) ([]byte, error) {
+	if component == nil {
+		return nil, nil
```
abrown-RV, line 129: coverage

sdks/go/pkg/transports/rest/rest_transport.go
```diff
@@ -629,32 +722,96 @@ func (t *Transport) handleAssetEndpoint(w http.ResponseWriter, r *http.Request,
...
+		} else if err := stream.sendError(result); err != nil {
+			logger.Error("failed to send asset SSE error event", "slot", slot, "fqoid", fqoid, "error", err)
+		}
+		return
+	}
+
+	// A successful handler that produced no chunks still gets a well-formed empty
+	// event stream so the client receives a valid 200 response.
+	if stream.sent == 0 {
+		stream.writeHeaders()
```
abrown-RV: coverage

sdks/go/pkg/transports/rest/rest_transport.go
```diff
@@ -606,13 +614,98 @@ func (t *Transport) handleValuesEndpoint(w http.ResponseWriter, r *http.Request,
...
+	stream := &restStream[st2138.DeviceComponent]{
+		w:       w,
+		flusher: flusher,
+		marshal: marshalDeviceComponentWire,
+		ctx:     r.Context(),
+		devMode: t.isDevMode(),
+	}
+	result := t.runtime.InvokeGetDeviceHandler(slot, stream, transportContext)
+
+	if result.IsError() {
```
abrown-RV: The stream code is almost all the same for each endpoint, other than the invoke call, see if you can aggregate into a handler.... kinda true for unary too for that matter we have a special stream that resolves to a single value for making into a response... see if there's something to be done there as well.

sdks/go/pkg/transports/rest/stream.go
```diff
@@ -170,3 +173,192 @@ func (s *lastStream[T]) Send(chunk T) error {
...
+
+var _ catena.Stream[st2138.DeviceComponent] = (*deviceAggregateStream)(nil)
+
+// Send merges one DeviceComponent chunk into the accumulating device. A Device
+// chunk becomes the base (a second Device chunk is proto-merged into it);
+// component chunks are placed at the location their oid names, creating the
+// device skeleton and any intermediate params on demand so chunk order does
+// not matter. Chunks with a nil body are ignored.
+func (s *deviceAggregateStream) Send(chunk st2138.DeviceComponent) error {
+	if chunk.Proto == nil {
```
abrown-RV, line 302: coverage, lots missing in this func

abrown-RV (changes requested):

sdks/go/examples/oneOfEverything/asset_handler.go
nelgd: How would I test this? I dont think dashboard loads any assets for the client, but not 100% sure.

Also changing param values from Dashboard is broken right now from leading slashes on params. Trying to work out a fix for that right now.

abrown-RV: the grid file, with that special oid_alias

nelgd: I dont think there is a grid file for go. Dashboard auto makes the ui im pretty sure. It might be a problem with the setValue handler?

nelgd: Tested with Dashboard and didn't have any issues. I added a read-only `dashboard_ui` param to slot 0 with the special `0xFF0E` oid alias pointing at `eo://oneofeverything.grid`. I changed the chunk size around a bit so the file would send in multiple chunks or in one. Didnt have any issues.

nelgd:
I had to add in a way to remove the leading slash from params for gRPC so that dashboard could set the param values since it kept adding a leading slash to all the oids.

sdks/go/examples/oneOfEverything/asset_handler.go
```diff
@@ -16,26 +16,52 @@ type storedAsset struct {
...
 			stored := val.(storedAsset)
-			catenaAsset, err := st2138.ToAsset(stored.payload, stored.cachable)
-			if err != nil {
-				logger.Error("Failed to convert payload to asset", "slot", slot, "fqoid", fqoid, "error", err)
-				return catena.ReplyError[st2138.Asset](catena.StatusCodeInternal, "failed to convert asset: "+err.Error())
+			data := stored.payload.Payload
+			sent := 0
+			for first := true; first || sent < len(data); first = false {
+				end := min(sent+assetChunkSize, len(data))
+				dp := st2138.DataPayload{Url: stored.payload.Url, Payload: data[sent:end]}
```
abrown-RV: st2138.ToAsset() will err if both Url and Payload is set. shouldn't be setting Url here.

sdks/go/examples/oneOfEverything/device_handler.go
```diff
@@ -17,25 +17,125 @@ func registerProductStructs(srv catena.Server, state *ExampleState) {
...
-			return catena.Reply(*device)
+			return catena.DeviceComponentsForRequest(device, stream)
 		})
 	}
 }
 
+// namedParam pairs an oid with its descriptor so slot 0's params and commands
+// can be listed once and consumed two ways: attached to a Device by
+// buildDeviceDefinition, or streamed chunk-by-chunk by the GetDevice handler.
+type namedParam struct {
```
abrown-RV: just use a map? saves declaring a new type.

sdks/go/examples/oneOfEverything/device_handler.go
```diff
@@ -17,25 +17,125 @@ func registerProductStructs(srv catena.Server, state *ExampleState) {
...
+// leading ComponentDevice chunk; buildDeviceDefinition attaches the params and
+// commands to it.
+func slotZeroSkeleton() *st2138.Device {
+	return st2138.NewDevice(0).
+		WithDetailLevel(st2138.DetailLevelFull).
+		WithMultiSetEnabled(true).
+		WithSubscriptions(true).
+		WithAccessScopes("st2138:mon", "st2138:op", "st2138:cfg", "st2138:adm").
+		WithDefaultScope("st2138:cfg").
+		WithMenuGroup("status", st2138.NewMenuGroup().
```
abrown-RV, line 98: menus should be sent separately as well. groups are in the skeleton, menus as components

sdks/go/pkg/catena/device_components.go
```diff
@@ -0,0 +1,168 @@
...
+		if group != nil {
+			group.Menus = nil
+		}
+	}
+	return skeleton
+}
+
+// sortedKeys returns m's keys in ascending order so component chunks are
+// emitted deterministically.
+func sortedKeys[V any](m map[string]V) []string {
```
abrown-RV: while I appreciate this makes testing easier, this is unnecessary cpu cycles for a real device.

sdks/go/pkg/catena/stream.go
```diff
@@ -98,3 +99,49 @@ var _ Stream[Message] = (*nullStream[Message])(nil)
...
+	}
+	switch kind := chunk.Proto.Kind.(type) {
+	case *protos.DeviceComponent_Device:
+		if device := kind.Device; device != nil {
+			if s.hasMon {
+				(&st2138.Device{Proto: device}).WithParam(ProductOid, ProductParam(s.product))
```
abrown-RV:
```suggestion
				device.Params[ProductOid] = ProductParam(s.product).Proto
```

sdks/go/pkg/catena/server.go
```diff
@@ -999,30 +1020,33 @@ func (s *server) productForSlot(slot uint16) (ProductStruct, bool) {
...
+			// callers with mon get the SDK-managed product injected into Device
+			// chunks (overwriting whatever the business logic sent); callers
+			// without mon get no product param at all, even one the business logic
+			// added, since the SDK-managed product struct is read-only and
+			// mon-scoped regardless of who created it.
+			if product, has := s.productForSlot(slot); has {
+				out = productDeviceStream{
+					inner:   stream,
+					product: product,
+					hasMon:  ctx.RequireReadScope(st2138.ScopeMon).IsOk(),
```
abrown-RV:
```suggestion
					hasMon:  ctx.HasReadScope(st2138.ScopeMon),
```

abrown-RV (changes requested):
nice work on the rest streaming aggregators

abrown-RV (approved):

PR #1513, opened 2026-09-17
Implement Helper to Stream Asset Chunks

### LINKED ISSUE #1498: [Go] stream helper for chunking assets
## 📖 Description

The Go `oneOfEverything` example's `ReadAsset` handler hand-rolls the logic to split a stored `DataPayload` into chunks and send each one over the stream (see `sdks/go/examples/oneOfEverything/asset_handler.go`). This chunking loop — deciding the chunk size, carrying metadata/digest/encoding/cachable on the first chunk, emitting payload-only continuation chunks, and short-circuiting for URL-kind assets — is generic SDK plumbing, not example-specific business logic.

This is exactly the same pattern `catena.ParamInfosForRequest` addresses for `ParamInfo` handlers: a reusable business-logic helper that a handler delegates to and returns the terminal `StatusResult` from. We should provide an equivalent helper for streaming an asset so handler authors don't have to re-implement (and potentially get wrong) the chunking loop.

---

## 🎯 Goal / Outcome

A `catena` package helper (e.g. `catena.ChunkAsset` / `catena.SendAssetChunks`) that takes a `DataPayload` (plus `cachable`) and a `Stream[st2138.Asset]`, streams it in bounded chunks following the ST 2138 asset semantics, and returns a `StatusResult` the handler can return directly. The `oneOfEverything` example's `ReadAsset` handler should collapse down to a single delegating call, mirroring `param_info_handler.go`.

---

## ✅ Definition of Done (DoD)
- [ ] Helper added to the `catena` package alongside `ParamInfosForRequest`
- [ ] First chunk carries metadata/digest/encoding/cachable; continuation chunks carry payload bytes only
- [ ] URL-kind payloads (no embedded bytes) and assets smaller than the chunk size go out as a single chunk
- [ ] Returns `Ok` on completion, and maps a failed `Send` to `Internal`, stopping the walk immediately
- [ ] `oneOfEverything` `ReadAsset` handler refactored to delegate to the helper
- [ ] Unit tests added (mirroring `param_info_test.go` coverage: single chunk, multi-chunk, URL kind, send failure)

---

- Model the API and doc-comment style on `catena.ParamInfosForRequest` in `sdks/go/pkg/catena/param_info.go`.
- The current inline implementation lives in `sdks/go/examples/oneOfEverything/asset_handler.go` (the `for first := true; ...` loop) and already uses `st2138.ToAsset(dp, cachable)` to build each chunk.
- Decide where `assetChunkSize` (currently `64 * 1024` in the example) should live — likely a package-level default in the helper, optionally overridable.
- Consider signature options: passing a `DataPayload` + `cachable` (matching the example's `storedAsset`) vs. accepting an already-built `Asset` and re-splitting it.

---

## 🔗 Related Issues / Links

- Reference implementation to mirror: `catena.ParamInfosForRequest` (`sdks/go/pkg/catena/param_info.go`), delegated to from `sdks/go/examples/oneOfEverything/param_info_handler.go`
- Current hand-rolled chunking: `sdks/go/examples/oneOfEverything/asset_handler.go`
- Asset conversion helpers: `st2138.ToAsset` / `st2138.FromAsset` (`sdks/go/pkg/st2138/asset.go`)

---

- Should the chunk size be a fixed default, a package constant, or a configurable parameter?
- Should the helper accept `DataPayload` + `cachable`, or a fully-formed `st2138.Asset`?

---

## 📝 Additional Context (Optional)

The existing example handler already documents the chunking contract in comments (first chunk = metadata + first slice; later chunks = payload bytes only; URL-kind and small assets = single chunk). That description can seed the helper's doc comment.

oshah-rv:
## 📖 Description
This PR adds a reusable asset chunking helper to the Go SDK and refactors the oneOfEverything example's asset_handler to use it. 

Adds three functions:
- Two public functions: `SendAssetChunks()` and `SendAssetChunksWithSize()`
- One private helper: `sendChunks()`

`SendAssetChunks()` provides the default chunking behaviour for for embedded payloads, while `SendAssetChunksWithSize()` provides a public override for the chunk-size, allowing callers to set a custom chunk size.

The private `sendChunks()` method implements the actual streaming logic, which remains the same as it was before - just moved into this function.

These three functions live in `pkg/catena/asset.go` with their tests being in `pkg/catena/asset_test.go`

---

## ✅ Definition of Done (DoD)
- [x] Helper added to the `catena` package alongside `ParamInfosForRequest`
- [x] First chunk carries metadata/digest/encoding/cachable; continuation chunks carry payload bytes only
- [x] URL-kind payloads (no embedded bytes) and assets smaller than the chunk size go out as a single chunk
- [x] Returns `Ok` on completion, and maps a failed `Send` to `Internal`, stopping the walk immediately
- [x] `oneOfEverything` `ReadAsset` handler refactored to delegate to the helper
- [x] Unit tests added (mirroring `param_info_test.go` coverage: single chunk, multi-chunk, URL kind, send failure)

---

## 🛠️ Implementation Notes
- Added `asset.go` to `sdks/go/pkg/catena`
- Added `asset_test.go` to `sdks/go/pkg/catena` <- 100% coverage of `asset.go`
   - single-chunk embedded payloads 
   - single-chunk URL payload
   - invalid payload combination
   - multi-chunk streaming
   - invalid chunk size handling
   - stream failure behaviour
- Modified `sdks/go/examples/oneOfEverything/asset_handler.go` to use helper function.

---

## 🔗 Related Issues / Links
Closes Issue: #1498


sdks/go/pkg/catena/asset.go
```diff
@@ -0,0 +1,104 @@
...
+func SendAssetChunks(slot uint16, fqoid string, stream Stream[st2138.Asset], payload st2138.DataPayload, cachable bool) StatusResult {
+	return sendChunks(slot, fqoid, stream, payload, cachable, defaultAssetChunkSize)
+}
+
+// SendAssetChunksWithSize streams payload using assetChunkSize and returns InvalidArgument when the size is not positive.
+func SendAssetChunksWithSize(slot uint16, fqoid string, stream Stream[st2138.Asset], payload st2138.DataPayload, cachable bool, assetChunkSize int) StatusResult {
+	return sendChunks(slot, fqoid, stream, payload, cachable, assetChunkSize)
+}
+
+// Stream the asset in chunks to demonstrate large-object delivery:
```
abrown-RV, line 61: This is not a demonstration here, this is the real SDK code.

sdks/go/examples/oneOfEverything/asset_handler.go
```diff
@@ -32,38 +27,8 @@ func registerAssetHandlers(srv catena.Server, assets *sync.Map) {
...
-				if err := stream.Send(chunk); err != nil {
-					logger.Warning("Asset download stream closed", "slot", slot, "fqoid", fqoid, "error", err)
-					return catena.StatusWithCode(catena.StatusCodeInternal, "failed to send asset: "+err.Error())
-				}
-				sent = end
-			}
-
-			logger.Info("Asset download complete", "slot", slot, "fqoid", fqoid, "size", len(data))
-			return catena.StatusWithCode(catena.StatusCodeOk, "")
+			return catena.SendAssetChunks(slot, fqoid, stream, stored.payload, stored.cachable)
```
abrown-RV, line 31: I think we should still have some comment on what the helper is doing.

sdks/go/pkg/catena/asset_test.go
```diff
@@ -0,0 +1,203 @@
...
+		result := SendAssetChunksWithSize(1, "asset", stream, payload, false, 3)
+
+		if result.Code != StatusCodeInternal {
+			t.Fatalf("expected INTERNAL, got %v", result)
+		}
+
+		if len(stream.Items) != 2 {
+			t.Fatalf("expected 2 chunks before failure, got %d", len(stream.Items))
+		}
+	})
```
abrown-RV, line 202: Add a test for 0 length asset. Should still work (it doesn't), and just return the empty payload. Consider an empty file used as a lock or something. Is a valid enough use case.

sdks/go/pkg/catena/asset.go
```diff
@@ -0,0 +1,104 @@
...
+// chunks carry only payload bytes (never the URL: a payload can
+// have one or the other, and continuation chunks only ever exist
+// for embedded payloads), which the client concatenates in send
+// order. Assets up to assetChunkSize (and URL-kind assets, which
+// have no embedded bytes) go out as a single chunk.
+func sendChunks(slot uint16, fqoid string, stream Stream[st2138.Asset], payload st2138.DataPayload, cachable bool, assetChunkSize int) StatusResult {
+
+	if assetChunkSize <= 0 {
+		logger.Error("Invalid asset chunk size", "slot", slot, "fqoid", fqoid, "assetChunkSize", assetChunkSize)
+		return StatusWithCode(StatusCodeInvalidArgument, "invalid asset chunk size")
```
abrown-RV, line 73: Should return a `StatusCodeInternal` StatusCodes are sent back to the client. So if BL called with a 0 size, the client would receive an invalid arguement code when they did nothing wrong.

abrown-RV (changes requested):
Good job Omar, just a few minor niggles.

sdks/go/pkg/catena/asset_test.go
oshah-rv: Added the test and also handled 0 length asset in the helper.

sdks/go/pkg/catena/asset.go
```diff
@@ -0,0 +1,127 @@
...
+	return sendChunks(slot, fqoid, stream, payload, cachable, defaultAssetChunkSize)
+}
+
+// SendAssetChunksWithSize streams payload using assetChunkSize and returns StatusCodeInternal when the size is not positive.
+func SendAssetChunksWithSize(slot uint16, fqoid string, stream Stream[st2138.Asset], payload st2138.DataPayload, cachable bool, assetChunkSize int) StatusResult {
+	return sendChunks(slot, fqoid, stream, payload, cachable, assetChunkSize)
+}
+
+// createEmptyAsset is an helper function that creates an empty asset with the given payload and cachable flag
+func createEmptyAsset(payload st2138.DataPayload, cachable bool) st2138.Asset {
```
abrown-RV: Sorry for the scope creep, but I'd prefer `ToAsset` to allow empty payloads, instead of this special case only available through chunking. So `ToAsset` should accept empty payloads as a real asset length 0.

| Case | Url | Payload | Result |
|------|-----|---------|--------|
| 1 | `!= ""` | `len == 0` | Url variant |
| 2 | `!= ""` | `len > 0` | error (both set) |
| 3 | `== ""` | `len == 0` | Payload as an empty `[]byte{}` |
| 4 | `== ""` | `len > 0` | Payload variant |

Just make sure that the opposite direction also works `st2138.FromAsset` and `st2138.dataPayloadFromProto`.

This I think is actually the cause of a bug we saw earlier, and couldn't nail down. So if you wish, you can revert `sendChunks` to just call `st2138.ToAsset` with no special handling for empty payloads. And then create and immediately address an issue allowing `To/FromAsset` to handle empty payloads.

abrown-RV (changes requested):

sdks/go/pkg/catena/asset.go
oshah-rv: Sounds good, I was thinking of doing that but was worried about it causing changes in elsewhere. To make these changes I updated `dataPayloadFromProto()` which is the underlying function that `ToAsset()` calls and which returns an error when payload and url fields are empty. 

 Right now, these changes are causing one unit test in `param_test.go` and `value_test.go` to fail. 
 1. In `param_test.go` it is a test to create an empty `Param` type (testing `NewParamData()`). The behaviour here changes as an empty payload input now creates a `Param` with an empty payload, rather than returning nil in the empty payload case. 
 2. In `value_test.go` it fails on a test of `ToProto()` with an empty payload. Similarly this would previously return an error if an empty payload was provided but now it returns an actual DataPayload proto type with an empty payload.

If we don't want these behaviour changes to `NewParamData()` and `ToProto()` then alternatively I can make changes directly to `ToAsset()` so only it will be able to accept empty datapayloads.

abrown-RV (changes requested):
Need to implement the reverse for empty payloads, in `dataPayloadFromProto` it shouldn't err on the 0 payload either

Issue #1478, opened 2026-07-27
[Go] Rest Transport CORS


abrown-RV:
## 📖 Description
The REST transport currently emits CORS headers only on the SSE `/connect` endpoint, and even there it reflexively sets `Access-Control-Allow-Credentials: true` while reflecting any `Origin`. Every other endpoint (device, `value`, `values`, `asset`, `command`, `param`, `param-info`, `language-pack`, `languages`, `devices`, `health`) sends no CORS headers at all, and no route answers preflight `OPTIONS`.

In practice this means a browser can only talk to the REST transport through a same-origin backend proxy today; a direct browser-to-device connection is impossible because preflights 405/404 and simple responses carry no `Access-Control-Allow-Origin`. We want first-class, opt-in CORS so browsers can connect directly, while staying out of the way when a gateway/proxy already handles CORS.

---

## 🎯 Goal / Outcome
Uniform, configurable CORS applied by a single middleware across **all** REST endpoints (including `/connect`), off by default, enabling direct browser access when the operator opts in. Because auth is JWT bearer only (no ambient credentials), we do **not** set `Access-Control-Allow-Credentials`, which lets a literal `*` work for the permissive case.

---

## ✅ Definition of Done (DoD)
- [ ] `RestOptions` gains CORS fields with env/flag wiring and defaults in `DefaultRestOptions`
- [ ] Single CORS middleware wraps the mux in `Start`; the inline CORS block is removed from `handleConnect`
- [ ] Preflight `OPTIONS` handled uniformly: `204` + `Allow-Methods`/`Allow-Headers`/`Max-Age`
- [ ] Off by default (empty/nil `AllowedOrigins`); enabled iff origins are configured
- [ ] Origin policy: `*` wins → literal `*` (no `Vary`); otherwise exact-match request `Origin`, echo it + `Vary: Origin`; no match → omit
- [ ] Exact origin string match only (no prefix/substring); `Origin: null` treated as not-allowed
- [ ] No `Access-Control-Allow-Credentials` header emitted (bearer-token auth; no cookies/mTLS ambient creds)
- [ ] `ExtraAllowedHeaders` (and optionally `ExtraAllowedMethods`) unioned onto SDK-owned required baselines, deduped case-insensitively
- [ ] Slice flag/env parser filters empty tokens so `--rest-allowed-origins=` == nil (off)
[...]

PR #1514, opened 2026-09-18
[GO] Added CORS support to REST transport

KeonF-ross:
## 📖 Description
The Go REST transport only emitted CORS headers on SSE `/connect`, and even there it reflected any `Origin` and set `Access-Control-Allow-Credentials: true`. Other routes sent no CORS headers, and no route answered preflight `OPTIONS`, so a browser could not talk to the device directly.

This PR adds opt-in, configurable CORS to the REST transport. CORS stays **off** when `AllowedOrigins` is empty. When 'AllowedOrigins' is configured, every REST endpoint (including `/connect`) gets the same origin policy, preflight `OPTIONS` returns `204`, and `Allow-Credentials` is never set.

---

## ✅ Definition of Done (DoD)
- [x] `RestOptions` gains CORS fields with env/flag wiring and defaults in `DefaultRestOptions` (`CorsMaxAge` is `600 * time.Second`)
- [x] Single CORS middleware wraps the mux in `Start`; inline CORS removed from `handleConnect`
- [x] Preflight `OPTIONS` handled uniformly: `204` + `Allow-Methods` / `Allow-Headers` / `Max-Age`
- [x] Off by default (empty/nil `AllowedOrigins`); enabled iff origins are configured
- [x] Origin policy: `*` wins → literal `*` (no `Vary`); otherwise exact-match request `Origin`, echo it + `Vary: Origin`; no match → omit
- [x] Exact origin string match only (no prefix/substring); `Origin: null` treated as not-allowed
- [x] No `Access-Control-Allow-Credentials` header emitted
- [x] `ExtraAllowedHeaders` / `ExtraAllowedMethods` unioned onto SDK-owned baselines, deduped case-insensitively
- [x] Slice flag/env parser filters empty tokens so `--rest-allowed-origins=` == nil (off)
- [x] Tests: preflight 204 + headers, allowed origin echoed, disallowed origin omitted, `*` literal, `*` beats specific list, empty-flag == off, extra headers in preflight
- [x] `REST.md` documents options, defaults, and gateway-vs-direct guidance

---

## 🛠️ Implementation Notes
- Added CORS fields on `RestOptions` in `sdks/go/pkg/config/config.go` 
    - `AllowedOrigins`, `ExtraAllowedHeaders`, `ExtraAllowedMethods`, `CorsMaxAge`.
    - `DefaultRestOptions` leaves origins/extras nil and sets `CorsMaxAge` to 10 minutes.
- Added `extractStringSlice` and `extractDuration` in `sdks/go/pkg/config/load.go`
    -  `parseStringSlice` comma-splits, trims, and drops empty tokens.
    -  `parseDuration` accepts Go durations (`10m`, `600s`) and treats a bare integer as seconds.
- Added `--rest-allowed-origins`, `--rest-extra-allowed-headers`, `--rest-extra-allowed-methods`, `--rest-cors-max-age` (and matching `CATENA_REST_*` env vars)
- Added `sdks/go/pkg/transports/rest/cors.go`
    - `withCORS` no-ops when origins are empty; otherwise applies origin policy on every request and short-circuits `OPTIONS` with 204; 
    - Contains SDK baselines `GET, POST, PUT, DELETE, OPTIONS` and `Content-Type, Authorization, Accept, Language, Detail-Level`
- `Transport` stores CORS config; `Start` sets `Handler' to  't.withCORS(t.mux)`
- Removed the per-request CORS block from `handleConnect`
- Added/modified Tests: 
    - `sdks/go/pkg/transports/rest/cors_test.go`
    -  config/load tests for slice/duration parsing; 
- Documented CORS in `sdks/go/pkg/transports/rest/REST.md` and listed the flags in `sdks/go/pkg/config/README.md`

---

## 🔗 Related Issues / Links

- [[GO] REST TRANSPORT CORS (ISSUE #1478)](https://github.com/rossvideo/Catena/issues/1478)

---

## 📝 Additional Context
Auth is JWT bearer only (no cookies / ambient credentials), so this PR never sets `Access-Control-Allow-Credentials`. This also allows literal `*` to be safe.

CORS was also tested manually by sending requests directly from a browser, below are the results:

**CORS off**
image
image

**CORS on, disallowed origin**
image
image
image

**CORS on, allowed origin**
image
image

*NOTE: Cors max age was set to 1200 when the tests were performed, hence the deviation from the default

---

## 🚧 Risks / Open Questions

---

## ✨ Updates


sdks/go/pkg/config/config.go
```diff
@@ -225,7 +231,13 @@ func DefaultDashboardOptions() DashboardOptions {
 }
 
 func DefaultRestOptions() RestOptions {
-	return RestOptions{Port: 9080}
+	return RestOptions{
+		Port:                9080,
+		AllowedOrigins:      nil,               // CORS OFF by default (respects gateway/proxy topology)
+		ExtraAllowedHeaders: nil,               // required headers already cover the ST 2138 surface
+		ExtraAllowedMethods: nil,               // fixed REST method set; extension unlikely
+		CorsMaxAge:          600 * time.Second, // 10 min preflight cache; only used once CORS is enabled
```
abrown-RV:
```suggestion
		CorsMaxAge:          10 * time.Minute, // 10 min preflight cache; only used once CORS is enabled
```

abrown-RV, line 239:
```suggestion
		CorsMaxAge:          10 * time.Minute, // preflight cache; only used once CORS is enabled
```

sdks/go/pkg/transports/rest/cors_test.go
```diff
@@ -0,0 +1,208 @@
...
+	transport.runtime.(*transporttest.StubServerRuntime).WithConnection(transporttest.MakeTestConnection(1))
+
+	req := httptest.NewRequest(http.MethodGet, "/st2138-api/v1/connect", nil).WithContext(ctx)
+	req.Header.Set("Origin", "https://example.com")
+	rec := httptest.NewRecorder()
+
+	go h.ServeHTTP(rec, req)
+	time.Sleep(100 * time.Millisecond)
+	cancel()
+	time.Sleep(50 * time.Millisecond)
```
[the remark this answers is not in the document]
abrown-RV: We've got an SSE wrapper in rest_transport_test you probably want to use.

sdks/go/pkg/transports/rest/cors_test.go
```diff
@@ -0,0 +1,208 @@
...
+	req := httptest.NewRequest(http.MethodOptions, "/st2138-api/v1/health", nil)
+	req.Header.Set("Origin", "https://example.com")
+	rec := httptest.NewRecorder()
+	h.ServeHTTP(rec, req)
+
+	if rec.Code == http.StatusNoContent {
+		t.Fatalf("expected OPTIONS not to be short-circuited when CORS is off, got %d", rec.Code)
+	}
+	assertStatus(t, rec, http.StatusMethodNotAllowed)
+	assertHeader(t, rec, "Access-Control-Allow-Origin", "")
```
abrown-RV: Should make a new `assetHeaderNotPresent` or something to differentiate between empty and not present.

sdks/go/pkg/transports/rest/cors_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	headers := unionCORSHeaders(requiredHeaders, []string{"X-Tenant-Id", "authorization"})
+	assertCSVContains(t, strings.Join(headers, ", "), "Content-Type", "Authorization", "X-Tenant-Id")
+	// first-seen casing for Authorization (from required), not the extra "authorization"
+	if headers[1] != "Authorization" {
+		t.Errorf("expected first-seen Authorization casing, got %q", headers[1])
+	}
+}
+
+func assertCSVContains(t *testing.T, csv string, tokens ...string) {
```
abrown-RV: may as well put this in `helpers_test.go` in case some other part of rest wants it

abrown-RV (changes requested):
Looking good Keon, those copilot comments make sense I think so deal with those as well.

sdks/go/pkg/transports/rest/helpers_test.go
```diff
@@ -155,6 +158,13 @@ func assertHeader(t *testing.T, rec *httptest.ResponseRecorder, key, expected st
 	}
 }
 
+func assertHeaderNotPresent(t *testing.T, rec *httptest.ResponseRecorder, key string) {
+	t.Helper()
+	if rec.Header().Get(key) != "" {
+		t.Errorf("expected header %s not to be present, got %q", key, rec.Header().Get(key))
```
abrown-RV, line 164: .Get() collapses absent and empty into "" so this assert doesn't actually asset not present.
```suggestion
	if values := rec.Header().Values(key); len(values) != 0 {
		t.Errorf("expected header %s not to be present, got %q", key, values)
```

sdks/go/pkg/transports/rest/cors.go
```diff
@@ -0,0 +1,124 @@
...
+			w.Header().Set("Access-Control-Max-Age", maxAge)
+			w.WriteHeader(http.StatusNoContent)
+			return
+		}
+		// Handle actual requests
+		next.ServeHTTP(w, r)
+	})
+}
+
+func applyCORSOrigin(w http.ResponseWriter, allowed []string, origin string) {
```
abrown-RV, line 79: mark this up with comments explaining the purpose of each of these cases. CORS is important enough I want to be explicit on its workings. And this is kind of the core of everything right here.

abrown-RV (changes requested):

abrown-RV:
> Negative max-age values must be handled, and server-level CORS wiring needs integration coverage.

Ok fine copilot, common handling for this seems to be just omit the header if its invalid, making it clearer what the browser should do. (Quick reading suggests that some will treat it as 0 and some will revert to their default). Also could add a `extractPositiveDuration` to `config/load.go` to protect against it on the cli side.

On the integration, we don't really have proper integration testing, but you could use the pattern in `rest_transport_test.go` where it has some helpers to loop through all endpoints. So like, call `OPTIONS` against each endpoint as (blackbox) proof that they all pipe through the CORS handler.

sdks/go/pkg/config/load.go
```diff
@@ -322,6 +327,25 @@ func parseDuration(s string) (time.Duration, error) {
 	return 0, fmt.Errorf("%s is not a valid duration", s)
 }
 
+// parseDuration but only returns a duration if it is greater than 0
+// otherwise it returns 0.
+func parsePositiveDuration(s string) (time.Duration, error) {
```
abrown-RV: Chain with `parseDuration`.

Comment is inaccurate, "otherwise, it errors"

sdks/go/pkg/transports/rest/rest_transport_test.go
```diff
@@ -1463,6 +1463,57 @@ func TestTransport_Health_MethodNotAllowed(t *testing.T) {
 	assertStatus(t, rec, http.StatusMethodNotAllowed)
 }
 
+func TestTransport_CORS_PreflightAllEndpoints(t *testing.T) {
```
abrown-RV: my apologies Keon, I got confused with server_test.go up in the catena pkg when I was talking about infrustruture for iterating through all the endpoints. On balance I think because we don't have that infrastructure and I don't think we want to add it at this time, this is more of a maintenance pain than its utility as a test. As long as (below) withCors works as tested, its probably fine.

What copilot was getting at was just a test to ensure that t.withCors is called within Transport.Start(). So lets go with a test that explicitly calls transport.start() instead of working around it with httptest. And then does the preflight to assert that cors is initialized properly when the http server is created in the normal way. Hint, use port: 0 for the OS to assign an ephemeral port that isn't already taken.

abrown-RV (changes requested):

sdks/go/pkg/transports/rest/cors.go
```diff
@@ -0,0 +1,128 @@
...
+	"net/http"
+	"slices"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// SDK-owned CORS baselines. Operators append via ExtraAllowed* only.
+var requiredMethods = []string{"GET", "POST", "PUT", "DELETE", "OPTIONS"}
+var requiredHeaders = []string{"Content-Type", "Authorization", "Accept", "Language", "Detail-Level"}
```
abrown-RV: Happened to be testing with swagger, and that's already putting in the `Request-Start` header even if go doesn't really deal with it yet. So please put that in now.
```suggestion
var requiredHeaders = []string{"Content-Type", "Authorization", "Accept", "Language", "Detail-Level", "Request-Start"}
```

sdks/go/pkg/transports/rest/rest_transport_test.go
KeonF-ross: I took the TestTransport_Start and modified it to send `OPTIONS` and check the returned headers. Is that what you wanted?

sdks/go/pkg/transports/rest/cors.go
```diff
@@ -0,0 +1,128 @@
...
+	headers := strings.Join(unionCORSHeaders(requiredHeaders, t.extraAllowedHeaders), ", ")
+	maxAge := strconv.FormatInt(int64(t.corsMaxAge/time.Second), 10)
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		applyCORSOrigin(w, t.allowedOrigins, r.Header.Get("Origin"))
+		// Handle preflight requests
+		if r.Method == http.MethodOptions {
+			w.Header().Set("Access-Control-Allow-Methods", methods)
+			w.Header().Set("Access-Control-Allow-Headers", headers)
+			w.Header().Set("Access-Control-Max-Age", maxAge)
```
abrown-RV: need to omit if a negative manages to still get here. (BL can hardcode options and bypass the cli args)

sdks/go/pkg/transports/rest/rest_transport_test.go
```diff
@@ -1463,6 +1463,75 @@ func TestTransport_Health_MethodNotAllowed(t *testing.T) {
 	assertStatus(t, rec, http.StatusMethodNotAllowed)
 }
 
+func TestTransport_CORS_Preflight(t *testing.T) {
```
abrown-RV: This let me notice that the main RestTransport_Start that this was based on has a TOCTOU race, so if you merge with develop that's available now, and can stamp out the TOCTOU here as well. 

The issue is that using a listener to "grab" a port and then closing it means something else could pop in a grab the port before the server. Unlikely, but now I can't un-see it.

Also the sleep is unnecessary because `transport.Start()` makes its listener synchronously meaning that the OS will queue requests arriving on the port before the http server is ready to handle them.

abrown-RV (changes requested):
shone a light on a TOCTOU race in the tests.

abrown-RV (approved):

